PR #10, opened 2025-01-28
update: fix keys for listing and error msg

bilalesi:
1. fix listing keys for users
2. fix error message when loading virtual labs and log it for debugging


pgetta (approved):
Looks good.

Left a small comment.

PR #11, opened 2025-01-28
Update dot env files

pgetta:
Now local dev server points to the staging environment.

Also, entries which are not common for all environments were moved out of .env, so that in deployments they have to be explicitly provided (otherwise the server will throw an error). This will prevent the leakage of wrong defaults.


bilalesi (approved):

pgetta:
Thank you, @bilalesi

PR #12, opened 2025-01-28
Hotfixes obi

g-bar:


bilalesi:
there is a regression, if you go to simulate-> single-neuron or synaptome and click on a row, then it will redirect to the details page with 404 error, but the redirection should not happen in the first place, unless the user select the row and click on the `View` button.

bilalesi:
> there is a regression, if you go to simulate-> single-neuron or synaptome and click on a row, then it will redirect to the details page with 404 error, but the redirection should not happen in the first place, unless the user select the row and click on the `View` button.

onCellClick should be passed as prop to https://github.com/openbraininstitute/core-web-app/blob/dd1542572a1aec5c5ec12b3276a50388986a3aa9/src/components/explore-section/ExploreSectionListingView/index.tsx

src/components/explore-section/ExploreSectionListingView/index.tsx
```diff
@@ -53,6 +56,16 @@ export default function ExploreSectionListingView({
...
 
+  const router = useRouter();
+
+  // TODO: Remove onCellClick prop
+  // eslint-disable-next-line
+  onCellClick =
+    onCellClick ??
+    ((basePath, record) => {
+      router.push(detailUrlBuilder(basePath, record));
+    });
```
bilalesi, line 67: this component is used everywhere, there is some places we don't need the onCellClick or has a different logic, so it should be passed as prop

g-bar: Yeah, hadn't notice sometimes we didn't need it, will fix.

bilalesi (approved):

pgetta:
I've tested and the issues seem to be resolved.

PR #14, opened 2025-01-29
Fix landing page for mobile screens (#24)

tolokoban:
This PR relates to [issue #24](https://github.com/openbraininstitute/prod-obi-site/issues/24).

It adapts the landing page layout to make it useable on screens of every size, even phones.


src/components/LandingPage/sections/SectionPricing/SectionPricing.tsx
```diff
@@ -35,9 +35,9 @@ export default function SectionPricing({ className }: SectionPricingProps) {
         Have unique requirements? Contact us to discuss custom pricing tailored to your special
         usage of the Open Brain Platform. We are here to create solutions that fit your goals.
       </p>
-      <p>
+      <div>
         <EmailButton email="support@openbraininstitute.org">Need some support?</EmailButton>
-      </p>
+      </div>
```
bilalesi, line 40: **nit**: it seems it need a bit of spacing on top 🤔 
Screenshot 2025-01-29 at 14 11 41

src/components/LandingPage/CenteredColumn/CenteredColumn.tsx
bilalesi, line 1: this is general comment and are nit-picks (so it's up to u @tolokoban):
1. for column style of the sections: it appear that there is double lines separator, (we can hide one)
Screenshot 2025-01-29 at 14 16 26


2. it might you discussed with loris about this, so ignore it 
the menu is centered, but it's different then the design


Screenshot 2025-01-29 at 14 19 47
Screenshot 2025-01-29 at 14 20 50



3. this test in Sumusung S20, the "terms and conditions" seems centered against others

Screenshot 2025-01-29 at 14 22 50


4. the vertical line is on top of the writing due text taking two lines (Terms & conditions)

Screenshot 2025-01-29 at 14 24 30

5. in different places the text seems too big for a mobile phone, 
6. for sumsung phone,  some pages is scrollable horizontally probably due some overflow,

tolokoban: Thanks a lot, @bilalesi for this thorough review that helped fix many issues.
I have added my corrections here.

src/components/LandingPage/sections/SectionPricing/SectionPricing.tsx
tolokoban: I let this one aside for now, because the interface will still change a lot and this is not directly related to mobile layout. But, you are right about the fact that we should add more space before the button and less between the lines inside the button.

bilalesi:
1. social media logos and text:

Screenshot 2025-01-30 at 11 14 06

2. OUR MISSION: need more spaces between the title "Advance Neuroscience Understanding" and the big text
second image from design

Screenshot 2025-01-30 at 11 15 33
Screenshot 2025-01-30 at 11 16 30

3. PRICING: the email box should be aligned with text and have some space between them

Screenshot 2025-01-30 at 11 17 23

4. CONTACT: two emails boxes, (in the design are vertical), but here are horizantal (it might you agree with loris to change 🤔)
Screenshot 2025-01-30 at 11 19 38

5. TERMS and CONDITIONS: there is tiny difference  (number and text are vertical), there is also a separator between each point (again up to u if need a change)

Screenshot 2025-01-30 at 11 21 31

Screenshot 2025-01-30 at 11 21 41

6. in the mobile version 

Screenshot 2025-01-30 at 11 25 07

7. mobile version (nit: the button is a bit overflowed)
Screenshot 2025-01-30 at 11 26 17

tolokoban:
1. **FIXED**. This was a regression due to the fact that I removed the text when the screen was too small.
2. **FIXED**. I think this was already here with the branch I used as a base, but it's good to fix it.
3. **FIXED**.
4. This one is a personal choice based on a remark from George that he does not like wasting space. For instance, we started by stacking all the portals one upon the other, but JD and Georges decided we needed to have at least two because "there is too much empty space". So, I prefer to let this as is for now because I'm sure everything will have to be changed again soon  :-)
5. **FIXED**. It was not acceptable on mobile.
6. Idem.
7. **FIXED**. Added a max-width depending on the screen width.

bilalesi (approved):

PR #16, opened 2025-01-29
OBI Notebooks

g-bar:


src/app/api/github/downloadNotebook/route.ts
```diff
@@ -0,0 +1,25 @@
...
+export async function GET(request: Request): Promise<NextResponse> {
+  const { searchParams } = new URL(request.url);
+  const folder = searchParams.get('folder');
+
+  if (!folder) {
+    return NextResponse.json({ error: 'Missing required parameters' }, { status: 400 });
+  }
+
+  try {
+    const zipContent = await downloadZippedFolder(folder);
```
pgetta, line 14: While this implementation works fine I'm a bit concerned about the following: given that compression is a resource intensive operation this might become a bottleneck and a possible target for a malicious actor - with relatively small amount of parallel requests the next.js server might become completely overwhelmed and stop serving legit requests.

Possible solutions: 
* Limiting the number of parallel compression processes.
* Moving compression into a separate service.
* Updating notebook repo (maybe with some automation) to include archived copies, so we don't re-compress multiple times what has been already compressed - I think that is the best solution, given we have a limited number of notebooks.

src/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/(pages)/notebooks/NotebookTable.tsx
```diff
@@ -0,0 +1,316 @@
...
+    const link = document.createElement('a');
+    const downloadUrl = window.URL.createObjectURL(blob);
+    link.href = downloadUrl;
+    link.download = `${notebookName}.zip`;
+    document.body.appendChild(link);
+    link.click(); // Trigger the download
+    document.body.removeChild(link); // Clean up the DOM
+
+    window.URL.revokeObjectURL(downloadUrl);
+  };
```
Comment (pgetta) [inline src/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/(pages)/notebooks/NotebookTable.tsx:85]: There is a file-saver module available that does the same job, you can have a look at an example [here](https://github.com/openbraininstitute/core-web-app/blob/develop/src/util/simulation-plotly-to-csv.ts#L35).

In general, a bit more efficient way to download publicly available resource is to create an "a" tag with download property, which almost doesn't require any additional memory, as it mostly streams directly to the file system (instead of loading the whole object into JS memory first) - so for the case in question I would use this method instead.

The only scenario where this can not be applied is for resources with authentication, as there is no way to set up auth header for an HTTP request generated by clicking on an "a" element.

pgetta (commented):
Great job, Gil!

I've left a couple of comments, please see what makes sense.

P.S. We need to do something with the linter complaining about the `console` calls, either disabling that rule in linter or switching to a separate logging library, for better control and more flexibility.

package.json
```diff
@@ -60,10 +60,12 @@
     "d3": "^7.8.4",
     "d3-sankey": "^0.12.3",
     "date-fns": "^3.6.0",
+    "dayjs": "^1.11.13",
```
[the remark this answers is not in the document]
g-bar: Antd-5 switched to dayjs
`By default, Ant Design uses [Day.js](https://day.js.org/) to handle time and date. Day.js is an immutable date-time library alternative to Moment.js with the same API.`

https://ant.design/docs/react/use-custom-date-library

But if it's not a direct dep, the linter doens't pick it up when importing it.

src/app/api/github/downloadNotebook/route.ts
g-bar: In the future we might do away with downloading altogether and users will be able to run, edit notebooks on the platform.

I don't want to spend that much time on a solution that will be thrown away (like new service).
We will need this as well for user repos, not only our own so last solution doesn't work.

Not sure what to do about this.

src/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/(pages)/notebooks/NotebookTable.tsx
Comment (g-bar) [reply src/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/(pages)/notebooks/NotebookTable.tsx:85]: Yeah there is no place to point the a tag to, I'll use saveAs for now.

package.json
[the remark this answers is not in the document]
pgetta: Yeah, I see. So it's even more, for the time being we have:
* date-fns as explicit dependency in package.json.
* dayjs as a dependency of antd.
* momentjs as a dependency of antd -> rc-picker.

For the linter, as you mentioned, we need to use explicit dependency.

But the question is, does it make sense to use two date-time management libraries at the same time in a single project?

I see in this PR the usage of both:
* [Dayjs in Filter.tsx](https://github.com/openbraininstitute/core-web-app/blob/8b74e9de2f35ce23c539c30815a024742515a492/src/components/FilterControls/Filter.tsx#L2).
* [date-fns in NotebookTable.tsx](https://github.com/openbraininstitute/core-web-app/blob/8b74e9de2f35ce23c539c30815a024742515a492/src/app/virtual-lab/lab/%5BvirtualLabId%5D/project/%5BprojectId%5D/(pages)/notebooks/NotebookTable.tsx#L7).

For consistency I think it's best to pick just one, unless there are reasons to do the opposite (that's why I was asking the reasons, probably was not clear enough).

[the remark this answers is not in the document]
g-bar: Nevermind, I saw that you can configure RangePicker to use `date-fns` instead of Dayjs. So I will do that and delete dayjs as a direct dependency.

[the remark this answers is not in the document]
pgetta: Great, I've added this to my tech debt list. Will need to clean up at some point.

src/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/(pages)/notebooks/NotebookTable.tsx
Comment (pgetta) [reply src/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/(pages)/notebooks/NotebookTable.tsx:85]: Just realized, can't we point to the `/api/github/downloadNotebook?folder=${encodeURIComponent(directory)}` endpoint directly?

Comment (g-bar) [reply src/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/(pages)/notebooks/NotebookTable.tsx:85]: So we actually can. 
Downside is if the endpoint takes 2 3 second can't show a loading spinner as I do now. 
But actually these zips only contain 1 notebook (so Readme, metadata.json and notebook.ipynb) the ones I've tried are 16-24kB.
So I think it's better with the saveAs and show the loading spinner.

Comment (pgetta) [reply src/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/(pages)/notebooks/NotebookTable.tsx:85]: Indeed, sounds good.

.eslintrc.json
```diff
@@ -73,6 +73,7 @@
       {
         "additionalHooks": "useDebouncedCallback"
       }
-    ]
+    ],
+    "react/jsx-props-no-spreading": "off"
```
pgetta, line 77: Are you sure we want to globally disable this?

From [the docs](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md):
_Enforces that there is no spreading for any JSX attribute. This enhances readability of code by being more explicit about what props are received by the component. It is also good for maintainability by avoiding passing unintentional extra props and allowing react to emit warnings when invalid HTML props are passed to HTML elements._

For simpler apps I presume it's fine, but for more complex this becomes an important factor.

@bilalesi , @tolokoban do you have an opinion on that?
I think we should keep it enabled, but If the majority is fine with the change - let's go for it.

pgetta (commented):
Great work, Gil!

I've left one minor comment about the linter config.

.eslintrc.json
bilalesi: I think you’re right—it’s better to keep it on. There are cases where it’s useful, especially when passing many props without explicitly listing each one. In such cases, we can enable it per line or module if it occurs frequently. The downside, however, is that it allows unintended props to be passed to the component.

g-bar: > It is also good for maintainability by avoiding passing unintentional extra props and allowing react to emit warnings when invalid HTML props are passed to HTML elements.

This can be avoided by only enabling it for html attributes.

> "react/jsx-props-no-spreading": [<enabled>, {
>     "html": "ignore" | "enforce",
>     "custom": "ignore" | "enforce",
>     "explicitSpread": "ignore" | "enforce",
>     "exceptions": [<string>]
> }]

I think setting it to ignore for custom and enforce for html offers a good middle ground between readability, maintainability and flexibility.

pgetta: Huh, good to know it's possible to do that!

Though I was mostly concerned about the readability for custom components, where one has to look at different places to understand what's being (and can be) passed down as props.

While looking at the rule and its configuration I've found we can allow explicit spreading, which looks readable while being not too verbose. What do you all think about using that as a general rule, and as @bilalesi suggested, if extra flexibility is needed -  enable prop spreading per line/module?

g-bar: > Though I was mostly concerned about the readability for custom components, where one has to look at different places to understand what's being (and can be) passed down as props.

Yeah good point! I was thinking of the opposite case when you pass all of them, but we can enable it per line indeed.

g-bar: Explicit spreading is on by default. So I'll remove the rule.

pgetta (dismissed):
Nice work, Gil!

pgetta (approved):

PR #34, opened 2025-02-13
Feat/initial accounting UI

pgetta:


src/state/virtual-lab/lab.ts
```diff
@@ -110,3 +114,19 @@ export const virtualLabPlansAtom = atom<
...
+});
+
+export const virtualLabBalanceAtomFamily = atomFamilyWithExpiration(
+  ({ virtualLabId }: { virtualLabId: string }) =>
+    atom(async (get) => {
+      get(virtualLabBalanceRefreshTriggerAtom);
+
+      return getVirtualLabAccountBalance({ virtualLabId, includeProjects: true });
+    }),
+  { ttl: 20_000, areEqual: fastDeepEqual }
```
g-bar, line 131: Why do we need fastDeepEqual? the comparisons are shallow.

src/util/atoms.tsx
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param readFn - A function that returns the initial value for each atom in the family.
+ * @param options - An object with the following properties:
+ *   - `ttl`: The time-to-live (in milliseconds) for each atom in the family.
+ *   - `areEqual`: An optional function to compare the equality of the atom getter parameters.
+ * @returns An atom family with expiration functionality.
+ *
+ * TODO: add an option to refresh after expiration
+ */
+export function atomFamilyFactoryWithExpiration<T, P>(
```
g-bar, line 31: What's the difference with atomFamilyWithExpiration? Do we need both?

src/components/VirtualLab/VirtualLabSidebar/index.tsx
```diff
@@ -36,12 +44,14 @@ export default function VirtualLabSidebarContent({ virtualLabId }: { virtualLabI
       ),
       href: 'team',
     },
-    { key: LinkItemKey.Admin, content: 'Admin', href: 'admin' },
+    { key: LinkItemKey.Admin, content: 'Admin', href: 'admin', requires: { userRole: 'admin' } },
   ];
 
+  const complientLinkItems = linkItems.filter(linkItemFilter);
```
g-bar, line 50: Unimportant but it's  "compliant".

g-bar (approved):
Minor comments

src/components/VirtualLab/VirtualLabSidebar/index.tsx
pgetta: Oh, thanks, I've corrected that.

src/util/atoms.tsx
pgetta: I've left a single one of these and refactored the rest of the code to use it.

src/state/virtual-lab/lab.ts
pgetta: I've replaced this with `isEqual` from lodash for now not to add a new dependency.

But the usage of shallow comparison (which is a default one) doesn't work here as the there will be different objects coming from different components, so: `{ virtualLabId} !== { virtualLabId }`.

g-bar: Ah yes. I’m stuck in Python terminology, where ‘shallow’ is call ‘identity’.

And shallow compares values 1 level deep.  

Deep compares values and nested values.

it looks like in JS shallow is identity.

g-bar (approved):
Nice job.

pgetta:
Thank you, @g-bar

PR #35, opened 2025-02-13
Mailchimp subscribe route integration

bilalesi:
1. Add route to subscribe to newsletter
2. `${basePath}/api/newsletter` will be used in the subscribe form:
```
const options = {
  method: 'POST',
  headers: {'Content-Type': 'application/json'},
  body: '{"email":"xxx@gmail.com"}'
};

fetch('http://localhost:3000/api/newsletter', options)
  .then(response => response.json())
  .then(response => console.log(response))
  .catch(err => console.error(err));
```
**Possible responses:**
```
200: {
          message: 'Awesome! You have successfully subscribed!',
        },
4xx: {
        message: // please look ErrorMessageMap messages
        reason: ""
      },
500:{
        message: 'Oops! There was an error subscribing you to the newsletter',
        reason: ""
      }
```

### DIFF
--- a/src/app/api/newsletter/route.ts
+++ b/src/app/api/newsletter/route.ts
@@ -42,10 +42,16 @@ export async function POST(req: Request) {
   }
 
   const url = `https://${API_SERVER}.api.mailchimp.com/3.0/lists/${AUDIENCE_ID}/members`;
+  const tags = ['newsletter', 'website'];
+
+  if (env.NEXT_PUBLIC_DEPLOYMENT_ENV === 'staging') {
+    tags.push('test');
+  }
+
   const data = {
     email_address: emailValidation.data,
     status: 'subscribed',
-    tags: ['newsletter'],
+    tags,
   };
 
   try {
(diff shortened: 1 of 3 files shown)


src/app/api/newsletter/route.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    return Response.json(
+      {
+        message: getErrorMessage(result.title ?? 'default'),
+        reason: result.title ?? 'unknown',
+      },
+      { status: result.status ?? 400 }
+    );
+  } catch (error) {
+    console.error('error sending newsletter email', error);
```
tolokoban, line 86: If this is executed on the server, where will this `console.error` end up?
Is there a mechanism to get this in some sort of log file?

tolokoban (commented):
I just added a quick question about a `console.error`.
But the rest looks OK to me.

src/app/api/newsletter/route.ts
tolokoban: OK, so you can read it in the AWS console, if I get you.

tolokoban (approved):

src/app/api/newsletter/route.ts
pgetta: In coming weeks we will have Sentry account. I believe those errors which require immediate attention (e.g. unexpected errors in prod) should go there directly (we can set up notification channels - via teams, email, etc.).

Maybe it's good to set up some logging library and use that also for redundancy.

pgetta (approved):

PR #37, opened 2025-02-17
Added Sanity headless CMS

tolokoban:
The landing page is not yet finished, but Bilal needs this pull request to be able to finish the newsletter registration feature.


loris-olivier-obi (approved):
Nice ! Simply Nice!

loris-olivier-obi (approved):
Super nice

tolokoban:
The tests are not working since we installed Sanity libraries.
I don't know why, but the problem is that we currently really need to have Sanity in production quickly.

src/app/api/github/fetchFile/route.ts
g-bar, line 1: @tolokoban  these files (fetchFile/route.ts and downloadNotebook/route.ts)  were deleted in a previous MR, seems like merge re-introduced them, we can delete them in a new MR.

PR #38, opened 2025-02-17
update: add coming-soon page

bilalesi:
after updated discussion with JDC: 

basically no need to do any routing for the moment, 
the user will land in coming-soon page which is the home page (this is why i replace the `app/page.tsx` with the coming-soon page)

after the landing page will be ready the user will land in the landing page if he try to login will be redirect to the` /coming-soon` page


public/images/coming-soon/background.svg
pgetta, line 1: There is an issue with this image: it seems to be an small size SVG with hires png image embedded, also, there is a lot of overhead due to base64 compression: 3.7 MB base64 encoded vs 2.8 MB raw png image.

I would suggest extract the embedded PNG and replace the SVG with it, so that the image component can optimise it ([by default it's not optimized](https://nextjs.org/docs/app/api-reference/components/image#dangerouslyallowsvg)):

background

Btw, in the `develop` branch the build is complaining about missing `sharp` package and I've noticed that image optimisation is very slow in staging, probably due to this issue. I'll have a look and hopefully fix it.

bilalesi: i remeber i fixed the issue last year, it must be deleted somehow ?!

pgetta (approved):

PR #40, opened 2025-02-18
Explicitly notify user when out of funds when starting a single cell sim

pgetta:


src/state/simulate/single-neuron-setter.ts
```diff
@@ -250,12 +256,24 @@ export const launchSimulationAtom = atom<
       });
 
       if (!response.ok) {
+        let errorMessage = DEFAULT_ERROR_MSG;
+
+        try {
+          const errResponseObj = await response.json();
+          if (errResponseObj.error_code === LOW_FUNDS_ERROR_CODE) {
+            errorMessage = LOW_FUNDS_ERROR_MSG;
```
bilalesi, line 264: i suggest to have email notification from the accounting srv

bilalesi (approved):

src/state/simulate/single-neuron-setter.ts
pgetta: Yeah, we need to notify users, indeed.

Not sure it's good to do that when a particular operation fails, but rather with some user defined threshold, e.g. if the credit balance drops below 50 credits - there is a notification email. Or maybe both options are  good candidates. We can discuss that with Jean-Denis.

PR #43, opened 2025-02-18
add privacy page

bilalesi:


tolokoban:
Just a question: as a user, how do you come back to the "coming soon" page once you are on the privacy?
Maybe it's OK to just let the user hit the browser's back button, but I just wanted to be sure Loris does not have another idea.

tolokoban (approved):

bilalesi:
> Just a question: as a user, how do you come back to the "coming soon" page once you are on the privacy? Maybe it's OK to just let the user hit the browser's back button, but I just wanted to be sure Loris does not have another idea.

the privacy page will be opened in the new page

PR #46, opened 2025-02-18
update link for privacy, fix some styles

bilalesi:


src/app/_coming-soon/page.tsx
```diff
@@ -1,32 +1,31 @@
...
 import NewsletterForm from '@/components/coming-soon/newsletter-form';
 import Header from '@/components/coming-soon/header';
 import Main from '@/components/coming-soon/main';
-
-import { basePath } from '@/config';
+import Hero from '@/components/coming-soon/hero';
 import { classNames } from '@/util/utils';
 
 export default function Page() {
+  const [hydrated, setHydrated] = useState(false);
```
tolokoban, line 12: Just out of curiosity and to learn something new: why do we need to wait for the component to be rendered with this state?

bilalesi: the components in server side can not take the last shape, so it seems laggy when it"s rendered, it start small taking initial width, and then getting larger, it make a bad UI/UX
may be there is a better way, (sometime specifying the exact width, ... make it work,)  but with this server-first approach we always may face it

bilalesi: please lemme know if you have another idea or approch how to tackle this

PR #47, opened 2025-02-18
Add build and deploy GitHub actions for AWS environments

pgetta:
This change refactors our GitHub Actions, add those specific to build and push images for staging and production.

Also, I've bumped node.js version used in docker and GitHub Actions to v23 as it should be more performant and I was using it locally for quite some time and had no issues.


Dockerfile
```diff
@@ -11,39 +11,27 @@ RUN npm ci
...
-ARG NEXT_PUBLIC_ENVIRONMENT
-ARG CI_COMMIT_SHORT_SHA
-ARG NEXT_PUBLIC_VIRTUAL_LAB_API_URL
-ARG NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY
-ARG NEXT_PUBLIC_BBS_ML_PRIVATE_BASE_URL
-ARG NEXT_PUBLIC_BRAIN_REGION_ONTOLOGY_RESOURCE_TAG
-ARG NEXT_PUBLIC_CELL_COMPOSITION_TAG
-ARG NEXT_PUBLIC_CELL_SVC_BASE_URL
-ARG NEXT_PUBLIC_BLUE_NAAS_URL
+FROM node:23-alpine AS builder
```
g-bar, line 14: Just a question. Why do we need the builder step. If in runner step we copy just the build?

g-bar (approved):
Left a question but LGTM.

g-bar:
What do we mean by "AWS ends"?

pgetta:
> What do you mean by "AWS ends"?

Thank you, I've corrected the name, it should have been "AWS envs", I put the full word for clarity.

Dockerfile
pgetta: This is coming from [the example](https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile#L21) of a Dockerfile from Next.js devs.

From what I understand it has something to do with how the cache works in docker, to minimize unnecessary rebuilds, but I never dig deeper.

g-bar: Yeah caching makes sense. In the context of github, doesn't matter right? Or does it cache between jobs?
Anyway better to leave it as is. Thanks for explaining,

pgetta:
Thank you, Gil

Dockerfile
pgetta: Ah, so these stages are all executed in a single `docker build` run, that's not always like this, but at least in our case. You can read more about these [multi-stage builds](https://docs.docker.com/build/building/multi-stage/).

Just remembered that another reason to have them is to have minimal clean builds at the end without any temporary build artefacts.

In our case, the `builder` and the `runner` separation is to have a clean final image, and the `base` and the `builder` separation - that's probably for better caching.

pgetta: In the context of Github - runners have their own docker cache, and if the action lands to the one where something is cached - docker will used that.

g-bar: Ah nice! Thanks for the explanation.

PR #52, opened 2025-02-19
GitHub action to automate redeployment of ECS container in staging

DriesVerachtert:


DriesVerachtert:
@pgetta I'm not 100% sure it will work, but as far as I know, this might be the way to run that redeploy after a build of a container for staging. I guess we can only test this by merging it in staging.

I've tested the redeploy thing on its own, but didn't test that "on: workflow_run:" trigger part at all.

bilalesi:
thanks dries, this is really helpful

LGTM, one suggestion i think it's better to tag the deployment id by the last commit hash


```
aws ecs tag-resource \
            --resource-arn ${{ vars.AWS_STAGING_ECS_SERVICE }} \
            --tags key=deployment_id,value=${{ github.sha }}
```

.github/workflows/staging-redeploy.yml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          role-to-assume: ${{ vars.AWS_STAGING_ECS_REDEPLOY_ROLE_ARN }}
+          aws-region: ${{ vars.AWS_STAGING_REGION }}
+
+      - name: Update ECS Service
+        run: |
+          aws ecs update-service \
+            --cluster ${{ vars.AWS_STAGING_ECS_CLUSTER }} \
+            --service ${{ vars.AWS_STAGING_ECS_SERVICE }} \
+            --force-new-deployment
```
bilalesi, line 31: I noticed in the aws console that the deployment takes times to be functional and stable, may be waiting her for that to mark the workflow as complete give us good feedback without checking the console

DriesVerachtert: thank you for the review, yes you're right those additions would be useful.

I'll test this first as it's possible the policy of the role needs to be extended a bit to be able to update the tag of the aws service.

pgetta: I've added a command to wait until the service is stable. We can try to merge it and, as Dries said, see if it works.

pgetta:
@DriesVerachtert thanks for the solution, it works: https://github.com/openbraininstitute/core-web-app/actions/runs/13501664173

DriesVerachtert:
great! I'll try to add that suggestion of Bilal for the tagging in a separate PR.

PR #57, opened 2025-02-20
impelment new design for creation/invites vlab/project

bilalesi:
1. create virtual lab process changed to be page instead of modal
2. create project process changed to be page instead of modal
3. invite users has changed too, and the design shared wherever invites used


pgetta (dismissed):
Looks good.

Left a couple of very minor comments.

PR #61, opened 2025-02-20
Correctly parse nexus self urls

pgetta:
The assumption that current nexus deployment domain will be always in the _self is not correct.
It turns out we have some entities with bbp.epfl.ch domain in our current staging nexus.

Current solution parses the url backwards, so the result doesn't depend on the domain.


g-bar (approved):

bilalesi (approved):

pgetta:
Thanks to both of you!

PR #66, opened 2025-02-20
Hard link to coming-soon page

tolokoban:


src/components/LandingPage/layout/Menu/Menu.tsx
```diff
@@ -48,7 +48,9 @@ export default function Menu({ className, scrollHasStarted, section }: MenuProps
             type="button"
             aria-label="Log in"
             className={styles.loginButton}
-            onClick={() => gotoSection(EnumSection.ComingSoon)}
+            onClick={() => {
+              window.location.href = sanitizeURL('/coming-soon');
```
bilalesi, line 52: Nextjs don't recommend using the window better to use the link component one advantage is to prefetch the next page on hover
But let's merge the pr for the website to work properly and fix it later

bilalesi (approved):
Left small comment, can be address later

PR #71, opened 2025-02-21
add depolyment envs to env files

bilalesi:


pgetta (approved):
I guess later on it's best if we merge this with the `DEPLOYMENT_ENV` to avoid any confusion. I'll try to do that maybe next week.

PR #72, opened 2025-02-21
Serve static pages from the /

pgetta:
With this change:
* Static pages with content coming from Sanity are served from the `/`.
* Application basePath is removed - essentially it becomes `/`.
* Platform related pages are moved to a separate `/app` folder.

In case you have some ongoing changes - I assume it would be easier to resolve potential conflicts by moving folders into the `/app` folder first in your branch, and after - attempting to merge from develop or rebase.

One downside of current implementation is that the `/api` path will be shared across Core Web app and other services, I've added some notes to a readme file under `src/api` folder. In theory this can be fixed by moving all `/api` routes into `/app`, but that breaks the next-auth.js in local dev environment. I'll see later if switching to the newest v5 helps (that requires some migration).

Still todo: clean up imports and usage of the `basePath`. Remove unused components, e.g. landing page for SFN.


pgetta:
@bilalesi I'll submit another PR with conversion button -> link conversion for the navigation items. The change is already quite big.

g-bar (approved):
LGTM.

bilalesi (approved):

pgetta:
Thank you all

PR #77, opened 2025-02-24
Second round of feedbacks

tolokoban:
- (22) Scales grid on laptop screen is too wide, you have to scroll to the right. It should fit on screen
- (37) Footer needs a 30px padding around the main content
- (40) The message that confirms the newsletter registration is too big on phones.
- Homepage is on / now


loris-olivier-obi (approved):
Super PR!

src/middleware.ts
```diff
@@ -3,6 +3,7 @@ import { NextRequest, NextResponse } from 'next/server';
 
 const FREE_ACCESS_PAGES = [
   '/',
+  '/home',
```
pgetta, line 6: @tolokoban a small question: what's the reason to have `/home` when the root page is already serving the same content?

PR #79, opened 2025-02-24
Add accounting for memodel creation

pgetta:


src/components/build-section/virtual-lab/synaptome/molecules/SynaptomeConfigurationForm.tsx
```diff
@@ -263,12 +265,21 @@ export default function SynaptomeConfigurationForm({ org, project, resource }: P
...
+        count: 1,
       });
 
+      const resp = await accountingSession.useWith<Response>(() =>
+        fetch(resourceUrl, {
+          method: 'POST',
+          headers: createHeaders(session.accessToken),
+          body: JSON.stringify(sanitizedResource),
+        })
+      );
```
bilalesi, line 281: can't await in the `useWith` callback so the type will be fixed, and also be consistent for all calls

src/services/accounting/index.ts
```diff
@@ -0,0 +1,74 @@
...
+      result = await executorFn();
+    } catch (error) {
+      await cancelOneshotReservation(jobId);
+      throw error;
+    } finally {
+      await reportOneshotUsage({
+        ...this.reservationRequest,
+        jobId,
+        timestamp: new Date().toISOString(),
+      });
```
bilalesi, line 69: idk the shape of the reporting, does it also track errors ?

src/types/accounting/index.ts
```diff
@@ -82,3 +84,30 @@ export type ProjectJobReports = {
 };
 
 export type ProjectJobReportsResponse = VlmResponse<ProjectJobReports>;
+
+/* -------------------------------------- Reservation types ------------------------------------- */
```
bilalesi, line 88: i think no need for this comments, since names are clear enough

bilalesi (approved):
LGTM, left small questions, suggestions

src/types/accounting/index.ts
pgetta: I've removed those.

bilalesi (approved):

PR #83, opened 2025-02-24
Resolve routing conflict for accounting service

pgetta:
After the recent routing change the `/api` subroutes are shared between core web app and other services, at the moment core web app has a conflict with accounting service for the `/api/accounting` route.

This change addressed that.

In a longer run the plan is to make accounting service publicly available, that will remove any need for the core web app to have mentioned routes.


g-bar (dismissed):

g-bar:
Not sure I fully fullo
/api/accounting is the accounting service.
/api/public-accounting is api route that does what? Proxies the accounting service?

pgetta:
@g-bar, that is a good question. Initially there was an idea that we will keep accounting service only accessible from VPC to increase security, also current implementation doesn't have any authorization checks.

Now, after trying to use it, it's evident that this is a bad design and needs to be refactored. I've already started adding authorisation logic to accounting services and hopefully will release it soon. Once this is done - I'll remove those routes from the web app.

Please treat this change as a temporary solution.

g-bar (approved):

PR #86, opened 2025-02-25
Consolidate push and deploy actions for staging ci

pgetta:
There is another small change that prevents running production build on branches different from `main`.


.github/workflows/publish-production-image.yml
```diff
@@ -13,6 +13,12 @@ jobs:
       ENVIRONMENT: staging
       IMAGE_NAME: ${{ vars.PUBLICECR_URI }}
     steps:
+      - name: Fail if branch is not main
+        if: github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main'
+        run: |
```
bilalesi, line 18: shouldn't specify the branch be enough ?
```
on:
  workflow_dispatch:
    branches:
      - main
 ```

pgetta: I'm not sure it's supported. My linter complains that there's no such property as `branches` under the `workflow_dispatch`, it seems there are bunch of discussions which use the same approach: https://github.com/orgs/community/discussions/137079#discussioncomment-11083116.

At the same time the official documentation: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch  doesn't have anything explicitly saying one can do that.

bilalesi: ahh, i think i had this discussion before about it for fusion, so i still have it wrong in my head, thanks for the info

pgetta:
> if we want to tag the deployment with last commit hash
> 
> ```shell
> - name: tag ecs deploy
>         run: |
>           aws ecs tag-resource \
>             --resource-arn ${{ vars.AWS_STAGING_ECS_SERVICE }} \
>             --tags key=deployment_id,value=${{ github.sha }}
> ```

The `vars.AWS_STAGING_ECS_SERVICE` is a name, not an ARN. 

However we can assemble one. It has the following pattern: `arn:aws:ecs:region:aws_account_id:service/cluster-name/service-name` so what we are missing is only the `aws_account_id`.

I propose to address this separately (as well as providing new tag message for production releases). We will need to check with Dries if the current credentials allow resource tagging.

bilalesi:
> I propose to address this separately (as well as providing new tag message for production releases). We will need to check with Dries if the current credentials allow resource tagging.

Agree 👍

bilalesi (approved):

PR #88, opened 2025-02-25
Make sure model .hasPart property is treated correctly

pgetta:
As from Nexus we can receive a single object or an array of objects (depending on the "set" length).


g-bar:
LGTM.

pgetta:
Thanks!

PR #91, opened 2025-02-26
Show real balances from API

pgetta:
The current change:
* Removes current balance component which has number of usability issues.
* Adds rendering of the balance to VLab/Project details banner.

Before:
before

After:
after

Removed content is highlighted in red, added - in green.

### DIFF
--- a/src/components/VirtualLab/projects/VirtualLabProjectList/VirtualLabProjectItem.tsx
+++ b/src/components/VirtualLab/projects/VirtualLabProjectList/VirtualLabProjectItem.tsx
@@ -1,6 +1,11 @@
-import { CalendarOutlined, LoadingOutlined, UserOutlined } from '@ant-design/icons';
+import {
+  CalendarOutlined,
+  CreditCardOutlined,
+  LoadingOutlined,
+  UserOutlined,
+} from '@ant-design/icons';
 import Link from 'next/link';
-import { loadable, unwrap } from 'jotai/utils';
+import { loadable } from 'jotai/utils';
 import { useAtomValue } from 'jotai';
 import { Spin } from 'antd';
 
@@ -10,6 +15,8 @@ import { Project } from '@/types/virtual-lab/projects';
 import { formatDate } from '@/util/utils';
 import { generateVlProjectUrl } from '@/util/virtual-lab/urls';
 import { virtualLabProjectUsersAtomFamily } from '@/state/virtual-lab/projects';
+import { useLastTruthyValue } from '@/hooks/hooks';
+import { virtualLabBalanceAtomFamily } from '@/state/virtual-lab/lab';
 
 function MemberAmount({ virtualLabId, projectId }: { virtualLabId: string; projectId: string }) {
   const users = useAtomValue(
@@ -28,14 +35,18 @@ function MemberAmount({ virtualLabId, projectId }: { virtualLabId: string; proje
 function ProjectStats({ project }: { project: Project }) {
   const { created_at: createdAt, id: projectId, virtual_lab_id: virtualLabId } = project;
 
-  const projectUsers = useAtomValue(
-    unwrap(
-      virtualLabProjectUsersAtomFamily({
-        virtualLabId,
-        projectId,
-      })
-    )
+  const projectUsers = useLastTruthyValue(
+    virtualLabProjectUsersAtomFamily({
+      virtualLabId,
+      projectId,
+    })
   );
+
+  const virtualLabBalance = useLastTruthyValue(virtualLabBalanceAtomFamily({ virtualLabId }));
+
+  const projectBalance = virtualLabBalance?.data.projects?.find(
+    (pr) => pr.proj_id === projectId
+  )?.balance;
 
   const iconStyle = { color: '#69C0FF' };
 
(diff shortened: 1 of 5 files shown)


pgetta:
FYI, @ayimaok, I'll let you know once this is deployed to staging.

src/components/VirtualLab/projects/VirtualLabProjectList/VirtualLabProjectItem.tsx
```diff
@@ -28,15 +35,19 @@ function MemberAmount({ virtualLabId, projectId }: { virtualLabId: string; proje
...
-      })
-    )
+  const projectUsers = useLastTruthyValue(
+    virtualLabProjectUsersAtomFamily({
+      virtualLabId,
+      projectId,
+    })
   );
 
+  const virtualLabBalance = useLastTruthyValue(virtualLabBalanceAtomFamily({ virtualLabId }));
```
g-bar, line 45: Could you explain `useLastTruthyValue`. Under what circumstances does a previously truthy value become false and why do we want to fallback to it? and not use the latest value?

pgetta: Good question. Short answer: when we don't want to see `undefined` while the value of underlying atom is getting re-evaluated (because of a change in dependent atom, for example).

Here is an example:
```
'use client';

import { atom, useAtomValue, useSetAtom } from 'jotai';
import { unwrap } from 'jotai/utils';

function fetchData(): Promise<string> {
  return new Promise((res) => {
    setTimeout(() => {
      res('fetched!');
    }, 2000);
  });
}

const refetchTriggerAtom = atom<number>(0);

const dataAtom = atom(async (get) => {
  get(refetchTriggerAtom);
  return fetchData();
});

export default function TestPage() {
  const data = useAtomValue(unwrap(dataAtom));
  const refetch = useSetAtom(refetchTriggerAtom);

  return (
    <div>
      <h1>Test Page</h1>
      <p>Data: {data}</p>
      <button type="button" onClick={() => refetch((prev) => prev + 1)}>
        Reset
      </button>
    </div>
  );
}
```

With unwrap/loadable-useLoadable when a re-fetch is triggered, until the "new" result arrives data becomes `undefined`. This is the case also with `atomWithRefresh`.

`useLastTruthyValue` returns previous value that is not `undefined` (if ever existed) until a new one arrives. This is handy when we don't want to see flashing content when we initiate a re-fetch.

g-bar: Isn't that the same as:
https://github.com/openbraininstitute/core-web-app/blob/0c71d1ac9606b7bf40083152cb381af6b74e4a7a/src/hooks/hooks.ts#L22

unwrapped is the same as loadable but drops the 'state' and just keeps the 'data', 
useLoadable just return the data so they're functionally identical. Is that not the case?

g-bar: Also unwrap seems to have a fallback option that keeps the previous value. https://jotai.org/docs/utilities/async#unwrap

Maybe is possible to use that and replace both `useLoadable` and `useLastTruthyValue`?

pgetta: Yeah, they are very similar.

With `useLoadable` one has to provide a loadable atom, but with the new method this is not needed.

At the end the difference is minor, only readability:
`useLastTruthyValue(atom)` vs. `useLoadable(loadable(atom))` (unless the loadable atom is created somewhere outside). I guess there is less overhead by using unwrap, but it's probably almost nothing.

pgetta: Oh, nice, I didn't know about the fallback.

Yeah, in this case we can replace all of these. how about a single method, like below?

```
function useUnwrappedValue(atom) {
  return unwrap(atom, prev => prev);
}
```

So that in components we just wrap an atom in a call to useUnwrappedValue and that's it.

pgetta: Maybe we can come up with a better (and shorter) naming.

g-bar: Yeah sounds good to me.

Can't think of a good name, I was thinking of something the reflects the fact that we're are showing a 'stale' value while it 'revalidates', because that's what this ultimately is.

Also for the future, I was thinking of some hook / utility that does this but also refreshes the atom on mount (or optionally after an interval), because a problem with our current approach is that atom values can become stale unless we refresh them explicitly. But that's an improvement for later, this sounds good in the meantime.

pgetta: Great, I'll prepare a PR in coming days then. Replacing `useLoadable` will require some refactoring as not in all the places the loadable atom is created inline.

g-bar (approved):

pgetta:
Thank you, Gil.

pgetta:
@ayimaok it's deployed

ayimaok:
@pgetta, thanks! I will check, most likely tomorrow.

PR #101, opened 2025-03-03
Open notebook spec link in new tab

g-bar:


src/middleware.ts
```diff
@@ -1,7 +1,7 @@
 import nextAuthMiddleware, { NextRequestWithAuth } from 'next-auth/middleware';
 import { NextRequest, NextResponse } from 'next/server';
 
-const PRIMARY_HOSTNAME = process.env.PRIMARY_HOSTNAME;
+const { PRIMARY_HOSTNAME } = process.env;
```
g-bar, line 4: This was prettier and eslint I think :).

pgetta: I'll try to spare some time to fix CI so this will not happen again. Will see what we do with tests, as @bilalesi is doing major refactoring of explore section, maybe after he is done - we can put some effort to fix those.

pgetta (approved):

g-bar:
Thanks Pavlo.

PR #107, opened 2025-03-03
store user journey indexdb for literature

bilalesi:
add indexdb store to save user interaction with brain region and artifacts and also the type of data

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,6 +55,7 @@
         "d3-sankey": "^0.12.3",
         "date-fns": "^3.6.0",
         "deepdash-es": "^5.3.9",
+        "dexie": "^4.0.11",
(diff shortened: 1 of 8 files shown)


tolokoban (dismissed):
Good job Bilal!

bilalesi:
> Not sure I understand, so we store that information but how/where it's going to be used?

ahh sorry did not give context, so it was a meeting yesterday, and new feature  will be implemented,
we are getting rid of literature for the sake of the new feature,
we track the use clicks on explore section and send it to LLM to have context of user exploring
currently we are only interested on parts in the pic,
the first step is to save the action in the browser and later when we have a PoC we think of moving  to server persistence 

Screenshot 2025-03-03 at 20 07 13

pgetta (approved):
Nice!

PR #114, opened 2025-03-06
Homepage popup

g-bar:
Capture d’écran 2025-03-07 à 11 36 52


jdcourcol:
Thanks for implementing it so fast. @g-bar 
Unfortunately, after discussing with @gkhazen , I thought a modal would be too intrusive for the end user (and would probably prevent SEO to enter the site).
Georges agreed to have some slide-in notification instead, to drive the user to the registration page. @loris-olivier-obi would you be able to update the mockup w/ that

PR #115, opened 2025-03-07
Pre render LandingPage for SEO

g-bar:
Currently the LandingPage fetches the content from sanity on the client, which means it's not getting pre-rendered which is not optimal for SEO.

To check this **try disabling javascript** on your browser and open the app on `develop` branch
No content gets displayed other than the menu and footer which are rendered in server components.
[
Capture d’écran 2025-03-07 à 09 00 33
](url)

Switch to this branch and reload.
Text renders for that SEO benefit.

Capture d’écran 2025-03-07 à 09 03 37


src/components/LandingPage/content/content.ts
```diff
@@ -22,26 +23,29 @@ export const urlFor = (source: SanityImageSource) => {
...
-      .catch((ex) => {
-        logError('There was an exception in this Sanity query:', query);
-        logError(ex);
-        setData(null);
-      });
-  }, [query, typeGuard]);
+export function useSanity<T>(query: string, typeGuard: (data: unknown) => data is T) {
+  const data = useAtomValue(dataFamily({ query, typeGuard })) as T | undefined | null;
   return data;
 }
```
g-bar, line 50: This was the main culprit. useEffect only runs client side, jotai's useAtom runs server and client side.

bilalesi: nice job gil
i am not sure i understand the last part "jotai's useAtom runs server and client side."

PR #125, opened 2025-03-12
First version of the chat is ready

tolokoban:
First shot to let people play round and give feedbacks.

For now, we aggregate the result of two tools for the papers: **web-search** and **literature-search**.


loris-olivier-obi (approved):
Tout va bien! Merci

PR #133, opened 2025-03-14
Added widget multiple button

tolokoban:
Added support for a new Sanity widget: multiple button.

image


jdcourcol:
@gkhazen is telling me there is a change in the plan. The real digital brain story is a new page and Georges will send the text.

jdcourcol:
I don't know for the OBI background thing.

tolokoban:
Ok. But, anyway, we still need this PR to be merged, because it allows to use the news Sanity structure.
That does not means that the buttons will be in the website. We control this in Sanity.

loris-olivier-obi (approved):
Great PR

PR #143, opened 2025-03-17
add purchases history

bilalesi:


src/components/VirtualLab/VirtualLabSettingsComponent/purchases-table.tsx
```diff
@@ -0,0 +1,133 @@
...
+  onPageChange,
+  loading = false,
+}: PaymentsTableProps) {
+  const columns: ColumnsType<SubscriptionPaymentDetails> = [
+    {
+      title: 'Credits',
+      key: 'credits',
+      width: 'max-content',
+      render: (_, record) => {
+        const credits = Math.round(record.amount_paid / 100 / CONVERSION_RATE);
```
pgetta, line 34: We will need to test this thoroughly, I have an impression we might get some discrepancies (+/- 1 credit) due to the use of floating point (with the actual value that was computed/applied by vlab api)

pgetta (approved):

src/components/VirtualLab/VirtualLabSettingsComponent/purchases-table.tsx
bilalesi: agree 👍

PR #149, opened 2025-03-18
use username as virtual lab default name on creation

bilalesi:


src/components/VirtualLab/create-entity-flows/virtual-lab/content.tsx
```diff
@@ -91,7 +93,7 @@ export default function CreateVirtualLabForm() {
           requiredMark={false}
           validateTrigger={['onChange']}
           initialValues={{
-            name: '',
+            name: `${data?.user.name}'s virtual lab`,
```
pgetta, line 96: I guess it's fine for now, but in future there might be conflicts as from what I recall we require the name to be unique. Not sure what to do in this case: maybe triggering validation with a default value on init?

pgetta (approved):

PR #153, opened 2025-03-18
base entity core migration

bilalesi:


bilalesi:
Please Note that this branch will keep have `Lint` issues, since the migration will be progressively and can keep both NEXUS and ENTITY_CORE 

@g-bar , @pgetta, @tolokoban

src/api/apiClient.ts
```diff
@@ -0,0 +1,415 @@
...
+      }
+
+      return true;
+    } catch (e) {
+      console.error('Failed to clear cache:', e);
+      return false;
+    }
+  }
+
+  get<T>(
```
tolokoban, line 367: This function relies of a `as unknown as T` casting that is made in the `_request` method.
This can be unsafe because we have no warranty the type will always be the one expected.

We could do something like that:

```ts
  get<T>(
    endpoint: string,
    typeGuard: (data: unknown) => asserts data is T,
    options?: RequestOptions,
    config?: RequestConfiguration & { cache?: CacheConfiguration }
  ) {
    const data = this._request<T>('get', endpoint, options, config) as unknown;
    typeGuard(data);
    return data;
  }
```

If you need to get a person, for instance, you can then do this:

```ts
api.get<IPerson>(url, assertIPerson)
```

tolokoban (commented):
I will continue to review this tomorrow.
You did a terrific job. Well done Bilal!

src/api/entitycore/guards.ts
```diff
@@ -0,0 +1,13 @@
+import type { IOrganization, IPerson } from '@/api/entitycore/types/shared/global';
+
+export function isOrganization(agent: unknown): agent is IOrganization {
+  return (
+    typeof agent === 'object' && agent !== null && 'pref_label' in agent && !('givenName' in agent)
+  );
+}
+
+export function isPerson(agent: unknown): agent is IPerson {
```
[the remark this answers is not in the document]
g-bar: I wonder if we could use zod schemas instead of type casts. Like that we get validation and get the typescript type too with ReturnType.

tolokoban (approved):
Great job!

src/api/apiClient.ts
bilalesi: do we really want to go so hard on typing, 
even without a type guard, an incorrect response will likely cause an error downstream when the data is used incorrectly.
type guards add an explicit validation step, but they don’t prevent runtime error, it just shift it when they happen.
and also strict validation will introduce potential overhead perf, if we applied it frequently and also slowdown the dev.
in the other hand typescript already provide decent protection, so i think we are good now ?

PR #154, opened 2025-03-18
Remove currency label

tolokoban:
We removed the "currency" label before the combo box.
And we added a "Contact Us" button for the Premium Plan.


loris-olivier-obi (approved):
"Sanity m'a tué!"

PR #155, opened 2025-03-19
First draft for AI

tolokoban:
image


loris-olivier-obi (approved):
Brain region side bar is life!

PR #156, opened 2025-03-19
47/ai literature search tool

tolokoban:
The output of the literature-search tool will soon change.
We had to adapt the code to prevent a compatibility break.


loris-olivier-obi (approved):
"Approuvée mais pas mergée" Charles de Gaulle

PR #160, opened 2025-03-20
Publc project V1

loris-olivier-obi:
Implementation with Sanity content
Public Project / Showcase on home
Implementation of three section: description, artifacts and notebooks


package.json
```diff
@@ -103,6 +104,7 @@
     "react-plotly.js": "^2.6.0",
     "react-select": "^5.7.3",
     "react-syntax-highlighter": "^15.6.1",
+    "react-table": "^7.8.0",
```
tolokoban, line 107: It seems that this is version 7 of `react-tool` and you already imported version 8 in `@tanstack/react-tool`.
I think we can delete this line.

src/api/sanity/client.ts
```diff
@@ -1,10 +1,12 @@
 import SanityClient from 'next-sanity-client';
+
 import tiersQuery from './tiers.query.groq';
+
 import { ContentForPricing } from '@/components/LandingPage/content/pricing';
 
 const client = new SanityClient({
   projectId: 'fgi7eh1v',
-  dataset: 'production',
+  dataset: 'staging',
```
tolokoban, line 9: Are you still using this to access Sanity?
The issue here is that we have `staging` that is hardcoded.
The `useSanity` hook takes care of this. I don't think we need this `/api/sanity`.

src/api/sanity/sanity.ts
```diff
@@ -0,0 +1,16 @@
+import imageUrlBuilder from '@sanity/image-url';
+import { SanityImageSource } from '@sanity/image-url/lib/types/types';
+import { createClient } from 'next-sanity';
+
+export const client = createClient({
+  projectId: 'fgi7eh1v',
+  dataset: 'staging',
```
tolokoban, line 7: Are you still using this to access Sanity?
The issue here is that we have `staging` that is hardcoded.
The `useSanity` hook takes care of this. I don't think we need this `/api/sanity`.

tolokoban (changes requested):

src/api/sanity/client.ts
loris-olivier-obi: @tolokoban There is someone else using it in the virtual lab. It's probably Bilal. I left it in case it's critical.

tolokoban:
We will redo a PR, because we had merge conflicts that we resolved locally.

PR #165, opened 2025-03-21
Corrections public project number 3

loris-olivier-obi:
Rename Showcases by Public Projects


tolokoban (approved):
Nothing to add. Looks good!

PR #169, opened 2025-03-24
Make me-model validation more reliable

pgetta:
These are mostly workarounds to make me-model validation work:
* Add websocket ping messages so that AWS ALB do not close the connection due to reaching inactivity threshold.
* Refresh access token before starting the validation - eventually this will have to be replaced by access token management in the service itself (separate KC client with token exchange capability and longer access token TTL).
* Set current API Gateway URL.


bilalesi (approved):
LGTM

PR #171, opened 2025-03-24
allow user to change role

bilalesi:
allow remove member

listing projects new design

don't allow add member on free subscription


src/hooks/useUserRole.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { UserGroup } from '@/api/virtual-lab-svc/queries/types';
+import { tryCatch } from '@/api/utils';
+
+type Props = {
+  virtualLabId?: string;
+  projectId?: string;
+};
+
+export default function useUserRole({ virtualLabId, projectId }: Props) {
+  const [userGroups, setUserGroups] = useState<Array<UserGroup>>([]);
```
pgetta, line 15: As this component might be used in multiple places, even in a single page, probably worth moving to a global state, maybe an atom, to avoid mulitple/constant re-fetching.

bilalesi: actually the idea is to make it refetch every-time, we can not know exactly when the user change the role (event the vlab-svc will cause an error but i think the UX is important for this kind of things)

pgetta (approved):
Looks good!

I've added a minor comments, but we can address those if needed, in a separate PR.

src/api/virtual-lab-svc/queries/stats.ts
```diff
@@ -0,0 +1,92 @@
+import {
+  VlmProjectStatsResponse,
+  VlmUserStatsResponse,
+  VlmVirtualLabStatsResponse,
+} from '@/api/virtual-lab-svc/queries/types';
+import { getSession } from '@/authFetch';
+import { virtualLabApi } from '@/config';
+
+const BASE_URL = `${virtualLabApi.url}/virtual-labs`;
+// const BASE_URL = `http://localhost:8000/virtual-labs`;
```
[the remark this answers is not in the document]
bilalesi: thanks for the catch

src/hooks/useUserRole.tsx
pgetta: My main concern was multiple-fetches on the page load.

If the data is getting re-fetched on the user action, when, for example, they open some menu or navigate to a different page - that's totally fine.

With the current implementation if there are multiple components that use the hook - the same request will be sent multiple times and this is something worth avoiding. And I can imagine this hook will be used everywhere (in certain pages).

PR #173, opened 2025-03-24
Display userfriendly eror when the rate has been exceeded.

tolokoban:
* Preventing the user to ask another question if the previous one is still pending.
* Display an understandable error message when the user has exceeded their rate limit.

image


loris-olivier-obi (approved):
Nice public release

PR #177, opened 2025-03-25
Prepare for release 2025.03.25.1

pgetta:


src/api/sanity/client.ts
```diff
@@ -1,10 +1,12 @@
 import SanityClient from 'next-sanity-client';
+
 import tiersQuery from './tiers.query.groq';
+
 import { ContentForPricing } from '@/components/LandingPage/content/pricing';
 
 const client = new SanityClient({
   projectId: 'fgi7eh1v',
-  dataset: 'production',
+  dataset: 'staging',
```
tolokoban, line 9: Are you sure we want to hardcode `staging` here?

tolokoban: Yes, but here I see `data: 'staging'` instead of `dataset: process.env.NEXT_PUBLIC_SANITY_DATASET`.

tolokoban (approved):

src/api/sanity/client.ts
pgetta: I've reverted this change back to `production` and submit another PR to properly parameterise this (but that will go via develop first).

PR #187, opened 2025-03-26
Fix invite redirects and non-pro subscription status

pgetta:
Few things:
* switched to use subscription instead of user session (as the session is not automatically update - will fix that later)
* subscription screen is pretty busy, so I've moved the message about pro subscription requirement to the invite page itself.


bilalesi (approved):
LGTM
let's invite each other in staging to test 😉

PR #191, opened 2025-03-26
Rate limit is also applied on suggestions, now

tolokoban:
If the user ask for too many suggestions, we internally get a 429 HTTP error that we silently ignore.
This way, the user won't spend AI money, but they will still see the hard-coded suggestions.


loris-olivier-obi (approved):
Le silence est d'or comme le disait si bien Marie-France Garaud

PR #206, opened 2025-03-27
Fix articles list when an abstract is null

tolokoban:


WonderPG (approved):
LGTM thank you !

loris-olivier-obi (approved):
Approved

PR #211, opened 2025-03-28
Do not show simulate button in Explore - synaptome model details

pgetta:
As there is no vlab/project context and the link doesn't work with such conditions.

FYI, @ayimaok


bilalesi (approved):

pgetta:
Thanks, Bilal

PR #215, opened 2025-03-28
Made all the rows expanded by default

loris-olivier-obi:


tolokoban (approved):

pgetta:
@loris-olivier-obi if you can resolve the conflict we can deploy this to prod as well with https://github.com/openbraininstitute/core-web-app/pull/218

PR #222, opened 2025-03-28
Layout circuit last minute hack

g-bar:


src/components/explore-section/ExploreInteractive/DataTypeGroupTotals.tsx
```diff
@@ -87,7 +88,7 @@ export default function DataTypeGroupTotals({
 
       {dataTypeGroup === DataTypeGroup.ModelData && (
         <StatItem
-          href={`${pathName}/model/circuit`}
+          href={`${rootbasePath}/app/virtual-lab/explore/interactive/model/circuit`}
```
bilalesi, line 91: do we really need the root base path, would /app/virtual ... be enough ?

bilalesi (approved):

src/components/explore-section/ExploreInteractive/DataTypeGroupTotals.tsx
g-bar: There was an issue with that url, this is a hack, will change.

src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/explore/interactive/(data)/model/circuit/layout.tsx
```diff
@@ -0,0 +1,24 @@
...
+export default function ExploreCircuitListingView({ children }: { children: ReactNode }) {
+  const params = useParams();
+
+  if (params?.id)
+    return <ErrorBoundary FallbackComponent={SimpleErrorComponent}>{children}</ErrorBoundary>;
+
+  return (
+    <div className="flex h-full w-full bg-primary-9" id="interactive-data-layout">
+      <ErrorBoundary FallbackComponent={SimpleErrorComponent}>
+        <div className="h-full w-full overflow-y-hidden overflow-x-scroll bg-white text-white">
```
Comment (pgetta) [inline src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/explore/interactive/(data)/model/circuit/layout.tsx:18]: `overflow-y-hidden`, just to confirm, we don't want scroll activated here, is that right?

There was an issue with circuit listing where scroll didn't work due to such class, fixed in other PR.

Comment (g-bar) [reply src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/explore/interactive/(data)/model/circuit/layout.tsx:18]: Yes, there are lots of issues, that was not the cause of that, we have to do a thourgh review of these layouts, it's a big mess.

PR #232, opened 2025-04-01
Fix the fact that last step in user journey was missing

tolokoban:


g-bar (approved):

g-bar:
Sorry @tolokoban I restored the package-lock because I thought I was by accident. Was that intended?

tolokoban:
I don't know what's going on. I just erased my `package-lock.json` and ran `npm install`. It should put the correct versions in it, right?

g-bar:
I ran 
`git checkout develop package-lock.json`

To get a working install without modififying the lockfile you can run.

`
npm ci
`

g-bar (approved):

PR #233, opened 2025-04-02
Hardcoded questions for when the user is on the top of the brain hierarchy

tolokoban:
This PR relates to this issue: https://github.com/openbraininstitute/prod-explore-functionality/issues/72

### DIFF
--- a/src/components/literature-suggestions/suggested-questions/hardcoded-suggestions/hardcoded-suggestions.ts
+++ b/src/components/literature-suggestions/suggested-questions/hardcoded-suggestions/hardcoded-suggestions.ts
@@ -1,18 +1,21 @@
 import React from 'react';
 import { useSnapshot } from './snapshot';
-import { SuggestionsList } from './suggestions-list';
+import { SuggestionsListPerRegion, SuggestionsListFullBrain } from './suggestions-list';
 import { isString } from '@/util/type-guards';
 
 export function useHardcodedSuggestions(): string[] {
   const snapshot = useSnapshot();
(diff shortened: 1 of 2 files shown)


src/components/literature-suggestions/suggested-questions/hardcoded-suggestions/hardcoded-suggestions.ts
```diff
@@ -1,18 +1,21 @@
...
-import { SuggestionsList } from './suggestions-list';
+import { SuggestionsListPerRegion, SuggestionsListFullBrain } from './suggestions-list';
 import { isString } from '@/util/type-guards';
 
 export function useHardcodedSuggestions(): string[] {
   const snapshot = useSnapshot();
   const [suggestions, setSuggestions] = React.useState<string[]>([]);
   React.useEffect(() => {
-    const questions = shuffle(SuggestionsList[snapshot?.artifact ?? ''])
+    const isBrainRoot = snapshot.region.toLowerCase().startsWith('basic cell');
```
bilalesi, line 10: I suggest to compare the root by it's `ID` not by `words`
it may need to change the snapshot hook to include the id too

bilalesi (approved):
tiny suggestion if you see it important

src/components/literature-suggestions/suggested-questions/hardcoded-suggestions/hardcoded-suggestions.ts
tolokoban: Good point. I'll create another PR for this.

PR #234, opened 2025-04-02
Use the region ID to know if it's the root

tolokoban:
We now use the region ID to know if we are at the top of the hierarchy.

### DIFF
--- a/src/components/literature-suggestions/suggested-questions/hardcoded-suggestions/snapshot.ts
+++ b/src/components/literature-suggestions/suggested-questions/hardcoded-suggestions/snapshot.ts
@@ -10,8 +10,6 @@ interface Snapshot {
   artifact: string;
 }
 
-export const ROOT_REGION_ID = 'http://api.brain-map.org/api/v2/data/Structure/8';
-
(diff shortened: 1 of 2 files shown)


src/components/literature-suggestions/suggested-questions/hardcoded-suggestions/snapshot.ts
```diff
@@ -5,22 +5,27 @@ import { selectedBrainRegionAtom } from '@/state/brain-regions';
...
 
 interface Snapshot {
-  region: string;
+  regionId: string;
+  regionTitle: string;
   artifact: string;
 }
 
+export const ROOT_REGION_ID = 'http://api.brain-map.org/api/v2/data/Structure/8';
+
```
bilalesi, line 14: it's already here `BASIC_CELL_GROUPS_AND_REGIONS_ID`

tolokoban: Good to know!
I have pushed a fix.

bilalesi (approved):

PR #235, opened 2025-04-02
Various ME-model validation UI improvements

pgetta:
Changes:
* Provide vlab context (org/proj) to image viewer, so that it can fetch images from non-default/public projects.
* Update service URL.
* Correct elapsed time formatting.
* Show validation status labels under the ME-model details page.
* Show a message to the user when the window can be closed without interrupting running analysis.


src/components/build-section/virtual-lab/me-model/AnalysisPreview.tsx
```diff
@@ -3,21 +3,41 @@ import { useAtomValue } from 'jotai';
...
-        No ME-Model analysis yet
+        {message}
       </div>
     );
   }
 
-  return (
-    <PDFViewerContainer distributions={ensureArray(image) as { '@id': string; about: string }[]} />
-  );
+  return <PDFViewerContainer distributions={distributions} />;
```
bilalesi, line 42: i think the `distributions={ensureArray(image)` is the right use, it was a change from Gil, since the analysis is now stored in `image` property rather then the distribution

src/components/build-section/virtual-lab/me-model/BluePyEModelContainer.tsx
```diff
@@ -20,6 +19,20 @@ function getMEModelPageUrl(meModelId: string, virtualLabInfo: VirtualLabInfo) {
...
+  const minutes = Math.floor((seconds % 3600) / 60);
+  const remainingSeconds = seconds % 60;
+
+  return [
+    hours.toString().padStart(2, '0'),
+    minutes.toString().padStart(2, '0'),
+    remainingSeconds.toString().padStart(2, '0'),
+  ].join(':');
+};
+
```
bilalesi, line 35:
```
import { format } from 'date-fns';

function msToTime(ms) {
  return format(new Date(ms), 'HH:mm:ss');
}
```
but date-fns   will take it as an absolute timestamp  "number of milliseconds since 01/01/1970 ", not sure will cause an issue (probably not)

.env.development
```diff
@@ -11,7 +11,7 @@ NEXT_PUBLIC_CELL_SVC_BASE_URL=https://staging.openbraininstitute.org/api/circuit
 NEXT_PUBLIC_KG_INFERENCE_BASE_URL=https://staging.openbraininstitute.org/api/kg-inference
 NEXT_PUBLIC_THUMBNAIL_GENERATION_BASE_URL=https://staging.openbraininstitute.org/api/thumbnail-generation
 NEXT_PUBLIC_SYNTHESIS_URL=https://synthesis.sbo.kcp.bbp.epfl.ch/synthesis-with-resources # TODO: change to staging
-NEXT_PUBLIC_ME_MODEL_ANALYSIS_WS_URL=wss://kjyo1qi5af.execute-api.us-east-1.amazonaws.com/prod/
+NEXT_PUBLIC_ME_MODEL_ANALYSIS_WS_URL=wss://8sjpv70kgl.execute-api.us-east-1.amazonaws.com/prod/
```
g-bar, line 14: Just out of curiosity why can't we have something like staging.openbraininstitute.org/api/me-model-analysis?

src/components/build-section/virtual-lab/me-model/BluePyEModelContainer.tsx
```diff
@@ -60,8 +73,8 @@ function ValidationInit({
         Initiating validation
       </h2>
       <p className="text-primary-8">
-        Once the validation process finishes, you can view the results in your project&apos;s
-        Activity section.
+        Once the validation process finishes, you can view the results in your project's Activity
```
bilalesi, line 76: `project's activity` 
may be the linter will yell at you to use this "&apos;"

src/components/build-section/virtual-lab/me-model/AnalysisPreview.tsx
bilalesi: @g-bar  may confirm regading this one

src/components/build-section/virtual-lab/me-model/AnalysisPreview.tsx
```diff
@@ -3,21 +3,41 @@ import { useAtomValue } from 'jotai';
...
-import { ensureArray } from '@/util/nexus';
+import { ensureArray, getOrgFromSelfUrl, getProjectFromSelfUrl } from '@/util/nexus';
+import { MEModel } from '@/types/me-model';
+
+const statusMessage: Record<MEModel['status'], string> = {
+  initialized: 'No ME-Model analysis yet',
+  running: 'ME-Model analysis is running',
+  done: 'ME-Model analysis done',
+  failed: 'ME-Model analysis failed',
+};
```
g-bar, line 14: In entitycore we have different values we should sync them for when we migrate:

https://github.com/openbraininstitute/entitycore/blob/8b5d992ef56d72bf45a14310de862437c42f4b58/app/db/model.py#L417C1-L422C20

But most other than initialized and done weren't used in the past, are they used now?

src/components/build-section/virtual-lab/me-model/AnalysisPreview.tsx
g-bar: Yes, indeed couldn't test this when i made that change.

.env.development
pgetta: The on-demand service, used by the me-model validation uses AWS API Gateway under the hood, and that is its URL/endpoint. A custom domain name can be added to API Gateway, but might be not trivial and never had chance to look into that.

The current setup has a flaw, where, if for whatever reason, API Gateway is recreated, its' URL will be different and that will require manual updates to the core web app.

A solution would require either: implementing a custom domain or adding a proxy in core-web-app (for example) that will route traffic to correct API Gateway URL which can be provided as a run-time env var by terraform.

src/components/build-section/virtual-lab/me-model/BluePyEModelContainer.tsx
pgetta: That was how the previous implementation was done, something like `format(now - startedAt, 'HH:mm:ss')` and it was showing `01:00:00` at the very start (one extra hour).

I've looked at the date-fns documentation and found only [formatDistance](https://date-fns.org/v4.1.0/docs/formatDistance) method somewhat close to what we need - but it only can give a result in form of words, the `HH:mm:ss` type of formatting can't be applied.  Thus - the current change.

src/components/build-section/virtual-lab/me-model/BluePyEModelContainer.tsx
pgetta: Thanks, missed that. It's fixed.

src/components/build-section/virtual-lab/me-model/AnalysisPreview.tsx
pgetta: @bilalesi That distribution value is defined above as:
```
const distributions = ensureArray(meModelResource?.image)
  .map(... // + extra props for org/proj context
```

So I'm not sure, what is your suggestion?

bilalesi: ahh sorry, bad call, did not capture that, all good, i thought you were using `distributions`

g-bar: I saw that you added org, proj to the type
`type Distribution = { '@id': string; about: string; org?: string; project?: string };`

but the PDFViewer container doens't use that at all is that type used somewhere else.
Probably should be changed to type Image = { '@id': string; about: string ??

src/components/build-section/virtual-lab/me-model/AnalysisPreview.tsx
pgetta: Somehow `failed` feels more natural to me along with `created` or `initialized`, but I guess it's a personal taste.
Indeed, we'll need to sync them on migration.

All these statuses were implemented in the service (there still pending [change](https://github.com/openbraininstitute/me-model-analysis/pull/1) to enable some of them) and used in the frontend.

I'm not sure why, but the `initialized` status is basically the one which means "created", I didn't touch it because we have already some models with this value, but I think we can drop it since there is no distinct use for it, is there?

g-bar: Created, user saved the MEModel,
Initialized user has called the validation service.

Was my thinking, we can change 'error' to 'failed'.

src/components/build-section/virtual-lab/me-model/AnalysisPreview.tsx
pgetta: PDFViewerContainer is not using that directly, yes, but it passes these props down to PDFViewer component and that one [uses the org/project context](https://github.com/openbraininstitute/core-web-app/blob/fix/me-model-validation-results/src/components/explore-section/common/pdf/PDFViewer.tsx#L46).

Before, it was always using the default/public nexus project, but that doesn't work with user generated data.
So that's why I've added these optional properties - to provide custom org/project context.

pgetta: Good thing is that for entitycore, if I understand correctly how it works, only having an ID will be enough.

g-bar: Ah, yes yes, all good then.

g-bar: > Good thing is that for entitycore, if I understand correctly how it works, only having an ID will be enough.

My understanding id that for these image arrays it will just be in the `assets` of the entity, and we will get the se path directly from there.

g-bar (approved):

src/components/build-section/virtual-lab/me-model/AnalysisPreview.tsx
pgetta: Good to know, thanks!

pgetta:
Thank you all.

PR #237, opened 2025-04-02
69/ai rate limit better message

tolokoban:
We added a "Cancel" button to interrupt a long AI response.
And we improved the error message displayed when the rate limit has been exceeded.


g-bar (approved):
LGTM

PR #240, opened 2025-04-03
Better me-model validation UX

pgetta:
* Show spinner only on a button clicked by the user, and just disable the rest.
* Do not text wrap the me-model link.


tolokoban (approved):
Looks good to me.

pgetta:
Thank you Fabien

PR #252, opened 2025-04-07
Prevent Filter panel from hidding the AI panel

tolokoban:
image


loris-olivier-obi (approved):
Don't hide the panel!

PR #256, opened 2025-04-08
Use current session to determine a state when refresh fails

pgetta:
Due to [some optimisations](https://github.com/openbraininstitute/core-web-app/blob/develop/src/hooks/session.ts#L71) to avoid unnecessary re-renders on session change, in order to detect token refresh failures we should rely on the session directly from the next-auth lib, unlike it is currently implemented.

Addresses https://github.com/openbraininstitute/prod-build-emodel/issues/4.


pgetta:
Thank you Gil.

PR #257, opened 2025-04-08
Make explore list span available space instead of full window height

pgetta:
Addresses https://github.com/openbraininstitute/prod-explore-functionality/issues/89#issuecomment-2786266974


pgetta:
Previous implementation had menu + list view where list view's height was set to 100% and after cut by overflow hidden, now list view spans available space instead.

PR #262, opened 2025-04-10
Upgrade deps

pgetta:


pgetta:
It seems I've got most of it working, but still need to fix some minor things - will do that in coming days.

bilalesi:
cool, thanks pavlo for the PR

PR #267, opened 2025-04-14
change ion to coming soon/disable autodeploy staging gh

bilalesi:
1. disable auto-deploy for staging
2. change ion to coming soon


pgetta:
Looks good. @bilalesi but why do we disable autodeploy in staging?

pgetta (approved):

PR #269, opened 2025-04-15
add bookmarking features

bilalesi:
1. fix antd lib (antd not yet support react 19, there multiple fixes, but still things may not work on `--turbopack` mode)
2. add bookmarking features 
3. fix the bookmark page design
4. fix some issues for SSR
5. refactor components and utils fn

**Feature update**
1. `Bookmark` refactored as a feature under `features`folder
2. The feature need some **more** refactoring after `type configuration` will be done.
3. The feature implemented only on explore pages, for `build` and `simulate` will come later when we have BE/data deployed.

**What has been changed for ANTD:**
1. using  `modal`, `notification`, `alert`, should be instantiated from `App` context (sometime does not work with turbopack)
5. two providers has been add to the `providers` file (`AntdRegistry` and `App`) 
6. the `hashed` property for the theme is `disabled`, it will give us freedom to override the styles from antd (previously was disabled on-need)

**What's useful from `Next 15` so far:**
1.  Enable `turbopack` for building and starting the app.
2. `useLinkStatus` to show some nice DX for the user when navigating using `Link` component.

**Changes on Entity Core:**
1. change the ID type to be `EntityCoreIdentifiable` to much the BE
3. `DateMetadata` changed to `Timestamps` and now do no include the IDs class

**Functionality Changes:**
1. create `string` file under `utils` to have all utility functions related to string manipulation.
2. create `array` file under `utils` to have array utility functions.
3. Add `next`and `cache` properties to the `apiClient.RequestOptions` to allow using them in server side calls.


src/api/entitycore/types/shared/context.ts
```diff
@@ -18,64 +21,88 @@ type EntityCoreMapper = {
...
     assetExtension: 'application/json',
     allowedFacets: undefined,
     allowedParams: ['page_size', 'page'],
     queryAll: entitycore.getExperimentalSynapsesPerConnections,
     queryOne: entitycore.getExperimentalSynapsesPerConnection,
   },
-  MESH: {
-    legacyType: undefined,
-    type: 'mesh',
+  EXPERIMENTAL_EXPERIMENTAL_TRACE: {
```
pgetta, line 73: Maybe `EXPERIMENTAL_ELECTROPHYSIOLOGY_RECORDING`, not to repeat experimental twice.

bilalesi: ahh, did not notice that, will change it

pgetta (approved):
Nice work!

PR #271, opened 2025-04-16
Use authentication with GitHub IDP for the AWS role which can publish a new container image in AWS ECR

DriesVerachtert:


DriesVerachtert:
@pgetta with this change, we normally would no longer need the secrets for the access key id and the secret key as the role would be allowed just based on the github repo.

pgetta (approved):
That is very nice, @DriesVerachtert, thank you.

PR #273, opened 2025-04-16
configure tests tools playright and vitest

bilalesi:


playwright.config.ts
```diff
@@ -0,0 +1,52 @@
...
+export default defineConfig({
+  testDir: './e2e',
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
```
pgetta, line 23: I wonder, what's the reason to use only one worker in CI?

pgetta (approved):
Looks good,

I've left one question in the comments.

playwright.config.ts
bilalesi: i will update it when we have the `ci` test stage
currently i used my usename, password, in my machine, i will ask daniel to create `CI-user` and add that stage to the workflow

bilalesi:
thanks @pgetta, i left an answer for the question

playwright.config.ts
pgetta: Thanks!

PR #278, opened 2025-04-23
Feat/read exp traces directly from nwb

pgetta:
This makes experimental trace viewer natively support reading NWB files directly using the [H5WASM](https://github.com/usnistgov/h5wasm) - WebAssembly HDF5 reader/writer module.

 Now both: trace overview and detailed/interactive view plots are generated on the fly:
 
trace overview
trace detailed view

 P.S. Since we don't have corresponding endpoints in entitycore I'll keep current PR targeting the `develop` branch - so that it can be easily tested, and after - will update to merge it into the `entitycore-migration`.


src/components/explore-section/EphysViewer/nwb-trace.tsx
```diff
@@ -0,0 +1,240 @@
+import { File, Group, Dataset, ready } from 'h5wasm';
+
+enum NWBKey {
+  DATA_ORGANIZATION = 'data_organization',
```
bilalesi, line 4: may be it's not big deal, don't you think running this in a worker is better, probably the wasm package support it, 🤔

bilalesi: this one i used before, seems good, https://github.com/alewin/useWorker
or just native way, if you plan to do it, not urgent though

bilalesi (commented):
LGTM, 
just a general note to be easy to migrate to entitycore standards: 
1. file names should be kebab case for all files (no distinctions between pages,  components, hooks, helper functions ..)
2. imports should use `path aliases` instead of relative imports

pgetta:
Added support for non-LNMC-complient NWB files (which do not contain data organisation hierarchy), e.g. human data from UHN. With such files all traces will be shown as sweeps of a "Custom" protocol.

src/components/explore-section/EphysViewer/nwb-trace.tsx
pgetta: Thank you Bilal for the suggestion.

So, as discussed, for a single trace file it's performant enough even in the main thread. For the list view - it might be more efficient to use the thumbnail generation service which have centralised cache for all users.

pgetta:
I'm closing this one in favour of https://github.com/openbraininstitute/core-web-app/pull/290 which will be merged to `entitycore-migration` branch.

PR #295, opened 2025-04-28
add exemplar mophology to emodel details view from entitycore

bilalesi:
- fetch the exemplar morphology in the component that render it
- rename files to follow new naming standards

**TODO**
- fix create/editing emodel when the full data is present as exemplar-traces, ...etc


bilalesi:
thanks @pgetta

PR #297, opened 2025-04-29
Explore circuits - json and api

loris-olivier-obi:
- Created the api route to fetch the circuit and subcircuits count
- Added the ALL_CIRCUITS.json file with all latest circuit and their subcircuits


public/circuits/ALL_CIRCUITS.json
pgetta, line 1: Not a big deal but two things:
* We can save almost 300 KB by minifying this json, you can do it even with online services, e.g. https://codebeautify.org/jsonminifier (or there are VSCode plugins for that).
* There is a typo in the data: _"j**oru**nal": "eLife"_, probably worth fixing (just find and replace should be enough) and notifying whomever generated the data so that next versions will be typo-free.

pgetta (commented):
Loris, looks good!

I've left two small comments regarding the data file.

public/circuits/ALL_CIRCUITS.json
loris-olivier-obi: Thank you very much for the resource. 
I have modify the typo of journal and minified the JSON file

danielaegassan:
@pgetta I generated the json, would something ike this would work?  I also fixed the typo, thanks. 

https://openbraininstitute.sharepoint.com/:u:/s/OBI-Scientificstaff/EfQ2NBbGF7hNkW9Yo4cQK0cB7CGFJVpGmpF0DrBaR-IwNg?e=BZsVw6

pgetta (approved):
Looks good, thank you all.

I'm giving an approval but please remember not to merge it yet during the release preparations.

I hope we will improve the process so that there is no such impact when releasing and we can merge changes right away. That is in the works.

loris-olivier-obi:
> Looks good, thank you all.
> 
> I'm giving an approval but please remember not to merge it yet during the release preparations.
> 
> I hope we will improve the process so that there is no such impact when releasing and we can merge changes right away. That is in the works.

Thank you very much @pgetta. For now, I will release only small PR.
I will wait to get the "go" from Jean-Denis to merge this one.

PR #300, opened 2025-04-30
Add synaptome to explore

bilalesi:
this is part 1; it remains the details view which misses the synaptome configuration panel and simulation section, this will be possible when we migrate bluenaas to entitycore


src/api/entitycore/types/index.ts
```diff
@@ -2,6 +2,7 @@ import type { IExperimentalSynapsesPerConnection } from '@/api/entitycore/types/
 import type { IReconstructionMorphology } from '@/api/entitycore/types/entities/reconstruction-morphology';
 import type { IExperimentalNeuronDensity } from '@/api/entitycore/types/entities/neuron-density';
 import type { IExperimentalBoutonDensity } from '@/api/entitycore/types/entities/bouton-density';
+import type { ISingleNeuronSynaptome } from './entities/single-neuron-synaptome';
```
pgetta, line 5: Probably we want to use an import alias here as well.

src/entity-configuration/definitions/fields-defs/experimental.tsx
```diff
@@ -43,7 +44,8 @@ export const FieldsDefinition: FieldsDefinitionRegistry<EntityCoreObjectTypes> =
         return renderEmptyOrValue(
           renderArray(ensureArray({ input: r.species }).map((s) => s.name))
         );
-      else if ('species' in r.subject) return renderEmptyOrValue(r.subject.species.name);
+      else if ('subject' in r && 'species' in r.subject)
```
pgetta, line 47: Does it make sense to use optional chaining in such places (I find that way more readable) ? Unless we need to distinguish between falsy values or absent properties (which is not the case here from what I understand).

```suggestion
      else if (r?.subject?.species)
```

pgetta (approved):
Looks good

src/features/entities/single-neuron-synaptome/listing-view.tsx
```diff
@@ -0,0 +1,45 @@
+import { useRouter } from 'next/navigation';
```
jdcourcol, line 1: can we have 1 "listing-view" for everything ? I feel that here, only the symbol "ISingleNeuronSynaptome" is changing no ?

bilalesi: yes, that's the plan, most cases uses the same listing-view, this is the first iteration to have the correct views, when i have all the types ready, I will abstract the listing view (soon)

src/entity-configuration/definitions/fields-defs/experimental.tsx
bilalesi: optional chaining will not fix type issues, the generic type will not cover case by case, so for example `IElectricalCellRecording` will not have subject, so even if you do r.subject you still have type error, but with forcing js checks typescript will infer the rest

src/api/entitycore/types/index.ts
bilalesi: this will be fixed in the next PR

PR #301, opened 2025-04-30
Ephys improvements

pgetta:
* Correct entitycore type for electrical cell recording.
* Make thumbnail service to use kebab-cased entity type as a route segment - to match actual api implementation.
* Move trace viewer to features folder.
* Fix entity config for electrical cell recording to allow all params and enable faceted search.


bilalesi (approved):
LGTM

bilalesi (approved):

PR #302, opened 2025-04-30
Add counts to explore interactive

bilalesi:
1. refactor explore interactive page 
2. create new function that request all entity types counts (currently we are using the `/get-many` endpoint with `page=1` and `page_size=1`, to get the total of items, waiting for the `/count` endpoint)
3. use the same function to get the counts in the explore listing page


src/entity-configuration/domain/index.ts
```diff
@@ -4,18 +4,27 @@ import {
...
   ElectricalCellRecording,
-  BoutonDensity,
   NeuronDensity,
-  SingleNeuronSynaptome,
-  MEmodel,
+  BoutonDensity,
+  SynapsePerConnection,
+};
+
+// NOTE: order is important (it's used in stats panel in explore)
```
pgetta, line 20: Maybe we can use an array somewhere somehow to define the order, otherwise this is fragile.

pgetta (approved):

src/entity-configuration/domain/index.ts
bilalesi: good point, will have it fixed in the next PR, today

PR #304, opened 2025-05-01
Creation of single circuit page base

loris-olivier-obi:
- Creation of the dynamic route for single circuit [key]/page.tsx
- Added a utils buildCircuitMap


public/circuits/ALL_CIRCUITS.json
pgetta, line 1: This was already added in the previous PR, is that right? Also this doesn't seem to be compressed.

loris-olivier-obi: It's now again the minified version. Do you see now the minified one @pgetta ?

loris-olivier-obi: @pgetta running prettier:writte unminified it. For future PR, I will run prettier, then replace ALL_CIRCUITS for its minified version and push a last time. Would that work like this?

pgetta: Agreed to add `/public` folder to be ignored by prettier.

pgetta (approved):
Nice work!

bilalesi (approved):

PR #305, opened 2025-05-01
use only one slug -type-  to decide which entity to render for both l…

bilalesi:
1. refactor listing/details pages 
2. create listing-view `/features/views/listing` as base component for listing entities (exception for e-models that has special logic which might be excluded later (when building the feature) to have one component for all listing) (cc: @jdcourcol )
3. create details-views `/features/views/details` for both experimental and model data
4. add `slugs` dictionary (composed of separated dict for experimental, model, .. ), this should be the source of truth for all base urls of all types (enforced with types)
5. all listing/details pages are now server pages, this is a preparation for moving away from unnecessary `atoms`, 
   - ex: the `summary` component is fetching the entity details and pass it to other components, which is a bit problimatic if we want to extend the data for children, one way it to fetch every thing in the server (near the db) and pass it down the road
   - for listing pages, it can be an option to load the first page in the server and then do pagination in the client
   - ...
6. add validation status to `memodel` listing page.
7. fix contribution for `emodel`.


src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/explore/interactive/(data)/experimental/[type]/page.tsx
Comment (pgetta) [inline src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/explore/interactive/(data)/experimental/[type]/page.tsx:1]: This is very similar to `src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/explore/interactive/(data)/model/[type]/page.tsx`, only one line of difference. Later on would be nice to de-duplicate.

src/entity-configuration/definitions/fields-defs/common.ts
```diff
@@ -143,6 +144,29 @@ export const FieldsDefinition: FieldsDefinitionRegistry<EntityCoreObjectTypes> =
     isFilterable: true,
     isDisplayable: true,
   },
+  [EntityCoreFields.Contributions]: {
+    title: 'Contributors',
```
pgetta, line 148: Just above there is the `EntityCoreFields.Contribution` field definition with the same title `Contributors`, is it different from this one?

Also, if possible, would be nice to have field key and title consistent (contribution vs contributor).

src/entity-configuration/definitions/fields-defs/enums.ts
```diff
@@ -50,7 +50,7 @@ export enum EntityCoreFields {
   EModelResponse = 'eModelResponse',
   MEModelMorphologyPreview = 'meModelMorphologyPreview',
   MEModelTracePreview = 'meModelTracePreview',
-  MEModelValidated = 'meModelValidated',
+  MEModelValidated = 'validation_status',
```
pgetta, line 53: Maybe 
```suggestion
  ValidationStatus = 'validation_status',
```

Just to be consistent (I don't know for sure if that makes sense because this enum is an aggregation of all fields)

pgetta: Or MEModelValidationStatus

pgetta (commented):
Nice work!

Left couple of questions.

src/entity-configuration/definitions/fields-defs/common.ts
bilalesi: yeah, i put a `TODO` on the top of `EntityCoreFields.Contribution` to delete it later, in the beginning there was some different naming sometimes w/o "`s`". 
so yes will need to delete one of them, 
for the naming:
legacy naming was "contributors" and with entitycore is "contributions", 
and the `EntityCoreFields` is the mapping of selectors from response objects, so it's a bit messy

src/entity-configuration/definitions/fields-defs/enums.ts
bilalesi: fixed

src/entity-configuration/definitions/fields-defs/common.ts
bilalesi: i kept only `contributions` but the field still named as entitycore

PR #308, opened 2025-05-02
Componisation of circuit table

loris-olivier-obi:
Created a single independent component for the circuit table. This component can be used in the list view of circuits and in detail view for parent circuit, subcircuits and derived from circuits.


package.json
```diff
@@ -146,6 +148,7 @@
     "postcss": "^8.4.23",
     "prettier": "^3.2.4",
     "prettier-plugin-tailwindcss": "^0.5.11",
+    "react-resizable": "^3.0.5",
```
pgetta, line 151: There is a duplication, it should be only listed once in `dependencies`.

src/app/app/virtual-lab/(free)/explore/(interactive)/interactive/(data)/model/circuit/[key]/page.tsx
```diff
@@ -13,7 +13,10 @@ export default function CircuitDetailPage() {
   const [error, setError] = useState<string | null>(null);
 
   const params = useParams();
-  const circuitKey = params?.key as string | undefined;
+  const circuitKey = params.key as string | undefined;
+
+  console.log('What is my params? ', params)
```
Comment (pgetta) [inline src/app/app/virtual-lab/(free)/explore/(interactive)/interactive/(data)/model/circuit/[key]/page.tsx]: In general it's best to not have any console calls in the client/browser code at all (unless it's a specific case where you need to debug something in staging/prod and don't have any other means).

src/components/explore-section/Circuit/global/CircuitTable.tsx
```diff
@@ -0,0 +1,278 @@
...
+  const filteredData = useFilteredData(data, { searchQuery, numericFilter });
+
+  const handleRowExpandClick = useCallback((row: CircuitSchemaProps, _index: number) => {
+    const rowKey = row.key;
+    setExpandedRowKeys((prev) =>
+      prev.includes(rowKey) ? prev.filter((key) => key !== rowKey) : [...prev, rowKey]
+    );
+  }, []);
+
+  const handleResize = useCallback(
```
pgetta, line 90: Does resizing work now? Last time I've tried in that huge PR and it was not possible to resize columns more then +- 10 px.

src/components/explore-section/Circuit/global/CircuitTable.tsx
```diff
@@ -0,0 +1,278 @@
...
+    const rowKey = row.key;
+    setExpandedRowKeys((prev) =>
+      expanded ? [...prev, rowKey] : prev.filter((key) => key !== rowKey)
+    );
+  }, []);
+
+  const renderSubcircuits = useCallback(
+    (circuit: CircuitSchemaProps) =>
+      circuit.subcircuits && circuit.subcircuits.length > 0 ? (
+        <SubcircuitTable
```
pgetta, line 139: Is Subcircuit table any different from the circuit table?

If they contain the same information in theory we can re-use single "CircuitTable" component, but maybe apply different styling to have some indentation (to highlight the hierarchy).

src/components/explore-section/Circuit/global/Columns.tsx
```diff
@@ -0,0 +1,149 @@
...
+      width: column.width ?? 150,
+      onResize: handleResize(column.key as string),
+    }),
+  },
+  {
+    title: 'Description',
+    key: 'description',
+    width: 200,
+    render: (_value: any, record: CircuitSchemaProps, _index: number) => (
+      <div className="text-ellipsis whitespace-nowrap font-normal">{record.description}</div>
```
pgetta, line 80: Oh nice, you've made it work with CSS ellipsis!

src/components/explore-section/Circuit/global/Columns.tsx
```diff
@@ -0,0 +1,149 @@
...
+      <div className="whitespace-nowrap font-normal">{record.numberOfNeurons}</div>
+    ),
+    onHeaderCell: (column) => ({
+      width: column.width ?? 150,
+      onResize: handleResize(column.key as string),
+    }),
+  },
+  {
+    title: 'Species',
+    key: 'specie',
```
pgetta, line 113: It's `species`, singular or plural - it's written the same way.

src/components/explore-section/Circuit/global/DownloadCircuitButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+  fileUrl,
+  selectedRowKeys,
+}: {
+  fileUrl: string;
+  selectedRowKeys: string[] | number[];
+}) {
+  return (
+    <a
+      href={fileUrl}
+      type="button"
```
pgetta, line 14: Type attribute of a hyperlink element is [used to define a MIME type](https://www.w3schools.com/tags/att_a_type.asp) of a linked document, so having it set to "button" is not a correct HTML - I think it's safe to just remove that.

Also it's best to add a `download` (without a value) attribute to make sure browser knows how to handle the content (to download it), see https://www.w3schools.com/tags/att_a_download.asp.

src/components/explore-section/Circuit/global/SubcircuitsTable.tsx
```diff
@@ -0,0 +1,120 @@
...
+            '[&_.ant-table-thead_th]:!text-sm',
+            '[&_.ant-table-thead_th]:!font-normal',
+            '[&_.ant-table-thead_th]:!text-[#8C8C8C]',
+            '[&_.ant-table-thead_th]:uppercase',
+            '[&_.ant-table-thead_th]:tracking-[0.05em]',
+            '[&_.ant-table-tbody > tr:last-child > td]:border-b-0',
+            '[&_.ant-table-thead > tr > th]:border-b-0',
+            '[&_.ant-table-expand-icon-col]:w-0',
+            '[&_.ant-table-expand-icon-col]:hidden',
+            styles.circuitTable
```
pgetta, line 63: From the maintainability perspective it's best to either use CSS modules or Tailwind classes for a single component/ React module. Mixing them hurts code readability.

src/components/explore-section/Circuit/global/SubcircuitsTable.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  return (
+    <div className="relative flex flex-col">
+      <div className="relative flex flex-row pl-2">
+        <ArrowSmall iconColor="#8C8C8C" className="relative -top-0.5" />
+        <span className="ml-3 pb-2 text-base font-semibold uppercase tracking-wider text-[#8C8C8C]">
+          Subcircuits
+        </span>
+      </div>
+      {downloadable ? (
```
pgetta, line 49: It seems there is some code duplication that can be avoided by replacing ternary condition creating Table by just one Table and customising props/attributes in-place.

pgetta (commented):
Nice work, Loris,

I've left some small comments.

src/app/app/virtual-lab/(free)/explore/(interactive)/interactive/(data)/model/circuit/[key]/page.tsx
Comment (loris-olivier-obi) [reply src/app/app/virtual-lab/(free)/explore/(interactive)/interactive/(data)/model/circuit/[key]/page.tsx]: Sorry, I removed it and pushed a new commit without it.

src/components/explore-section/Circuit/global/CircuitTable.tsx
loris-olivier-obi: @pgetta No, it's not working. I will tackle that in another PR. 
Do you want me to remove the handleResize or can I keep it for one of the next PR?

src/components/explore-section/Circuit/global/CircuitTable.tsx
loris-olivier-obi: @pgetta Ah yes, right. I think the styling of subcircuit are handle automatically by AntD for the indentation. I will replace SubcircuitTable with the same CircuitTable

src/components/explore-section/Circuit/global/DownloadCircuitButton.tsx
loris-olivier-obi: My mistake. When I turned the button into an hyperlink, I forgot to remove this one.

src/components/explore-section/Circuit/global/SubcircuitsTable.tsx
loris-olivier-obi: Ahh, great. I will use then the module.

src/components/explore-section/Circuit/global/CircuitTable.tsx
pgetta: It's already a lot of work, so let's keep it not have any extra. But would be nice to either fix it by the next deployment time or remove UI parts which suggest that user can resize, not to confuse the user with something broken.

package.json
loris-olivier-obi: Removed

src/components/explore-section/Circuit/global/CircuitTable.tsx
loris-olivier-obi: Done

src/components/explore-section/Circuit/global/SubcircuitsTable.tsx
loris-olivier-obi: I tried different ways of displaying or not the radio button (or checkbox) but it doesn't appear to work.
It concerns this part:
data-row-selection={downloadable.toString()} and rowSelection={rowSelection}. Do you have any idea in mind how I could do it without duplicating the Table @pgetta ?

src/app/app/virtual-lab/(free)/explore/(interactive)/interactive/(data)/model/circuit/page.tsx
```diff
@@ -10,5 +15,47 @@ export type ColumnType = {
...
+      <div className="relative flex h-[50vh] w-full items-center justify-center text-lg font-normal text-primary-9">
+        Loading...
+      </div>
+    );
+  }
+
+  if (error) {
+    return (
+      <div className="relative flex h-[50vh] w-full items-center justify-center text-lg font-normal text-red-600">
+        Loading...
```
pgetta, line 55: Probably a copy/paste, we need to have an error message here.

src/app/app/virtual-lab/(free)/explore/(interactive)/interactive/(data)/model/circuit/page.tsx
pgetta, line 1: Does this component identical to `src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/explore/interactive/(data)/model/circuit/page.tsx` ? If so, we can create a separate component somewhere, import and re-use it in these pages.

That will make it more maintainable and less error-prone (in case someone updates one component and doesn't know/forgets about the second).

src/components/explore-section/ExploreListingLayout/index.tsx
```diff
@@ -126,6 +127,32 @@ export default function ExploreListingLayout({
     };
   });
 
+  useEffect(() => {
+    async function fetchCircuitCount() {
+      try {
+        const response = await fetch('/api/circuits/count');
```
pgetta, line 133: It seems we have 4 separate components in this PR that load the circuit data, it would be best to move this logic into a shared hook/function somewhere so that, again, if we need to modify something - it's all in one place.
Please see how it's best to address that: here or maybe in a separate PR.

src/app/app/virtual-lab/(free)/explore/(interactive)/interactive/(data)/model/circuit/page.tsx
loris-olivier-obi: Ah yes, true. I will create that now.

src/components/explore-section/ExploreListingLayout/index.tsx
loris-olivier-obi: Thank you. I will address that in a separate PR like that I can tackle in a cleaner way.

src/app/app/virtual-lab/(free)/explore/(interactive)/interactive/(data)/model/circuit/page.tsx
loris-olivier-obi: Ah yes, true. I have added the error in the message

src/components/explore-section/Circuit/global/CircuitsListingPageComponent.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <div className="relative flex h-[50vh] w-full items-center justify-center text-lg font-normal text-primary-9">
+        Loading...
+      </div>
+    );
+  }
+
+  if (error) {
+    return (
+      <div className="relative flex h-[50vh] w-full items-center justify-center text-lg font-normal text-red-600">
+        An error occurred: {error}
```
pgetta, line 55: A small thing, but it's better not to expose the error directly to the user, because in most of the cases, either it's a network error or an invalid status code due to some issues with the API, the error will contain technical details which will not make much sense to the end user.

A common scenario is to match the error to some expected ones and provide more descriptive message, but in this case it might be an overkill.
So I would just leave there some generic message without details.

loris-olivier-obi: Ahhh. Okay. Next PR I will fix this message. 
In that case I can say that there is an error to retrieve the data associated to the circuit?

PR #318, opened 2025-05-06
Fix detailed view scroll and top level layout

pgetta:
This is fixes the circuit layout so that:
* The view can be scrolled.
* The sticky menu appears on the top of the view.


loris-olivier-obi (approved):
Thanks for all the fixes

PR #324, opened 2025-05-07
Fix all lint issues

tolokoban:
There was few linting errors in the AntD Resizable Columns.


loris-olivier-obi (approved):
Mieux que le chocolat et PR en plus

PR #325, opened 2025-05-07
Removed the component of search bar and filter options

loris-olivier-obi:


src/components/explore-section/Circuit/utils/filtering-circuits.ts
```diff
@@ -0,0 +1,35 @@
...
+        circuit.brainRegion.toLowerCase().includes(searchQuery.toLowerCase());
+
+      const matchesFilter = matchesNumericFilter(circuit, filter);
+
+      const filteredSubcircuits =
+        circuit.subcircuits && circuit.subcircuits.length > 0
+          ? filteringCircuits(circuit.subcircuits, filter, searchQuery)
+          : [];
+
+      const includeCircuit = (matchesSearch && matchesFilter) || filteredSubcircuits.length > 0;
```
pgetta, line 23: This might be a bit confusing behaviour, let's imagine the following situation:

* root circuit A (1M neurons.
  * subcircuit AX (100 neurons).
  * subcircuit AY (1K neurons).
  
User applies a filter neuron count < 500 and gets:
* root circuit A
  * subsircuit AX

The issue is that the user might think that the root circuit A somehow conforms to the filter, but that's not correct.

One way to solve this is to somehow let user know that we show (some) parent circuits in order to preserve hierarchy, but not all of them are matching the filters. (maybe by disabling/graying them out). - with such a fix we should have an additional property to control that UI behaviour.

(Please let me know if this is not how it works and I misunderstood).

pgetta (commented):
Good work, Loris,

I've left couple of small comments.

src/components/explore-section/Circuit/utils/filtering-circuits.ts
loris-olivier-obi: Ah yes, absolutely. We discuss with Daniela yesterday and we have found a solution to 1) get only circuit that pass the condition; 2) keep the hierarchical structure. We can discuss about that whenever you want this week @pgetta

pgetta (approved):

package-lock.json
pgetta, line 1: It seems `react-resizable` has been removed from package-lock, is that intentional?

loris-olivier-obi: Should I reinstall it @pgetta ?

PR #326, opened 2025-05-07
189/prevent emails not approved by insurance

tolokoban:
There is a list of countries forbidden by insurance contract.
This PR is about adding validation rules for email inputs on Virtual lab creation and update.
We just check the end part of the email (the two char country code).


loris-olivier-obi (dismissed):
Nice PR

src/util/email.ts
```diff
@@ -13,3 +13,32 @@ export function isValidEMail(email: string): boolean {
...
+  by: 'Belarus',
+  cu: 'Cuba',
+  ir: 'Iran',
+  mm: 'Myanmar',
+  kp: 'North Korea',
+  ru: 'Russia',
+  sy: 'Syria',
+  ua: 'Ukraine',
+  ve: 'Venezuela',
+};
```
bilalesi, line 28: can we remove this country from the countries json file too

tolokoban: Oh, yes. That's a good idea. This way they cannot choose them in the combo.

src/components/VirtualLab/create-entity-flows/profile/form.tsx
```diff
@@ -220,3 +152,19 @@ function Profile({ data }: Props) {
 }
 
 export default Profile;
+
+function useInitialValues(data: UserProfileResponse | undefined): ProfileFormData {
+  return React.useMemo(() => {
+    const initialValues: ProfileFormData = {
+      email: data?.email ?? '',
```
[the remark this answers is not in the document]
tolokoban: You're right.
At first, I was planning to add some validation logic in here, but I dropped the idea because I'm doing it elsewhere.

bilalesi (approved):

PR #330, opened 2025-05-07
Add entitycore download (UI and backend API)

pgetta:


src/api/apiClient.ts
```diff
@@ -209,7 +209,7 @@ class ApiClient {
     method: string,
     endpoint: string,
     options: RequestOptions = {},
-    config: RequestConfiguration & { cache?: CacheConfiguration } = {},
+    config: RequestConfiguration & { cache?: CacheConfiguration; asRawResponse?: boolean } = {},
```
pgetta, line 212: Not sure I'm happy with all these extensions to the API Client, maybe there are better solutions.

src/api/entitycore/queries/assets/index.ts
```diff
@@ -67,21 +67,29 @@ export async function getAsset({
...
   id,
+  asRawResponse = false,
+  retryOnError = false,
 }: {
   ctx?: WorkspaceContext;
   entityType: EntityCoreDataType;
   entityId: string;
   id: string;
+  asRawResponse?: boolean;
+  retryOnError?: boolean;
```
pgetta, line 83: Same here

src/api/entitycore/types/index.ts
```diff
@@ -7,7 +7,7 @@ import type { IExperimentalBoutonDensity } from '@/api/entitycore/types/entities
 import type { IMEModel } from '@/api/entitycore/types/entities/me-model';
 import type { IEModel } from '@/api/entitycore/types/entities/e-model';
 
-export type * from '@/api/entitycore/types/entity-type';
+export * from '@/api/entitycore/types/entity-type';
```
pgetta, line 10: This is so that we can use that entity type enum to provide values as well.

src/app/api/entity-download/[entityType]/route.ts
```diff
@@ -0,0 +1,463 @@
+import { pipeline, Readable } from 'stream';
```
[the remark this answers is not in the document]
Comment (bilalesi) [reply src/app/api/entity-download/[entityType]/route.ts:1]: there is already (`src/app/api/entity-core`), probably move it here is better to keep all related endpoints for `entitycore` under one path
fyi: the name should be `entity-core` not `entitycore` so the load balancer not conflict it with the real service

bilalesi (approved):
LGTM 👍

pgetta:
Agreed to merge this and address the feedback in a separate PR.
Thank you @bilalesi.

PR #333, opened 2025-05-08
Fix download for 3rd depth level subcircuits

loris-olivier-obi:
- Now level 3 and lower depth can be downloaded


loris-olivier-obi:
> @loris-olivier-obi I'm not sure I understand, how removing a file fixes download for 3rd depth level subcircuits?

My apologies @pgetta, some content wasn't pushed. Now it should be all good.

pgetta (approved):
Looks good!

PR #337, opened 2025-05-09
Articles cards do not appear anymore in the chat AI.

tolokoban:
Literature search tool has a new output format now, but the current code was just silently ignoring it.

Since AI agent is going under heavy development, we may face many API breaks in the future.
Before this PR, if the format of the result of a tool was not as expected, the tool result was just silently ignored.
Now, we still ignore it, but we report a detailed error.

### DIFF
--- a/src/components/literature-suggestions/message-item/tools/articles/tool-articles/types.ts
+++ b/src/components/literature-suggestions/message-item/tools/articles/tool-articles/types.ts
@@ -1,4 +1,5 @@
-import { isType } from '@/util/type-guards';
+import { logError, logInfo } from '@/util/logger';
+import { assertType } from '@/util/type-guards';
 
 export interface ScientificArticle {
   title: string;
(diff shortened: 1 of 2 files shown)


src/components/literature-suggestions/message-item/tools/articles/tool-articles/types.ts
```diff
@@ -14,16 +15,34 @@ export interface LiteratureSearchToolItem {
...
-  ]);
+export interface LiteratureSearchToolResult {
+  articles: LiteratureSearchToolItem[];
+  error: unknown;
+}
+
+export function isLiteratureSearchToolResult(data: unknown): data is LiteratureSearchToolResult {
+  try {
+    assertType(data, {
+      articles: [
```
bilalesi, line 26: suggestion: we have `zod` that is straightforward, less boilerplate and also nice, extensible api

bilalesi (approved):
left a suggestion

src/components/literature-suggestions/message-item/tools/articles/tool-articles/types.ts
tolokoban: I will definitly have a look at __zod__, but this code is not so boilerplate and it gives precise error on what part of the type failed.

PR #343, opened 2025-05-12
Add ticket based entity archive downloads, split into modules

pgetta:
Changes
* Add ticket based downloads (it turned out that AWS doesn't like long URLs and those requests were failing I guess due to WAF, but this implementation prevent such issues to happen).
* Split implementation into different modules.
* Extend metadata to create files in CSV and JSON (as requested by scientists).
* Extend metadata to return number of entries by request.


pgetta:
Thank you, Bilal

PR #347, opened 2025-05-13
Download options for circuits

loris-olivier-obi:
- Removed the checkbox for each row
- Remove the button appearing at the bottom right after selecting a circuit
- Added a download icon for each row
- Added a sliding panel with all the download options
- Updated the circuit JSON file with the correct structure for files


tolokoban (approved):
LGTM

PR #349, opened 2025-05-13
Add filter to circuit tables

loris-olivier-obi:
- Reintegrated filters for number of neurons for now
- Added style for when row matches the filter
- Cleaned some of the previous code for resizable columns


src/components/explore-section/Circuit/global/CircuitTable.tsx
```diff
@@ -37,11 +29,18 @@ export default function CircuitTable({
...
 
-  // const filteredData = useFilteredData(data, { searchQuery, numericFilter });
+  useEffect(() => {
+    if (numericFilter) {
+      const expandableKeys = collectExpandableKeys(data);
+      setExpandedRowKeys(expandableKeys);
+    } else {
+      setExpandedRowKeys([]);
+    }
+  }, [numericFilter, minValue, maxValue, data]);
```
tolokoban, line 43: `minValue` and `maxValue` are not used in this effect code.
Have you forgot something, or can we remove them from the dependencies?

bilalesi: no need for the `useEffect` here, it can be handled in the even listner meaning in `onFilterChange`

src/components/explore-section/Circuit/global/SubcircuitsTable.tsx
```diff
@@ -51,11 +56,6 @@ export default function SubcircuitTable({
             '--ant-table-expand-icon-col-width': '0px',
           } as React.CSSProperties
         }
```
bilalesi, line 58: i could not find a use for this one `--ant-table-expand-icon-col-width` it can be deleted

bilalesi (commented):
in another follow up PR can we have the naming as discussed [here](https://teams.microsoft.com/l/message/19:kRU5l_iuGbBd50PKiXKDDIr7trhAb0nM8_YuJEJAMnA1@thread.tacv2/1738921198549?tenantId=c355c1d8-2b0b-4398-a8ea-b2134d2db027&groupId=ee40e654-3bca-4126-a5f9-020fd958c8b3&parentMessageId=1738921198549&teamName=Web%20application%20devs&channelName=OBI%20web%20application&createdTime=1738921198549)

so the files/folders should all be in `kebab-case`, this will also facilitate the migration rebase

src/components/explore-section/Circuit/global/SubcircuitsTable.tsx
loris-olivier-obi: @bilalesi I will change with this PR the file naming. 
I also removed this part of the code that was part of column resizing work I did in the past.

src/components/explore-section/Circuit/global/CircuitTable.tsx
loris-olivier-obi: @bilalesi If I assign to onFilterChange, I don't get my logic working again.
May I ask you how you would remove the useEffect and use it in the event listener please?

tolokoban (approved):
Good point on changing the file casing.

PR #351, opened 2025-05-13
Circuit detail active link licence parent circuit

loris-olivier-obi:
- Added link to parent circuit parameter box
- Added link to the license parameter box


tolokoban (approved):
Approved!

PR #354, opened 2025-05-13
Circuit table hierarchical and flat list views

loris-olivier-obi:
Adding the logic of showing table either hierarchically or flatten.

### DIFF
--- a/src/components/icons/FlatListViewIcon.tsx
+++ b/src/components/icons/FlatListViewIcon.tsx
@@ -1,0 +1,22 @@
+export default function FlatListViewIcon({
+  iconColor,
+  className,
+}: {
+  iconColor: string;
+  className?: string;
+}) {
+  return (
+    <svg
+      width="19"
+      height="13"
+      viewBox="0 0 19 13"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+    >
+      <path d="M19 0V1H0V0H19Z" fill={iconColor} />
+      <path d="M19 6V7H0V6H19Z" fill={iconColor} />
+      <path d="M19 12V13H0V12H19Z" fill={iconColor} />
+    </svg>
+  );
+}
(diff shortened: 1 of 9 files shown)


src/components/icons/FlatListViewIcon.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <svg
+      width="19"
+      height="13"
+      viewBox="0 0 19 13"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+    >
+      <path d="M19 0V1H0V0H19Z" fill={iconColor} />
```
tolokoban, line 17: You could have this attribute in common for the 3 `path`, like this:

```ts
<g fill={iconColor ?? "currentColor"}>
  <path d="M19 0V1H0V0H19Z"/>
  <path d="M19 6V7H0V6H19Z" />
  <path d="M19 12V13H0V12H19Z" />
</g>
```

But it's just a suggestion.

src/components/icons/HierarchicalViewIcon.tsx
```diff
@@ -0,0 +1,30 @@
+export default function HierarchicalViewIcon({
+  iconColor,
+  className,
+}: {
+  iconColor: string;
+  className?: string;
+}) {
+  return (
+    <svg
+      width="19"
```
tolokoban, line 10: Another suggestion, but you can ignore it.

Here, you have a fixed pixel size for the icons.
If you use something like this, the you can control the size in CSS:

```ts
width="1.36em"
height="1em"
```

tolokoban (approved):
Few suggestions, but I approve.

src/components/icons/FlatListViewIcon.tsx
loris-olivier-obi: @tolokoban Thank you very much. I take note of that for future use.

src/components/icons/HierarchicalViewIcon.tsx
loris-olivier-obi: Take note of that too for future usage. Thank you very much

PR #355, opened 2025-05-14
Circuit table add published in column

loris-olivier-obi:
- Added type to the Circuit schema
- Added the column to the table


tolokoban (approved):
LGTM

PR #359, opened 2025-05-14
62/ai more suggestions

tolokoban:
From now on, when we ask AI agent for suggestion, we send the thread_id.


loris-olivier-obi (approved):
LGTM

PR #364, opened 2025-05-14
refactor models to be server component

bilalesi:
1. Add `mechanisms` to the `e-model` detail page as `ion-channels`, but it misses the `location`, so now there is mock locations randomly assigned (will be fixed when ready in the backend).
2. Migrate models detail page to server component, so all related data will be fetched in the server and pass it down to the children, for that reason, the domain definition of an entity has a new property `expand`, which will have the expanded fields needed in each model with correspondent function that resolve the entity
```ts
api: {
    config: {
      allowedFacets: true,
      allowedParams: 'all',
    },
    query: {
      list: getSingleNeuronSynaptomes,
      one: getSingleNeuronSynaptome,
      create: createSingleNeuronSynaptome,
    },
    expand: {
      memodel: (source, context) => getMEModel({ id: source.me_model.id, context }),
      config: (source, context) => getSingleNeuronSynaptomeConfiguration({ source, context }),
    },
  },
```
and other function `applyEntityExpansions` that fetch all this expansions 
```ts
const source = await entity.api.query.one({
    id: id,
    context: { virtualLabId, projectId },
  });

  let data = {} as ExpandType;
  if (entity.api.expand) {
    data = await applyEntityExpansions<IEModel, ExpandType>(entity, source);
  }
```


src/features/entities/single-neuron-synaptome/build/elements/synapse-config-form.tsx
```diff
@@ -166,17 +166,15 @@ export default function SynaptomeConfigurationForm({
           SingleNeuronSynaptome.api.query.create!({
             context: { virtualLabId, projectId },
             body: {
-              brain_region_id: 31,
+              brain_region_id: '23a7711a-8133-4876-b7eb-dcd9e87a1613',
```
bilalesi, line 169: this is will be fixed when we have brain region PR merged

tolokoban (approved):
Well done Bilal!

PR #382, opened 2025-05-21
Circuit detail view fix z index position of sections tabs

loris-olivier-obi:
Fixed the problem of z-index between the circuit detail view section tab header which was sticky. Apparently, the sticky position mess with the z-index even if you have explicitly define a value. 
Now it's fixed


tolokoban (approved):
Nothing suspiscious to me.

PR #384, opened 2025-05-22
Simulation thumbnails

g-bar:


src/api/thumbnail-svc/index.ts
```diff
@@ -1,38 +1,57 @@
 import find from 'lodash/find';
 import kebabCase from 'lodash/kebabCase';
-
+import { EntityTypeEnum } from '../entitycore/types';
```
bilalesi, line 3: can we use `path alias import` as agreed

src/entity-configuration/definitions/view-defs/experiment.ts
```diff
@@ -38,6 +39,7 @@ export const ViewsDefinition: { [key: string]: ViewDefinitionConfig } = {
       EntityCoreFields.Description,
       EntityCoreFields.SimulationStimulus,
       EntityCoreFields.SimulationResponse,
+      EntityCoreFields.SimulationStatus,
```
bilalesi, line 42: we don't have status for `synaptome simulation`

g-bar: Ah, is there a way to add it only for single cell? Or do we remove it from both?

bilalesi: you already have it in single cell (line 19), so it's fine

bilalesi (approved):

g-bar:
Thanks @bilalesi

PR #387, opened 2025-05-23
Fix latest entitycore

g-bar:
Latest entitycore commits introduced a braking change causing the explore list view to fail to render.
This fixes it.


bilalesi:
👍 
better to merge when we have the latest entitycore image in staging

PR #388, opened 2025-05-23
Replace svg icons within notebooks table with svg components, move the icons within the buttons so users can click on them

DriesVerachtert:
@bilalesi , @pgetta,
This is a follow-up of that previous PR for the 'run' button in notebooks: the other items in the popup menu now also use svg components instead of Image tags pointing to an svg file. The icons are also now within the button, so users can click on them too. Please let me know if there's something I should change.

image


bilalesi (approved):
LGTM

PR #392, opened 2025-05-26
fix search results in listing page

bilalesi:
when you search in listing page, the history of paginated entities kept so on search the atom that keep the previous results should be reseted


g-bar (approved):
LGTM

PR #393, opened 2025-05-26
fix-view-button

g-bar:


src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/simulate/page.tsx
```diff
@@ -98,6 +98,7 @@ function BrowseSimsTab() {
               tableClass={classNames('overflow-y-auto grow', Styles.table)}
               dataKey={dataKey}
               showLoadingState={false}
+              useBrainRegion={false}
```
Comment (bilalesi) [inline src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/simulate/page.tsx:101]: I think it should be `false`, we doin't have the brain region panel in simulate page ?

Comment (g-bar) [reply src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/simulate/page.tsx:101]: 1. We don't
2. If we don't pass this prop here it introduces a bug becasue internally the ExploreListingView changes the dataKey dynamically based on this prop and it fails to select the rows, because those rely on selectedRowsAtom which get passed the dataKey.

This is pretty odd design and shoud be refactored maybe, but not sure how it works internally do not sure.

Comment (bilalesi) [reply src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/simulate/page.tsx:101]: ignore my comment, I want to say it's correct, but my mind slipped bahh, 

for the design, previously the `dataAtom` and all other related atoms (as selectedRows) were relying on two atoms, the selected brain region, and the scope, 
and the new idea now is to have the brain region as query param as `source of truth`, so this require to pass the brain region id as param to `dataAtom`.
the `useBrainRegion` was introduced because some listing pages as bookmark, browse simulations in simulate page do not require the brain region

please check this [PR](https://github.com/openbraininstitute/core-web-app/pull/348)
happy to get feedback and improve it specially in persisting the brain region in local storage (which was the case before but was not that great)

Comment (g-bar) [reply src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/simulate/page.tsx:101]: Yes, I think it makes sense, what I'm not so sure about is having the dataKey change implicitly, because that was the source of the bug and was pretty hard to debug, not intuitive that another prop changes the key.

https://github.com/openbraininstitute/core-web-app/blob/5860a64a9061a6f7af36701f9e8dbc073c4b3775/src/components/explore-section/ExploreSectionListingView/index.tsx#L66

Not sure if that logic can be put in the `resolveDataKey` function?

PR #398, opened 2025-05-28
Send requests to the notebook service to start a new notebook

DriesVerachtert:


src/services/notebooks/index.ts
```diff
@@ -0,0 +1,60 @@
...
+    notebook,
+    vlabId,
+    projectId,
+    session,
+    implementationType: NotebookImplementationType.DUMMY,
+  };
+  // console.log("Request body:", request);
+
+  const res = await authFetch(`http://127.0.0.1:8000/api/notebook_service/notebook/start`, {
+    method: 'POST',
```
bilalesi, line 54: do you want this to be merged now, or until will have a dedicated deployment for the notebook service ?

src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/(pages)/notebooks/NotebookTable.tsx
```diff
@@ -105,7 +107,23 @@ function NotebookTable({
...
 
   const runNotebook = async (notebook: Notebook) => {
-    notification.info(`Request received to run notebook: ${notebook.name}`);
+    // notification.info(`Request received to run notebook: ${notebook.name}`);
+    try {
+      const session = await getSession();
+      if (session === null) {
+        notification.error('You must be logged in to run a notebook');
+        return;
+      }
```
Comment (bilalesi) [inline src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/(pages)/notebooks/NotebookTable.tsx:116]: this is okey, but I see sessions more like service level, not at the UI level, 
i.e: the `getSessions` should be in a service function, that throw an error, and then catch it here and show a notification ()
```ts
      // this is in the service
      const session = await getSession();
      if (!session) {
        throw Error("no session found", {
             cause: {
                     error_code: "SESSION_NOT_FOUND",
                     hint: "Make sure the user is authenticated before calling this 
                                function."
             }
        }) 
     }
```

```ts
// in the component
try{}
catch(error){
  if(error instanceof Error && "cause" in error){
      if((error.cause as ({error_code: string, hint: string})).error_code === "SESSION_NOT_FOUND")
      notification.xxx
  }
  
}
```

this can be also achievable easily, if we have a custom error  (iirc, we have it in the migration)
```ts
// this can be under src/utils/errors
class SessionError extends Error {
  errorCode: string;
  cause?: unknown;

  constructor(message: string, errorCode: string, cause?: unknown) {
    super(message);
    this.name = "SESSION_NOT_FOUND";
    this.errorCode = errorCode;
    this.cause = cause;

    Object.setPrototypeOf(this, SessionError.prototype);
  }
}
```
later in the catch block 
```ts
if(error instanceof SessionError) {notification.xx)
```

Comment (DriesVerachtert) [reply src/app/app/virtual-lab/lab/[virtualLabId]/project/[projectId]/(pages)/notebooks/NotebookTable.tsx]: Bilal, yes it makes sense to keep the UI stuff like notification.something out of the service which runs on the server. I've used your example now.
I didn't use the 'SessionError extends Error' example yet as this doesn't seem to be used yet in the current develop branch. I'll gladly switch to this kind of errors at a later stage when the changes for the migration get merged.

bilalesi (approved):
LGTM

PR #402, opened 2025-06-02
110/change user journey format

tolokoban:
Here is an example of what we are now sending to AI Agent:

```json
{
    "thread_id": "a7230411f5d447c68c52014125925df1",
    "click_history": [
        {
            "timestamp": 1748348242148,
            "region": "Cerebral nuclei",
            "artifact": "Bouton density"
        },
        {
            "timestamp": 1748863561271,
            "region": "Fastigial nucleus",
            "artifact": "Bouton density"
        },
        {
            "timestamp": 1748866999087,
            "region": "Vestibulocerebellar nucleus",
            "artifact": "Bouton density"
        }
    ]
}
```


bilalesi (approved):
LGTM,

this will introduce breaking changes due to refactoring and renaming of folders/files as part of the migration to entitycore.
could you please remind me when we move the migration to production? or, if I’m unavailable at that time, feel free to make the necessary changes yourself.

PR #403, opened 2025-06-02
Public project artifact tabs system

loris-olivier-obi:
- Change for AntD tables
- Use the same columns as in explore for coherence
- Used a simpler button system for the different artifacts


tolokoban (approved):
LGBT

PR #404, opened 2025-06-02
use entitycore for meshes and point cloud

bilalesi:


src/features/brain-atlas-viewer/context.tsx
```diff
@@ -0,0 +1,74 @@
...
+      asRawResponse: false,
+      id: atlasAssetId,
+    })
+  );
+
+  if (assetError)
+    throw Error(
+      `Failed to download mesh asset (ID: "${atlasAssetId}") for brain region ID "${atlasRegionId}`
+    );
+  const data = arrayBufferToString(asset);
```
pgetta, line 45: Conceptually more correct way to do that is in downloadAsset function so that res.text() is called under the hood which relies on response headers to determine encoding; otherwise we assume that it's utf-8.

Maybe we add a note to address later? But that's very minor.

src/features/brain-region-hierarchy/event.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  if (isBrowser()) {
+    const handler: EventListener = (event: Event) => {
+      if (isBrainRegionClickEvent(event)) {
+        cb(event);
+      }
+    };
+    window.addEventListener(BrainRegionClickEvent, handler, { signal });
+
+    return () => abortController.abort();
```
pgetta, line 38: Oh nice, I've just read about using signals in even listeners. Very cool stuff!

PR #407, opened 2025-06-03
110/change user journey format 3

tolokoban:
When a new user goes to explore section for the first time, their history is empty. And since they haven't clicked any brain region yet, the AI agent does not know what suggestion to produce.

This PR is to add an initial history with the default brain region. Today, it's Cerebrum.


loris-olivier-obi (approved):
lGtM

PR #412, opened 2025-06-04
Public projects corrections ayima feedback

loris-olivier-obi:
Correction based on ticket #110 https://github.com/openbraininstitute/prod-obi-website/issues/110


tolokoban (approved):
OK

PR #446, opened 2025-06-16
Resize AI Assistant panel

tolokoban:
We can now:
1. Minimize the panel
2. Maximize the panel
3. Drag to change the width of the intermediate size

If you drag more than 80%, it goes fullsreen.


loris-olivier-obi (approved):
LGTM

PR #459, opened 2025-06-18
Fix/linter 1

tolokoban:
I tried to fix the 300+ linting errors.
When in doubt, I put a `@FIXME:` comment in the code.

The linter is now happy, but Typescript is still not.
I tried to fix it too, but there are too many places where I do not what choice to make. Mostly when there are mixing between Nexus and EntityCore properties.


bilalesi (commented):
🫡 for you fabien, i left some comments; 
I did not finish all the files yet, i will continue tomrw

PR #493, opened 2025-06-25
use cell-composition asset instead the endpoint to get the summary

bilalesi:
- use cell-composition asset instead the endpoint to get the summary
- add seo to explore detail page (short but better)
- fix asset download endpoint overloading types


tolokoban (approved):
LGTM

PR #511, opened 2025-06-26
Create the image for the deployment in Azure in a separate GitHub action, which has to be triggered manually for now

DriesVerachtert:
Create the image for the deployment in Azure in a separate GitHub action, which has to be triggered manually for now


bilalesi (approved):
LGTM

PR #516, opened 2025-06-27
Circuit filter side panel

loris-olivier-obi:


package.json
```diff
@@ -104,6 +104,7 @@
     "lexical": "^0.30.0",
     "lodash": "^4.17.21",
     "memoize-one": "^6.0.0",
+    "moment": "^2.30.1",
```
tolokoban, line 107: It seems that this library is deprecated:
> Moment.js is a legacy project, now in maintenance mode. In most cases, you should choose a different library.

see https://www.npmjs.com/package/moment

tolokoban: We agreed to change this lib in a future PR.

tolokoban (approved):

package.json
loris-olivier-obi: I will replace this library in a future PR

PR #546, opened 2025-07-01
Add basic sim UI for small circuits

pgetta:
The change:
* Adds types and queries for CircuitSimulationCampaign, CiccuitSimulation, CircuitSimulationResult, SimulationExecution (as well as Generic Activity).
* Extends NWB trace viewer to support NWBs produced by the simulator (which do not have stimulus recording, but rather only response one).
* Extends tiny circuit simulation page with functionality to run simulations and see the results (this is very basic now) Also, the page component requires some refactoring and splitting into multiple modules - I'm doing that now, will submit a separate PR.

### DIFF
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -1,3 +1,24 @@
+/**
+ * Reads and processes NDJSON (newline-delimited JSON) data from a streaming HTTP response.
+ * 
+ * This function handles streaming responses containing multiple JSON objects separated by newlines.
+ * Each line is parsed as JSON and passed to the optional callback function as it becomes available.
+ * 
+ * @template T The type of the JSON objects expected in each line
+ * @param response The HTTP Response object to read from
+ * @param onMessage Optional callback function called for each parsed JSON object
+ * 
+ * @throws {Error} When the HTTP response status is not ok (status >= 400)
+ * @throws {Error} When the response body is empty or null
+ * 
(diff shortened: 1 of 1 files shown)


src/features/details-view/button-copy-id/button-copy-id.tsx
```diff
@@ -8,11 +8,17 @@ import styles from './button-copy-id.module.css';
...
   tooltip?: string;
+  value: string;
 }
 
-export function ButtonCopyId({ className, value, tooltip = 'Database ID' }: ButtonCopyIdProps) {
+export function ButtonCopyId({
+  className,
+  value,
+  tooltip = 'Database ID',
+  label = 'Copy ID',
```
bilalesi, line 20: there is a hook use src/hooks/useCopyClipboard.ts can be used
i improved it in my next PR to have loading state you check [here](https://github.com/openbraininstitute/core-web-app/pull/532/files#diff-f65e853a4ed6c4bd30ea2fbbbc47d0ee5456c67521cd130a1776f22fc2168c7e)

bilalesi (approved):

src/features/details-view/button-copy-id/button-copy-id.tsx
pgetta: Thank you!

PR #547, opened 2025-07-01
Added descriptions tooltips, and fixed scrolling issue

tolokoban:
Added description as tooltips for all generated inputs.
Fixes issue with scrolling of the columns.

image


src/features/small-microcircuit/types.ts
```diff
@@ -1,6 +1,9 @@
 import { atom } from 'jotai';
 
 import { ConfigValue } from './_components/components';
+import { WorkspaceContext } from '@/types/common';
+
+export type Params = WorkspaceContext & { circuit_id: string };
```
g-bar, line 6: I think we can remove this, it moved to the "page".

tolokoban: Good catch! I missed this during merging/copying.

g-bar (approved):

PR #553, opened 2025-07-01
Enable entity download from the details page

pgetta:


pgetta:
Relates to https://github.com/openbraininstitute/prod-platform-architecture/issues/98

PR #565, opened 2025-07-01
245/fix tooltip alignement

tolokoban:
Added circuit id and circuit name as disabled inputs:

image

Now, the tooltips are more accurate:

image
image

### DIFF
--- a/src/features/small-microcircuit/_components/circuit-name/circuit-name.tsx
+++ b/src/features/small-microcircuit/_components/circuit-name/circuit-name.tsx
@@ -1,19 +1,18 @@
 import React from 'react';
 import { InfoCircleFilled } from '@ant-design/icons';
 
-import { useCircuit } from '../hooks/circuit';
-
+import { ICircuit } from '@/api/entitycore/types/entities/circuit';
 import { classNames } from '@/util/utils';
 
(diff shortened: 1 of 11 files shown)


src/features/small-microcircuit/_components/circuit-name/circuit-name.tsx
```diff
@@ -25,11 +24,12 @@ export default function CircuitName({ className, circuitId }: CircuitNameProps)
   );
 }
 
-function useCircuitNameAndDescription(circuitId: string) {
-  const [name, setName] = React.useState(circuitId);
+function useCircuitNameAndDescription(circuit: ICircuit | undefined | null) {
```
g-bar, line 27: I'm thinking instead of this we just show a global error message, or 404 redirect if the circuit doesn't exist. The whole page is useleess without the circuit, so no point in even rendering it.

g-bar (approved):

src/features/small-microcircuit/_components/circuit-name/circuit-name.tsx
tolokoban: Good point.
I can put a check for the circuit in the main page, above.
I'll do this in another PR.

PR #575, opened 2025-07-02
Restore the CopyCampaign ID button

tolokoban:
The button was gone after the removal of the circuit name


loris-olivier-obi (approved):
LGTM

PR #584, opened 2025-07-02
Add circuit simulaiton input/output files

pgetta:
This is rebased version.


bilalesi:
> @bilalesi I've made a change according to your suggestion for the downloadAsset call.

apparently it's not in this PR

pgetta:
I guess we need to separate labels and colors so this can be re-usable in different places.

bilalesi (approved):

PR #596, opened 2025-07-03
219/ai suggestions do not take current region into account

tolokoban:
In this PR, we focused in two things:

1. making sure we get the current selected region when we ask for suggestions
2. fixing the JSON for the hardcoded suggestions, to always have 3 items and not 1 or 2.

In another PR, I will investigate deeper to see if we always ask for suggestions bases on the selected region and artifact.
Because it seems that with rapid clicks, we can confuse the app, right now.


loris-olivier-obi (approved):
LGTM

PR #615, opened 2025-07-07
Fix suggestions for root brain region

tolokoban:
We were using an old way (pre-entitycore) to check if a region is the root region.


loris-olivier-obi (approved):
LGTM

PR #634, opened 2025-07-09
Fix/website lcp

bilalesi:
fix [#44](https://github.com/openbraininstitute/prod-virtual-lab/issues/44)


pgetta:
It seems that the current change is not related to LCP. Maybe something went wrong with the branch?

bilalesi:
> It seems that the current change is not related to LCP. Maybe something went wrong with the branch?

there was another PR was merged on top of this as this was the base, 
for the branch name is wrong, i started to look into lcp and then switched to small issues

PR #636, opened 2025-07-10
Prompt can have multilines and we cannot send empty messages

tolokoban:
To add a new line in the prompt, you can use Shift-Enter.
Moreover, no empty message can be sent now.


loris-olivier-obi (approved):
LGTM

PR #642, opened 2025-07-14
Website pricing refinment

loris-olivier-obi:
- Added string or null for id in pricing plan list to fix the typeguard issue
- Aligned the blocks for individual pricing


tolokoban (approved):
LGTM

PR #648, opened 2025-07-15
246/add url for ai chat

tolokoban:
When calling the chat, we add a n ew parameter to tell them what URL we are in. That will help the agent have a context.


loris-olivier-obi (approved):
LGTM

PR #649, opened 2025-07-15
Fixed the issue with textarea auto-growth

tolokoban:
When adding new lines, the prompt was not increasing its since. Therefore, it looked like the prompt was empty.


loris-olivier-obi (approved):
LGTM

PR #653, opened 2025-07-16
Fix the new thread feature

tolokoban:
There was an issue when the user clicked on "Clear Chat".
The newly created thread being empty by default, it was not visible in the history.


loris-olivier-obi (approved):
LGTM

PR #656, opened 2025-07-17
fix hydaration mismatch for ai assistant

bilalesi:
due the change of the ai-assistant panel width by using the local storage, the first render in the server is different from the client render, this is a fix is for:
- prevents `content flash` during hydration
- ensures content is only rendered after the component has been fully hydrated on the client side


tolokoban (approved):
Thanks!

PR #659, opened 2025-07-18
feat: add cloudfront cdn to the app

bilalesi:


.github/workflows/publish-production-image.yml
```diff
@@ -64,10 +66,43 @@ jobs:
...
+          # create temporary container and extract static assets
+          docker create --name assets-extractor ${{ vars.PUBLICECR_URI }}:${{ env.NEW_TAG }}
+          docker cp assets-extractor:/app/.next/static ./assets-for-s3/${{ steps.tag.outputs.version }}/_next/static/
+          docker cp assets-extractor:/app/public ./assets-for-s3/${{ steps.tag.outputs.version }}/public/
+          docker rm assets-extractor
+
+      - name: Sync static assets to S3 and invalidate CloudFront
+        run: |
+          # upload extracted static assets to S3
+          aws s3 sync ./assets-for-s3/public s3://${{ vars.PRODUCTION_S3_BUCKET_NAME }}/public --delete --cache-control "public, max-age=2592000, immutable"
```
pgetta, line 101: Maybe for public data max age should be smaller:

Since the files there do not have hash in the filename, once cached by the client/browser - we don't have any control.

The updates to these files are rare, but if some image will be eventually updated - it will take a month for the change to appear on the client machine (if previous version is already cached).

I've just checked and "stale while revalidate" might be a good solution for that, but unfortunately AWS Cloudfront doesn't support that.

Another option is to rename the image if it needs to be updated to serve it immediately. But people forget stuff.

What do you think?

pgetta (commented):
Nice work, Bilal,

I've left one question about caching policy for the `public` folder, please see below.

.github/workflows/publish-production-image.yml
bilalesi: good catch @pgetta 
I update it to have [--exact-timestamps](https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html), I think this is enough to compare the files (so now not only the file and size is used but also the modified date)
LMKWYT?

for ref: this is the terraform PR, is till need some improvement
https://github.com/openbraininstitute/aws-terraform-deployment/pull/473

fyi: there will be another PR that 
- update the images to use the basePath (which is the cdn for images), public folder will not be automatically using the `assetPrefix` in `next.config.ts`
- update the image headers to include `Cache-Control` with value: `public, max-age=xxx, must-revalidate` to allow the CDN to cache them

@pgetta

pgetta (dismissed):

pgetta:
It seems this needs some merge conflict resolution, I'll re-approve when needed.

tolokoban (approved):

PR #665, opened 2025-07-21
Provide workspace context for experimental data views

pgetta:
The workspace context is not provided in the explore detailed views. The issue originally reported in https://github.com/openbraininstitute/prod-human-data-and-models/issues/6#issuecomment-3089471036.

This change adds missing context for experimental data views (mostly exp morphologies and traces) so that user generated content can be loaded.

@bilalesi feels like we should find a better way to provide context instead of doing prop drilling through the whole tree of components. Perhaps via a global state initialised with some server component? But that will require something else for the nested server components to work.
If we don't have a good solution for the time being - I propose to merge this to unblock Ilkan, and seek for it afterwards.


bilalesi (approved):

pgetta:
> for nested components, we can just use `useParams<WorkspaceContext>()`, without passing the context down the tree

Ah, yes, I'll make another PR with corresponding changes. Thank you Bilal.

PR #669, opened 2025-07-22
Improve simulation details view

pgetta:
Changes:
* Switch trace thumbnail to use actual plot (non-interactive, with downsampled data). The plot is using inView dependency to conditionally render itself, so that many simulations within a single model are not going to have any bad performance impact.
* Remove Copy simulation ID from the list of recordings in the simulation details page (there is already button for that in the header of the page).
* Add copy simulation ID button in the list of simulations on the model details page.
* Simulation results layout on th model details page, now it switches to having only one grid column on smaller screens.
* Add a link to simulation details view from the list of simulations (within model details page).


pgetta:
Relates to https://github.com/openbraininstitute/prod-explore-functionality/issues/190.

PR #681, opened 2025-08-04
feat: circuit using entitycore 

bilalesi:


bilalesi:
please, if anyone can test it locally that would be good, just testing that `circuit-dev` route in explore are not accessible for staging and production only in dev (`NEXT_PUBLIC_DEPLOYMENT_ENV`)

tolokoban (approved):
I have tested in my local environment.
With NEXT_PUBLIC_DEPLOYMENT_ENV=
* `development`: I can get the list of circuits even by replacing `circuit` with `circuit-dev`
* `staging`: `circuit-dev` is not accessible.

PR #682, opened 2025-08-04
feat: filter electrical cell recording by recording-origin in explore/experimental data

bilalesi:
in explore -> experimental, electrophysiology is filtered by the `recording_origin: in_vitro`,

Note that the electrophysiology (electrical cell recoding) will be added in the model data too with the filter of `recording_origin: in_silio`, this change will require to change the entity count components and also the domain `src/entity-configuration/domain/experimental/electrical-cell-recording.ts`

for future changes:

1. `DataType` should be renamed to `ExtendedTypes`.
2. create a new domain entity: (`ElectricalCellRecordingInVitro`, `ElectricalCellRecordingInSilio`) this will allow us to use the same entity model api and diverge when needed as listing page 
3. use DataType (ExtendedTypes) in the counter instead of the `EntityTypes`

### DIFF
--- a/src/api/entitycore/types/entities/electrical-cell-recording.ts
+++ b/src/api/entitycore/types/entities/electrical-cell-recording.ts
@@ -18,13 +18,74 @@ import type {
   EntityCoreOwnership,
 } from '@/api/entitycore/types/shared/global';
 
+const RecordingType = {
+  Intracellular: {
+    key: 'intracellular',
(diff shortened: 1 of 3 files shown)


tolokoban (approved):
LGTM

bilalesi:
@tolokoban i close this one in favor of the new one https://github.com/openbraininstitute/core-web-app/pull/685

PR #695, opened 2025-08-11
Url builder: Throw error for non existing entities

g-bar:
For entities that are not supported in core-web-app, the resolveDetailUrl function returns an url like

https://staging.openbraininstitute.org/app/virtual-lab/explore/**undefined**/**undefined**/04e95bdd-1427-454a-ab3e-d6e0706d5b1a

Which is a bit odd, this makes the function throw an error for such cases.


bilalesi (approved):
👍

PR #707, opened 2025-08-13
allow extensible types other than default from entitycore

bilalesi:
this is the first step to have extensible types

**NOTE**

- entitycore has the default types which you can find in src/api/entitycore/types/entity-type.ts
but now we have different subtypes that should be treated as first class citizen so in this PR we extend the types of entity core with new one that are basically scale (ex: for circuit) or origin (for cell-recording) (may be others later) properties
- this also will get rid of the legacy types that are renamed by entitycore which help us to maintain only one set of types
- this PR will have need another PR to use Dict instead of Enum which help extension of the original types rather then creating another Enum, just to keep on object to maintain
- this change also require a change in virtual lab to use new types for bookmarks


src/api/entitycore/types/extended-entity-type.ts
```diff
@@ -0,0 +1,30 @@
+// NOTE: this types has changed be named to "ExtendedEntitiesType"
+// NOTE: rename the old ones to entity core naming and update
+// TODO: update virtual-lab-api with the new names
+// NOTE: There are now various nested types associated with a single parent type in EntityCore.
+// We should enable selection and data manipulation for both the nested types and those directly defined in EntityCore.
+export enum ExtendedEntitiesType {
+  ExperimentalBoutonDensity = `experimental_bouton_density`,
+  ExperimentalNeuronDensity = `experimental_neuron_density`,
```
bilalesi, line 8: in the next PR this will be a dict same as `EntityType` to allow extending rather then change in both source

PR #730, opened 2025-08-20
allow to deploy to next staging for v2

bilalesi:


bilalesi:
do the new dev will erase next.staging ?
if so then this PR is not needed, 
IMHO: we should keep both, so at least QA can test both until we finish the migration

pgetta:
Yeah, that's pretty much the change, the "next staging" becomes "dev".

But if you need to keep next staging for the v2 - I can set it up in coming days (that would be an easy solution).

We keep coming back with the need to have separate deployments for some features/branches, maybe long term we can consider setting up something alike to "Review apps", where a new deployment can be created from a feature branch. I presume it makes sense to start thinking about that only after migration to Azure.

How does that sound?

bilalesi:
I will close this PR

PR #735, opened 2025-08-20
Exclude myelin from synapse target selector

pgetta:
Selecting myelin as a target for synapses is not meaningful so this change excludes that from the selector.

Relates to https://github.com/openbraininstitute/prod-singlecell-simulation/issues/103.


bilalesi:
this should also applied in v2

pgetta:
Thank you, @bilalesi. Now it's actually not clear, last time Darshan suggested to keep the myelin in the selector and fix the underlying issue so that people can generate synapses on it.

I'll keep this for some time and if not needed - will close it.

P.S. I should have put that notification in advance.

PR #750, opened 2025-08-26
Restore panel width for internal links

tolokoban:
When we click on an internal link, we do not open it in another tab.
Instead, we stay in the same tab and the screen changes.
The problem is that is the AI Assistant panel is overlapping the rest of the screen, we do not notice that something has changed.

That's why, now, we restore the panel width to the minimum as soon as an internal link is clicked.


loris-olivier-obi (approved):
LGTM

PR #751, opened 2025-08-26
Add github action which runs pnpm audit and stores it as a build artifact

DriesVerachtert:
@pgetta  or @bilalesi , would it be OK to merge this PR?
It basically adds an 'on demand' github action to store the output of pnpm audit in some build artifact. It doesn't affect the regular builds and it doesn't fail any builds if some security issue is found.


pgetta (approved):
Thank you, Dries, looks good.

PR #773, opened 2025-09-02
Added tool thumbnail-generation-electricalcellrecording-getone

tolokoban:
This tool do return images in the same exact way that tool `thumbnail-generation-morphology-getone` does:
1. The tool returns a file identifier.
2. We use it to ask `storage/${fileIdentifier}/presigned-url` backend entrypoint for the AWS bucket URL.
3. We fetch from the bucket and read the Response Header `X-Amz-Meta-Category` to get the mime type.


loris-olivier-obi (approved):
LGTM

PR #778, opened 2025-09-03
Enable preview environment (via AWS)

pgetta:


bilalesi (approved):

pgetta:
Thank you, @bilalesi. It's mostly done, the only part that is left is configuring the `preview.` subdomain, I'm waiting for the answer from Dries on who/when can address that.

Will let you know when this works.

PR #796, opened 2025-09-08
Circuit detail + small microcircuit

g-bar:
This adds the circuit detail, so far it only ports the old content, I'm working on a follow up PR to change the styles to closely match the new design.


loris-olivier-obi (approved):
LGTM

PR #813, opened 2025-09-12
Move model creation to backend

pgetta:


src/api/small-scale-simulator/index.ts
```diff
@@ -2,9 +2,10 @@
 export { runSimulation as runCircuitSimulation } from './circuit/simulation';
 
 // Single neuron scale
+export { createModel as createSingleNeuronSynaptome } from './single-neuron/synaptome';
+export { createModel as createSingleNeuronModel } from './single-neuron/single-neuron';
```
bilalesi, line 6: small thing, can use path aliases everywhere

bilalesi (approved):
clean !
fyi: there will be quite of changes in `v2`, but it's well organized so probably it only need imports fixes, and get rid of atoms,

src/api/small-scale-simulator/index.ts
pgetta: Ah, missed that.

I'll make sure these are used when will replicate the change in v2 branch.

PR #816, opened 2025-09-12
migrate from reconstruction to cell morphology

bilalesi:


pgetta (dismissed):

bilalesi:
@pgetta 
entity core has changed also the `strain` to be within `subject`, so i think this is the final change to use cell-morphology 
can you have-a-look/re-approve it again please

g-bar (approved):

pgetta (approved):
Looks good, thank you @bilalesi

PR #820, opened 2025-09-15
New UX: Detail view downloads and fixes

g-bar:
Fixes downloads for circuit (shows panel).
Fixes dowloads for synaptome, single neuron simulation and single neuron synaptome simulation.
Improves close button behaviour.


src/entity-configuration/definitions/fields-defs/common.tsx
```diff
@@ -260,3 +254,26 @@ export const FieldsDefinition: Partial<FieldsDefinitionRegistry<EntityCoreObject
     isDisplayable: true,
   },
 };
+
+function DownloadButton({ entity }: { entity: EntityCoreObjectTypes }) {
+  const [, setDownloadPanelCircuit] = useAtom(downloadPanelCircuitAtom);
+
+  return (
```
bilalesi, line 261: i don't know why there is a new atom; there is the event that handle that ?

src/ui/segments/explore/circuit/elements/download-panel/index.tsx
```diff
@@ -31,15 +31,6 @@ export function DownloadPanel() {
...
   useHotkeys('Escape', onClose);
-  useEffect(() => {
-    const unsubscribe = customRowSelectionEventListener<ICircuit>((event) => {
-      setCircuit(event.detail?.record ?? null);
-    });
-
-    return () => {
-      unsubscribe();
-    };
-  }, []);
```
bilalesi, line 42: this is related to the event

src/ui/segments/mini-detail-view/index.tsx
```diff
@@ -4,14 +4,15 @@ import { useMutation } from '@tanstack/react-query';
 import { match, P } from 'ts-pattern';
 import { useState, useEffect } from 'react';
 import { Image } from 'antd';
+import { useAtom } from 'jotai';
 
 import kebabCase from 'lodash/kebabCase';
 import Link from 'next/link';
-
+import { downloadPanelCircuitAtom } from '../explore/circuit/elements/download-panel';
```
bilalesi, line 11: better to stick to path aliases

bilalesi:
I will approve it, but please can you check the listing page download panel

src/entity-configuration/definitions/fields-defs/common.tsx
g-bar: Yes, I agree about moving async atoms to tanstack.

Actually, I agree with you in that I prefer native implementations most of the time. Except when a library simplifies the code. Complexity is subjective, I personally struggled to understand it, which was the main motivation to simplify it; however this one does require much fewer lines code which is a marker of complexity.

I have indeed also fixed it on the list view:

Capture d’écran 2025-09-18 à 11 04 15

PR #827, opened 2025-09-17
Fix me-model selection for synaptome build

pgetta:
It turns out the me-model selection during the synaptome build was broken.

@bilalesi I've fixed it, but not sure that's the correct way with the current changes.

This is being addressed in the scope of https://github.com/openbraininstitute/prod-platform-architecture/issues/140 but not really related to that ticket.


pgetta:
Yeah, I'm not sure I understand that either. Thank you!

PR #837, opened 2025-09-18
191/ionic current recording 2

tolokoban:
I tested with the development and the staging versions of BlueNaas. It's compatible with both formats.

* The logic that splits the data and convert them for the plotting is in `src/features/entities/neuron-simulation/experiment/visualization/plots-parser.ts`.
* The component that displays all the plots is in `src/features/entities/neuron-simulation/experiment/visualization/multi-plots-view`.
* I had to take the legend outside of the plots, because in some cases they were overlapping with the lines and sometime they were removed.

image

I hope I haven't broken too many things.


pgetta (approved):
Looks good, thank you Fabien.

PR #840, opened 2025-09-19
Small microcircuit view campaign in read only mode and duplicate. Hide related publications.

g-bar:


src/api/entitycore/types/entities/circuit-simulation-campaign.ts
```diff
@@ -32,6 +33,7 @@ interface ICircuitSimulationCampaignBase {
   scan_parameters: { [key: string]: any };
   entity_id: string;
   simulations?: Array<ISimulationBase>;
+  circuit: ICircuit;
```
bilalesi, line 36: may be i missing smthg, are you sure the simulation campaign now return circuit, i just checked entitycore staging, and it's not ?

src/ui/segments/mini-detail-view/index.tsx
```diff
@@ -402,7 +415,7 @@ function ExploreActions<T extends EntityCoreObjectTypes>({ record }: { record: T
         className="hover:bg-primary-7/40 h-12 border border-white/16 px-10 font-bold shadow-[8px_8px_20px_0px_#0000005C,-12px_-8px_32px_0px_#FFFFFF1F]"
       >
         <Link
-          href={`${ROOT_ROUTE}/${virtualLabId}/${projectId}/data/view/${kebabCase(record.type)}/${record.id}`}
+          href={`${ROOT_ROUTE}/${virtualLabId}/${projectId}/data/view/${kebabCase(dataType)}/${record.id}`}
```
bilalesi, line 418: do we really need the `dataType` here ?, If I'm not mistaken, i think `record.type` be enough to get the data ? (extended type can not really have mini-details view as are a extracted from a parent type that comes from `record.type`

src/entity-configuration/domain/simulation/paired-neurons-simulation.ts
```diff
@@ -157,9 +157,9 @@ export const PairedNeuronCircuitSimulation: EntityCoreTypeConfig<ICircuitSimulat
   asset: {
     extension: 'application/json',
   },
-  detailViewSections: ['overview', 'related-publications'],
+  detailViewSections: ['overview'],
   isBookmarkable: false,
   isDownloadable: false,
   isCopyable: true,
-  isSimulatable: true,
+  isSimulatable: false,
```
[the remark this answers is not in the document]
bilalesi: yeah, u are right, it's actully a sim, 🤦🏻

src/ui/segments/workflows/elements/workflow-activity.tsx
```diff
@@ -9,6 +9,7 @@ import find from 'lodash/find';
 import get from 'lodash/get';
 
 import type { ColumnsType } from 'antd/es/table/interface';
+import { useRouter } from 'next/navigation';
```
bilalesi, line 12: may be using `import { useRouter} from "@bprogress/next"` can help showing the progress on top so the user can see the page loading

src/ui/segments/workflows/elements/workflow-activity.tsx
```diff
@@ -348,23 +359,26 @@ export function WorkflowActivity({
...
-                  <a
-                    href={resultsLink}
-                    target="_blank"
-                    rel="noopener noreferrer"
-                    aria-disabled={activityType === ActivityValues.Build}
-                  >
-                    View results
-                  </a>
-                </Button>
+                {entityType !== 'small_microcircuit_simulation' &&
```
bilalesi, line 362: better to use `ExtendedEntitiesTypeDict.SmallMicro...`, this will help when there will be renaming of things, 
unify using the types in one way rather then using the explicit name

src/ui/segments/workflows/elements/workflow-activity.tsx
```diff
@@ -348,23 +359,26 @@ export function WorkflowActivity({
...
+                  entityType !== 'paired_neuron_circuit_simulation' && (
```
bilalesi, line 363: same here

src/api/entitycore/types/entities/circuit-simulation-campaign.ts
g-bar: Not sure how to handle this case, the simulation campaign doesn't have the circuit but when we fetch we use this 
https://github.com/openbraininstitute/core-web-app/blob/f08ed873df27faad8bdd2c932f392f932ae1cbc6/src/entity-configuration/domain/simulation/simulation-campaign.ts#L86

function which extendes it to include the circuit. Perhaps we should add it to the interface but make it optional?

bilalesi: i would say create another type that handle this case, the types within /entitycore/types/xxx, should not be modified as it's related to the domain (entitycore), `resolveSimulationByCampaignId` should extend that

src/ui/segments/workflows/elements/workflow-activity.tsx
g-bar: Wasn't aware of bprogress/next, that's pretty neat! Thanks.

src/api/entitycore/types/entities/circuit-simulation-campaign.ts
g-bar: done

src/ui/segments/detail-view/overview.tsx
```diff
@@ -73,6 +75,32 @@ export default async function Overview({
     (entity as ISingleNeuronSynaptome).me_model = meModel; //eslint-disable-line
   }
 
+  if (
+    extendedType === 'small_microcircuit_simulation' ||
+    extendedType === 'paired_neuron_circuit_simulation'
```
bilalesi, line 80: preferably using the `ExtendedType...` as mentioned below

g-bar: crap, forgot that file, done.

g-bar:
Thank you Bilal.

PR #852, opened 2025-09-24
make derivation type required

bilalesi:
- use `DerivationTypeDictionary` instead of DerivationType
- use unspecified derivation type for experimental traces


bilalesi:
thanks @tolokoban

PR #857, opened 2025-09-24
Using GLB for meshes and new painter for point clouds

tolokoban:
image

The old viewer was rendering this:
image


tolokoban:
Thanks @loris-olivier-obi !

PR #861, opened 2025-09-25
Embedding tarball of morphoviewer

tolokoban:
Flat morphologies were not displayed in the 3D viewer.
See this [issue](https://github.com/openbraininstitute/prod-explore-functionality/issues/301) for more details.

We fixed the issue in `@bbp/morphoviewer` and created a tarball that we embedded here.
The reason is that we cannot push to `npm.js` from the code that is now in OBI repo.

Another solution would be to have a repository for OBI on https://npmjs.org


tolokoban:
Thanks @loris-olivier-obi !

PR #875, opened 2025-09-26
incrase page size

g-bar:


src/ui/segments/detail-view/related-publications.tsx
```diff
@@ -65,7 +65,7 @@ function PerTypePublications({
   }>({
     loading: false,
     page: 1,
-    pageSize: 5,
+    pageSize: 20,
```
bilalesi, line 68: There is default values for page size you may use it in src/constants

bilalesi: With that we can have consistent pagination across pages

bilalesi (approved):

PR #880, opened 2025-09-29
313/count filters not displayed

tolokoban:
For range filters, the widget with the number of active filters was not updates.
It was due to the function `isEmpty` that does not work with numbers.


tolokoban:
Thanks Bilal

PR #893, opened 2025-10-01
191/optimize ionic current plots

tolokoban:
* Fix color mismatch between plot and legend
* Optimize plots rendering to prevent browser freeze
* Report config error when Save button is disabled
* Refresh the plots when resized


loris-olivier-obi (dismissed):
LGTM

PR #923, opened 2025-10-08
added SFN 2025 section

jdcourcol:
This is a new page in preparation for SFN.


jdcourcol:
I would need that in staging, and then see how we push that in prod.

bilalesi (approved):

jdcourcol:
I put some text in staging sanity already. but this is fine for now. It looks like this:
grafik

PR #964, opened 2025-10-20
Move logout to account panel

loris-olivier-obi:


bilalesi:
@loris-olivier-obi you have an issue with prettier again, so many things should not change, it add complexity to review and file changes

PR #973, opened 2025-10-22
fix: return to data scope from detail view

bilalesi:


bilalesi:
@tolokoban  thank you !

PR #990, opened 2025-10-24
Address circuit simulation UI feedback

pgetta:
* Labels for scan params show only the param name after last ".", e.g. `extracellular_calcium_concentration` instead of  `initialize.extracellular_calcium_concentration` with the full name on hover
* Add select/deselect all control
* Use Intl.Collator to sort simulations - for natural ordering
* When selecting a simulation - try to show already selected file for the corresponding circuit, if it doesn't exist - show a circuit config
* When changing a file selection - the viewer shows the old content (blurred and with a loader) until the new content is fetched
* Other small changes/refactoring


pgetta:
Thank you, Fabien

PR #996, opened 2025-10-27
Fix circuit simulations NWB viewer, other minor changes

pgetta:
* Fix circuit simulations NWB viewer
* Rename "Check all" -> "Select all"
* Correct types

### DIFF
--- a/src/features/small-microcircuit/_components/atoms/index.ts
+++ b/src/features/small-microcircuit/_components/atoms/index.ts
@@ -149,7 +149,7 @@ export const circuitAtomFamily = readAtomFamilyWithExpiration(
   }
 );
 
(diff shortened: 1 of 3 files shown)


src/features/small-microcircuit/_components/atoms/index.ts
```diff
@@ -163,7 +163,7 @@ export const fileAtomFamily = readAtomFamilyWithExpiration(
     assetPath?: string;
     context: WorkspaceContext;
   }) =>
-    atom<Promise<ICircuit>>(async () => {
+    atom<Promise<any>>(async () => {
```
g-bar: Maybe we could write types for the actual shape of the json?

PR #1013, opened 2025-10-29
Fix circuit simulation file auto-selection

pgetta:
Auto-selection was not working for the output files, this PR fixes the issue as well as simplifies the implementation.


pgetta:
Thank you, Fabien!

PR #1022, opened 2025-11-03
Enable ME-model circuit simulation (unified UI)

pgetta:
This PR makes it possible to create and launch a simulation campaign using obi-one with a similar UI used in circuit simulations. For the moment this is going to be available only in dev.

Related to https://github.com/openbraininstitute/prod-singlecell-simulation/issues/200.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -2,7 +2,7 @@ KEYCLOAK_ISSUER=https://staging.openbraininstitute.org/auth/realms/SBO
 KEYCLOAK_CLIENT_ID=obp-core-web-app-dev
 KEYCLOAK_CLIENT_SECRET=dummy-secret
 
-NEXT_PUBLIC_DEPLOYMENT_ENV=local
+NEXT_PUBLIC_DEPLOYMENT_ENV=staging
 
 NEXTAUTH_SECRET=dummy-secret-for-dev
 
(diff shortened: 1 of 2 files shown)


src/features/small-microcircuit/_components/atoms/index.ts
```diff
@@ -150,11 +153,23 @@ export const simulationsByCampaignIdAtomFamily = readAtomFamilyWithExpiration(
...
+      const modelEntityBase = await getEntity(params);
+
+      return match(modelEntityBase.type)
+        .with(EntityTypeDict.Circuit, () => getCircuit(params))
+        .with(EntityTypeDict.Memodel, () => getMEModel(params))
+        .otherwise((entityType) => {
+          throw new Error(`Unsupported model entity type ${entityType}`);
+        });
     }),
   {
```
bilalesi: probably switching to tanstack query ? not required though,

.env.development
```diff
@@ -2,8 +2,9 @@ KEYCLOAK_ISSUER=https://staging.openbraininstitute.org/auth/realms/SBO
 KEYCLOAK_CLIENT_ID=obp-core-web-app-dev
 KEYCLOAK_CLIENT_SECRET=dummy-secret
 
-NEXTAUTH_SECRET=dummy-secret-for-dev
+NEXT_PUBLIC_DEPLOYMENT_ENV=local
```
bilalesi: can we just mention in the dev channel, it's important also to not push sentry logs

bilalesi (approved):

PR #1023, opened 2025-11-03
Added fallback for font height in Google Chrome

tolokoban:
`measure.emHeightAscent` is not defined in Google Chrome.


loris-olivier-obi (approved):
LGTM

PR #1031, opened 2025-11-04
Various fixes for circuit sim UI

pgetta:
* Fix output files async loading - so that preselection cat pick up the voltage trace
* Make sure NWB viewer is re-rendered on entity change


pgetta:
Thanks again!

PR #1032, opened 2025-11-04
189/fix colors and injection for simulation protocol 2

tolokoban:
-    Always use white as color for injection
 -   Reduce the "Add recording" dialog to let the user still see where they have clicked

image


loris-olivier-obi (approved):
LGTM

tolokoban:
Thank you @loris-olivier-obi !

PR #1034, opened 2025-11-04
Add recording has now more details

tolokoban:
When the user clicks a neurite, the info panel disappear and the dialog get all the details.

image


loris-olivier-obi (dismissed):
LGTM

PR #1047, opened 2025-11-06
Added buy option to credit-transfer-modal

loris-olivier-obi:


bilalesi (approved):
thanks @loris-olivier-obi 
nice job, 
i don't know if it's temporary because i noticed some space between the tabs on top, and the content, the model is shrinking/expanding when changing tabs and proceeding to checkout, but can be fixed later

PR #1055, opened 2025-11-07
189/rest zoom on all dendrites

tolokoban:
Now the rest camera position will show all the dendrites. The axon and myelin can be out of the viewport, but the max zoom will always include them.

There is also a new attribute on the viewer component: `disableElectrodes`.

When set, the viewer will not display the labels for recordings and injection anymore. And the user won't be able to click to add a recording nor move the injection.


tolokoban:
Thanks @loris-olivier-obi

PR #1082, opened 2025-11-11
Another issue with synapses

tolokoban:
The code is ugly, and I will take some time to improve it after SfN.
But this PR should fix the bugs that we don't already have in production.

One of this known bug that we also have in prod is the following:

1. Build -> Synaptome
2. Add several synapse sets  
image
3. Click __Build Synaptome__
4. Click on __Simulate__
5. Go to __Synaptic Input__
6. You cannot see the color in the __Synapses Group__ selector  
image
7. And if you open the dropdown, you see just blue  
image


loris-olivier-obi (approved):
LGTM

tolokoban:
Thank you @loris-olivier-obi

PR #1084, opened 2025-11-12
[Chat] Code syntax highlighter

jankrepl:
Closes https://github.com/openbraininstitute/prod-explore-functionality/issues/383

### DIFF
--- a/src/components/ai-assistant/message-item/message-item.tsx
+++ b/src/components/ai-assistant/message-item/message-item.tsx
@@ -41,6 +41,9 @@ function MessageChild({
   debug: boolean;
 }): React.ReactNode {
   const { setPanelWidth } = usePanelWidth();
+  const deferredContent = React.useDeferredValue(value.content);
+  const isContentPending = value.content !== deferredContent;
+
   switch (value.role) {
     case 'user':
       return (
@@ -57,15 +60,17 @@ function MessageChild({
       return (
         <>
           <ToolsProgress message={value} />
-          {value.content.trim().length > 0 && (
-            <GithubFlavorMarkdown
-              className={styles.markdown}
-              onLinkClicked={(external) => {
-                if (!external) setPanelWidth(MINIMAL_PANEL_SIZE);
-              }}
-            >
-              {value.content}
-            </GithubFlavorMarkdown>
+          {deferredContent.trim().length > 0 && (
+            <div style={{ opacity: isContentPending ? 0.8 : 1, transition: 'opacity 0.2s' }}>
+              <GithubFlavorMarkdown
+                className={styles.markdown}
+                onLinkClicked={(external) => {
+                  if (!external) setPanelWidth(MINIMAL_PANEL_SIZE);
+                }}
+              >
+                {deferredContent}
+              </GithubFlavorMarkdown>
+            </div>
           )}
           {!hideTools && (
             <>
--- a/src/components/github-flavor-markdown/highlighter.tsx
+++ b/src/components/github-flavor-markdown/highlighter.tsx
@@ -1,7 +1,12 @@
 import { PropsWithChildren } from 'react';
 import { BundledLanguage } from 'shiki';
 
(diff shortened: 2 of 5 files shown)


tolokoban:
I tested it locally, but it kills my browser  :-)

The kind of prompt that got me in trouble is this one:

```
Give me the Python code to plot this function: y = cos(x*x) + 1
```

After the answer, I ask this:

```
Can you convert this into Javascript?
```

And when the result is a long code, the browser freezes.

tolokoban:
> I don’t think the code-block i mentioned above will have any issues with streaming. also `shiki` is much more lightweight compared to `react-syntax-highlighter`, so using both libraries for essentially the same purpose isn’t optimal,

I agree on the fact that we should use only one library. Since we already have one, we should definitively use this one.

But we still may need to have a specific component for the AI.
When the AI agent gives us a piece of code, we receive it bit by bit. And then, the component is re-rendered for every new chunk of few chars we receive. That means that `shiki` will have to parse the code hundred of times.

Unless `shiki` has a debouncer function, of course.

tolokoban:
@jankrepl I have pushed a bit of code for the debouncing.
With this, the browser do not freeze anymore, but the code is not highlighted before the streaming stops.

Feel free to remove this if you find a better way to avoid the issue.

And, if you could use `shiki` instead of `react-highlioght` as Bilal suggested, it would be great.
Thanks for your PR. The result is quite cool to look at :-)

bilalesi:
btw: i could use the code-blocks i shared this the changes i did 
```ts
import { CodeBlock as DCodeBlock } from '@/ui/molecules/code-blocks';
import { BundledLanguage } from 'shiki';
function Block({ children }: PropsWithChildren) {
  const codeElement = children as React.ReactElement<{ children?: string; className?: string }>;
  const codeClassName = codeElement.props.className;
  const languageMatch = codeClassName?.match(/language-(\w+)/);
  const language = (languageMatch ? languageMatch[1] : 'text') as BundledLanguage;
  const code = codeElement.props.children || '';

  return <DCodeBlock code={code} language={language } />;
}
```
```ts
GithubFlavorMarkdown({}){
....
 pre: Block,
})
```

i have tested it with your prompts and it worked fine without any lagging, i have to confess is a BIT slow, but memoization of the markdown and this component should add some more optimization

tolokoban:
I have tested your piece of code @bilalesi , thanks for sharing!

Shiki is indeed more optimized and by just using it, the browser do not freeze anymore.

Now, we can use it directly or embedded in the debouncer.
With the debouncer, we avoid to perform the code parsing/highlighting during the message streaming. We just do it, once the message is complete.
The drawback is that the code is not highlighted "on the fly".

tolokoban:
Oh! What I meant is just that **without** debouncer, we have the highlighting on the fly as expected.
The code appears piece by piece, and it's properly highlighted.

But if we add the debouncer, then while the code appears, it is not highlighted because we use a simple `<pre>` and not the Shiki component. It will be highlighted only when complete (replaced by the Shiki component).

PROS: the code is parsed only once
CONS: we don't see it highlighted until we get to the end.

tolokoban:
So it's related on how we are using **Shiki**.
Ask @bilalesi for help on this: I'm not familiar enough with **Shiki**.

jankrepl:
Note that unfortunately the problem persists even after merging the latest `develop` that introduced some performance optimization in https://github.com/openbraininstitute/core-web-app/pull/1103.

Note that my issue (snippets not being displayed fully) happens within one AI message.

jankrepl:
Note that when I set the `experimental_throttle: 200` in `useChat` I am way less likely to hit that issue. That is why my intuition is that this is again related to the fact that certain components render way too often without being memoized.

bilalesi (dismissed):

jankrepl:
I confirm that the issue disappeared:) thanks @bilalesi

bilalesi (approved):

tolokoban (approved):

PR #1085, opened 2025-11-12
Show morphometrics for entities without legacy_id

pgetta:


pgetta:
@bilalesi I'm not sure about the reason why we had that switch, but seems like it's not needed anymore. Could you please confirm?

loris-olivier-obi (approved):
LGTM

PR #1094, opened 2025-11-13
Feedback service with GitHub part 1

loris-olivier-obi:
- Implementation of a new icon in project to provide feedbacks
- New form in a modal to provide feedbacks
- User, for now, can specify which section and the type of feedback they want to give
- Feedbacks will be pushed in the feedback repository on the openbraininstitute organisation in Github


src/app/api/feedback/create-ticket/route.ts
```diff
@@ -0,0 +1,379 @@
...
+      process.env.GITHUB_FEEDBACK_TOKEN ? 'Yes' : 'No'
+    );
+
+    const octokit = new Octokit({ auth: process.env.GITHUB_FEEDBACK_TOKEN });
+
+    // Check token validity early - fail fast if token is invalid
+    try {
+      const tokenInfo = await octokit.request('GET /user');
+      // eslint-disable-next-line no-console
+      console.log('Token authenticated as:', tokenInfo.data.login);
```
bilalesi: can we use logger in src/utils/logger.ts so the logs does not appear in prod

tolokoban (approved):
LGTM

PR #1096, opened 2025-11-13
rateLimit was a constantly-changing object that forced multiple rerenders

tolokoban:
We passed an attribute called `rateLimit` to the AI message component.
But we were only using it's attribute `rateLimit.remaining`.

And even when the `remaining` stayed the same, the object `rateLimit` was different.
So the whole list of messages where rendered at every new message sent by the user.
Meaning a lot of Markdown parsing that has not changed.

### DIFF
--- a/src/hooks/debug.ts
+++ b/src/hooks/debug.ts
@@ -1,4 +1,5 @@
 import React from 'react';
+import { logDebug } from '@/utils/logger';
 
 /**
  * Sometime a component is rendering more than it should.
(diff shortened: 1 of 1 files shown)


src/hooks/debug.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Sometime a component is rendering more than it should.
+ * This hook will log a message as soon as a property changes.
+ * This help track the culprit.
+ */
+export function useDebugPropHasChanged(label: string, prop: unknown) {
+  React.useEffect(() => {
+    // eslint-disable-next-line no-console
+    console.log('[Changed!] ', label, ':', prop);
```
bilalesi: can we use logger in src/utils/logger.ts so the logs does not appear in prod

tolokoban: I can change this, of course. But this hook should just be used in dev while tracking an issue with React rendering.
But, i will use the logger, in case someone forget to remove it before going in production.

tolokoban: I have pushed a fix.

PR #1102, opened 2025-11-13
Optimizing the messages of the AI chat

tolokoban:
1. I removed the "free credits" component from the messages and put it next to the "Clear Chat" button.
2. I memoized the `<MessageItem />` component.

In the next video, we can see that the renderings are now only going on the new message.
Message item has a border of random color. If the color does not change, that means that the component has not been rendered. Moreover, the profiler confirms that.

https://github.com/user-attachments/assets/a1db621e-6390-4926-9ad9-86be542b6647

### DIFF
--- a/src/components/ai-assistant/message-item/index.ts
+++ b/src/components/ai-assistant/message-item/index.ts
@@ -1,1 +1,1 @@
-export { default } from './message-item';
+export * from './message-item';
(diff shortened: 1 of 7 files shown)


tolokoban:
@jankrepl The memoization done now is at a higher level than the markdown. I can add one in the markdown too.
For now, the key includes the status of the chat (streaming, done). That's why we can have a rerender at some point, because it will change the appearance of the tools.

I will add another memoization in another PR.

PR #1108, opened 2025-11-17
Runtime configuration system

pgetta:
### Runtime configuration system

Replaces Next.js build-time NEXT_PUBLIC_* environment variables with a runtime configuration system for "build once, deploy everywhere" deployments.

### Key changes:

* **New config system** (`src/config/`): Zod-validated runtime configuration with automatic server/client separation
* **Runtime injection**: Environment variables injected via `window.__ENV__` script tag instead of build-time substitution
* **API URL fallbacks**: Platform API URLs automatically fallback to `API_ORIGIN` if not explicitly set
* **Type-safe access**:
  * Server: `serverConfig` (all properties)
  * Client: `useConfig()` hook or config object (public properties only)
* **Version endpoint**: New `/app/version` route exposing deployment metadata
* **Removed**: `src/env.ts`, `src/config.ts` - replaced by new config system
* **Updated**: All imports to use new config API

### Benefits:

* Single Docker image deployable to multiple environments (in our case it's staging and prod)
* No rebuild required for configuration changes
* Centralized schema with validation
* Clear public/private property separation

### DIFF
--- a/src/features/small-microcircuit/index.tsx
+++ b/src/features/small-microcircuit/index.tsx
@@ -4,8 +4,10 @@
 import { LoadingOutlined } from '@ant-design/icons';
 import { Suspense, useState } from 'react';
 // import { useRouter } from 'next/navigation';
+
 import { useEntries, useModel } from './_components/hooks';
 import Left from './_components/left';
+import Middle from './_components/middle';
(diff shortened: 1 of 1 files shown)


loris-olivier-obi (dismissed):
LGTM

PR #1114, opened 2025-11-17
added electrical cell recording ui

dkeller9:
This branch allows registration of electrical cell recordings. It currently needs this obi-branch to be deployed locally: https://github.com/openbraininstitute/obi-one/tree/nwb_validation_endpoint/obi_one


bilalesi (commented):
I think most of the components are shared with cell-morphology contribution
it's better to create `src/ui/segments/contribute/shared` and move those components there
I can do it if you see this is tricky to do (we need a ticket though)

PR #1123, opened 2025-11-20
112/webgl context lost

tolokoban:
I have added a recovery mechanism in case of context lost.
But after some fixes, I am not able to make it lose its context anymore.


tolokoban:
Thanks @loris-olivier-obi !

g-bar (approved):
LGTM

tolokoban:
Thanks @g-bar !

PR #1130, opened 2025-11-21
Use feature flag to persist ai panel state

pgetta:
This is to make sure that the server rendered page has the same state for the AI panel compared to the client - so that there is no flashing content on page load. Also that addresses the problem when the panel is closed but not rendered correctly: occupying more space than it should ([Enlarged AI assistant panel - that should be in a closed position](https://github.com/openbraininstitute/prod-ai/issues/3)).

@tolokoban as you initially developed the feature - probably best for you to be aware of this.

Additional changes outside of the respective topic:
* Add readme file for feature flags module
* Clean up unused `unifiedSingleNeuronSimulationFlowFlag`
* Extend feature flags so that they can operate with non-boolean values (strings, numbers).

### DIFF
--- a/src/features/feature-flags/actions.ts
+++ b/src/features/feature-flags/actions.ts
@@ -24,10 +24,7 @@ export async function getAllFlags(): Promise<FeatureFlags> {
   }
 }
 
-export async function setFlag<K extends keyof FeatureFlags>(
-  key: K,
-  value: FeatureFlags[K]
-) {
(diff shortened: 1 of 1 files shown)


tolokoban (dismissed):
LGTM!
Thanks Pavlo.

pgetta:
@tolokoban could you please re-approve, had to apply prettier because one file was not formatted.

tolokoban (approved):

PR #1139, opened 2025-11-24
Disable run sim button during an active simulation request

pgetta:
This is fixes the issue where the user can click multiple times on the run simulation button which will lead to multiple simulations being launched/saved and broken plots.

Relates to [Synaptome simulation> if running a simulation with more than 5 steps - error message - unable to save simulation files](https://github.com/openbraininstitute/prod-singlecell-simulation/issues/216).


loris-olivier-obi (approved):
LGTM

pgetta:
Thank you Loris.

PR #1143, opened 2025-11-25
82/add pass failed capsules

tolokoban:
Added Passed/Failed capsules, and custom sorting of the analysis.

image


tolokoban:
Thanks @loris-olivier-obi

PR #1149, opened 2025-11-25
371/only 5 publications per category

tolokoban:
The pagination was not working for the list of publications.

image


tolokoban:
Thanks again @loris-olivier-obi !

PR #1171, opened 2025-12-01
Add simulation complexity evaluation, limit and a poptip warning

pgetta:
The app fails to save the results of highly complex single neuron simulations due to reaching the maximum asset size limit in [Entitycore limit - 150 MB](https://github.com/openbraininstitute/entitycore/blob/main/app/config.py#L51).

This change provides an approximate evaluation and prevents the submission of simulations by disabling the run button and displaying a warning in the form of a tooltip explaining how to reduce the complexity.

Screenshot 2025-12-01 at 17 52 26

Relates to https://github.com/openbraininstitute/prod-singlecell-simulation/issues/216.


pgetta:
Thank you, Fabien.

PR #1177, opened 2025-12-02
Re-route tool help page

WonderPG:
Fixes https://github.com/openbraininstitute/prod-virtual-lab/issues/287

### DIFF
--- a/src/components/ai-assistant/prompt/tools-selector/tool-card/tool-card.tsx
+++ b/src/components/ai-assistant/prompt/tools-selector/tool-card/tool-card.tsx
@@ -4,7 +4,6 @@
 /* eslint-disable jsx-a11y/click-events-have-key-events */
 import React from 'react';
 import Link from 'next/link';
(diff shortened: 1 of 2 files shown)


src/components/ai-assistant/prompt/tools-selector/tool-card/tool-card.tsx
```diff
@@ -19,6 +20,7 @@ interface ToolCardProps {
 }
 
 export default function ToolCard({ className, tool }: ToolCardProps) {
+  const currentPath = usePathname();
```
bilalesi: you can use `useWorkspace()` which will return the `virtual-lab-id` and `project-id` as those are params too, so extraction should be simple and more clear than usePathname with splits

WonderPG: Done in cf646bedb

bilalesi (approved):

PR #1184, opened 2025-12-04
82/analysis pass fail summary

tolokoban:
* Expandable explanation for the results.
* Integrate _Input Resistance_ as a special case of validations.
* Fix the order to match what is said in the explanation.

image


tolokoban:
Thanks Loris!

PR #1199, opened 2025-12-09
333/filter by registration date

tolokoban:
Added filter **Registration Date** for the following artifacts:

* emodel
* memodel
* single_neuron_synaptome
* me_model_circuit
* single_neuron_circuit
* me_model_with_synapses
* single_neuron_simulation
* single_neuron_synaptome_simulation
* me_model_circuit_simulation
* single_neuron_circuit_simulation
* ion_channel_model

image

### DIFF
--- a/src/entity-configuration/definitions/view-defs/model/emodel.ts
+++ b/src/entity-configuration/definitions/view-defs/model/emodel.ts
@@ -17,6 +17,7 @@ export const ViewDefForEmodel: ViewDefinitionConfig = {
     EntityCoreFields.EModelScore,
     EntityCoreFields.Contributions,
     EntityCoreFields.CreationDate,
+    EntityCoreFields.RegistrationDate,
   ],
   curated: true,
   summaryViewFields: [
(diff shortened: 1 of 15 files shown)


loris-olivier-obi (approved):
LGTM

src/entity-configuration/definitions/view-defs/model/emodel.ts
```diff
@@ -17,6 +17,7 @@ export const ViewDefForEmodel: ViewDefinitionConfig = {
     EntityCoreFields.EModelScore,
     EntityCoreFields.Contributions,
     EntityCoreFields.CreationDate,
+    EntityCoreFields.RegistrationDate,
```
bilalesi: fyi: registration_date is the same as creation_date, the only diff is that some artifact are using creation date, and other using registration date (from last year discussion), so here i don't think we should keep both, better to confirm with scientists, ayima

bilalesi: iirc: last year we had experiment data was using registration date, for model, simulation we used creation date but both was pointing to creation_date in `nexus`, this may ring a bell for them

PR #1205, opened 2025-12-09
redirect to synaptome-beta from ai-assistant for circuit with scale "…

bilalesi:
…single"

### DIFF
--- a/src/app/app/entity/[id]/page.tsx
+++ b/src/app/app/entity/[id]/page.tsx
@@ -1,15 +1,43 @@
 import { redirect, notFound } from 'next/navigation';
+
+import { ExtendedEntitiesTypeDict } from '@/api/entitycore/types/extended-entity-type';
+import { CircuitScaleDictionary } from '@/api/entitycore/types/entities/circuit';
+import { getCircuit } from '@/api/entitycore/queries/model/circuit';
 import { getEntity } from '@/api/entitycore/queries/general/entity';
 import { resolveExploreDetailsPageUrl } from '@/utils/url-builder';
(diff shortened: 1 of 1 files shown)


bilalesi:
@g-bar can you have a look at this please, as this is a special case

src/app/app/entity/[id]/page.tsx
```diff
@@ -25,11 +34,20 @@ export default async function EntityDetail({ params }: { params: Promise<{ id: s
       redirectCtx.virtualLabId = workspace.recentWorkspace.virtual_lab_id;
       redirectCtx.projectId = workspace.recentWorkspace.project_id;
     }
+    if (entity.type === EntityTypeDict.Circuit) {
+      const circuit = await getCircuit({ id });
```
Comment (g-bar) [inline src/app/app/entity/[id]/page.tsx]: Do we need to download the circuit here? The entity above should already be the circuit, we might have to type coerce it tho.

src/app/app/entity/[id]/page.tsx
```diff
@@ -25,11 +34,20 @@ export default async function EntityDetail({ params }: { params: Promise<{ id: s
...
-      resolveExploreDetailsPageUrl({ ctx: redirectCtx, entityId: id, dataType: entity.type })
-    );
-    return;
+    const redirectUrl = resolveExploreDetailsPageUrl({
+      ctx: redirectCtx,
+      entityId: id,
+      dataType: entityType,
+    });
+
+    redirect(`${redirectUrl}/overview`);
```
Comment (g-bar) [inline src/app/app/entity/[id]/page.tsx]: Maybe this logic can be put into a function since seems to be the same as below.

g-bar (approved):
👍

PR #1208, opened 2025-12-10
376/ feat: add synthesized morphology to model data

bilalesi:


bilalesi:
this was already approved, I just rebased it, @tolokoban

PR #1211, opened 2025-12-10
Removed resources from footer

loris-olivier-obi:


next-env.d.ts
```diff
@@ -1,6 +1,5 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-/// <reference path="./.next/types/routes.d.ts" />
```
pgetta: Why this is changed?

loris-olivier-obi: I have no idea. I didn't touch or modify this file.

PR #1213, opened 2025-12-11
357/atlas viewer new settings

tolokoban:
Set James Isbister preferred settings for the Atlas Viewer.

image


tolokoban:
Thanks @loris-olivier-obi !

PR #1219, opened 2025-12-11
Bug/altas viewer settings

tolokoban:
The old settings were back again.
The bug occurred when we removed the settings panel.


loris-olivier-obi (approved):
LGTM

PR #1223, opened 2025-12-12
Hotfix: migrate on-the-fly circuit simulation configs

pgetta:
This is a temporary fix to address an issue with circuit simulations where an updated OBI-ONE JSON schema failed to be validated agains older configs rendering the circuit simulation pages unusable.


loris-olivier-obi (approved):
LGTM

PR #1226, opened 2025-12-16
Merge from main: migrate on-the-fly circuit simulation configs (#1223)

pgetta:


pgetta:
Sorry I didn't check the linter right away. Should be ok now.

PR #1242, opened 2025-12-19
Update suggestions

WonderPG:
This PR is to reflect a change in the way we do suggestions for the chat. Related to https://github.com/openbraininstitute/neuroagent/issues/648

I find the logic of the chat suggestions very complex when it should be a simple query being sent to our backend when we switch pages. I tried to keep the current logic but modify it a bit. I also got rid of the hardcoded questions. Please feel free to take over and simplify even more if I missed something.


pgetta:
@WonderPG next time maybe you can ask for the review from `openbraininstitute/webapp` - this way more people will get notified and potentially the PR will be reviewed faster.

PR #1098, opened 2025-11-13
upgrade nextjs v16

bilalesi:
still few changes to fix the styling of some ANTD components as also upgraded to latest version v5.28
preparing for ANTDv6 

fyi: @openbraininstitute/webapp

PR #1245, opened 2025-12-22
Upgrade nextjs to v16 (with Biome linter)

pgetta:
This is based on https://github.com/openbraininstitute/core-web-app/pull/1098 with Biome linter/formatter replacing ESLint.


.github/workflows/run-checks.yml
```diff
@@ -94,31 +34,22 @@ jobs:
...
 
-      # Restore cached dependencies
-      - name: Restore pnpm cache
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/.pnpm-store
-            node_modules
-            */*/node_modules
-          key: ${{ runner.os }}-pnpm-${{ hashFiles('**/pnpm-lock.yaml') }}
```
bilalesi: why remove the cache layer in the CI ?

src/app/app/providers.tsx
```diff
@@ -35,9 +30,6 @@ export function Providers({ children, session, flags }: ProvidersProps) {
             <NuqsAdapter>
               <QueryProvider>
                 <JotaiProvider>
-                  {process.env.NEXT_PUBLIC_JOTAI_DEVTOOLS_ENABLED && (
-                    <DevTools {...{ isInitialOpen: false, position: 'bottom-right' }} />
```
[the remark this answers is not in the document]
pgetta: I've added it back, but there is one caveat: switching to babel (to use `jotai/babel/plugin-react-refresh`, `jotai/babel/plugin-debug-label` plugins) makes everything slower.

Not having these plugins degrades DevTools experience: e.g. one has to provide debug labels manually to have meaningful atom names.

If you are fine with that - we can keep it that way or if you use devtools extensively - we can switch to Babel for now.

.github/workflows/run-checks.yml
pgetta: Caching only a Pnpm store (and not node_modules or both) seems to be a faster alternative  - that is already enabled in the `actions/setup-node@v4`, so no extra configuration is needed.

On the other hand I've realised we were not caching the `.next/cache` - I've just added that, should make subsequent builds even faster.

pgetta: Just tested and the total build time is down from ~4 to ~2 minutes.

pgetta:
Thanks Bilal!

PR #1250, opened 2026-01-05
store credit left in local storage to avoid issue on refresh

jdcourcol:
Free credits are displayed as "0" free credits when we open a new entity in the chat. When asking for another question, that number is back to the proper value coming from the header.
The issue is also happening when refreshing.

Example:
- ask for litterature search -> shows "29 free credits left"
- ask for a morphology in thalamus -> shows "28 free credits left"
- click on the entity to view it -> shows "0 free credits left". Should show "28 free credits left"

### DIFF
--- a/src/services/ai-agent/hooks/chat.ts
+++ b/src/services/ai-agent/hooks/chat.ts
@@ -49,7 +49,7 @@ export function useServiceAiAgentChat(threadId: string) {
   const initialMessages = assistant.initialMessages.useValue();
   const { accessToken } = assistant.useContext();
   const activeTools = useAIActiveTools();
-  const [rateLimitRemaining, setRateLimitRemaining] = React.useState(getStoredRateLimit);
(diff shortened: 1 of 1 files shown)


src/services/ai-agent/hooks/chat.ts
```diff
@@ -17,12 +17,39 @@ export interface AiAgentRateLimit {
...
+  }
+}
+
 export function useServiceAiAgentChat(threadId: string) {
   const assistant = useAiAssistant();
   const initialMessages = assistant.initialMessages.useValue();
   const { accessToken } = assistant.useContext();
   const activeTools = useAIActiveTools();
-  const [rateLimitRemaining, setRateLimitRemaining] = React.useState(0);
+  const [rateLimitRemaining, setRateLimitRemaining] = React.useState(getStoredRateLimit);
```
tolokoban: There is a typo here.
I think you missed the parenthesis:

```ts
const [rateLimitRemaining, setRateLimitRemaining] = React.useState(getStoredRateLimit());
```

PR #1254, opened 2026-01-06
fix style of temperature input

bilalesi:


bilalesi:
@pgetta can we push this tiny style discrepancy between temp/ljp inputs to stg please

pgetta (approved):

PR #1260, opened 2026-01-08
Added role-specific message for build and simulate

loris-olivier-obi:
Add role-specific when the user doesn't have credits anymore.

TIcket: https://github.com/openbraininstitute/prod-virtual-lab/issues/278#issuecomment-3722801734


tolokoban (approved):
Sounds good!

PR #1264, opened 2026-01-12
Implement GUI from JSONSCHEMA Spec

g-bar:
Related to: https://github.com/orgs/openbraininstitute/projects/10/views/21?pane=issue&itemId=145374385&issue=openbraininstitute%7Cprod-platform-architecture%7C175

To be merged together with: https://github.com/openbraininstitute/obi-one/pull/516

### DIFF
--- a/src/components/icons/activity-execution/cancelled.tsx
+++ b/src/components/icons/activity-execution/cancelled.tsx
@@ -1,4 +1,4 @@
-import { CSSProperties } from 'react';
+import type { CSSProperties } from 'react';
 
 type LinkIconProps = {
   className?: string;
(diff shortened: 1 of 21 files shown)


src/features/scan-config/_components/components.tsx
```diff
@@ -0,0 +1,280 @@
+import { Input } from 'antd';
```
tolokoban: Maybe the question is not related to your current code,
but do you know why there is an underscore before `components` in the path?

src/features/scan-config/_components/reference.tsx
```diff
@@ -0,0 +1,56 @@
...
+    !schema ||
+    !schema.default_block_reference_labels ||
+    !schema.default_block_reference_labels[referenceSchema.reference_type]
+  ) {
+    return null;
+  }
+
+  const configOptions = referenceTypeDict[referenceSchema.reference_type];
+
+  const options: { label: string; value: string | null }[] = Object.keys(
```
tolokoban: Do you think you could memoize this? Or is it not worth it?
I don't know if we already have React compilation on.

PR #1266, opened 2026-01-13
Update staging Keycloak URL

pgetta:
According to the latest infra changes the Keycloak should be accessed with `https://staging.cell-a.openbraininstitute.org/auth`.


pgetta:
Thank you!

PR #1268, opened 2026-01-13
Show error status in the detailed view if there is no sim config asset

pgetta:
This PR:
* adds a simple circuit simulation (config) validation by checking the presence of sonata simulation config
* disables launching simulations which do not pass the validation above
* marks the simulation status as `error` with additional tooltip (and explanation) in the detailed view

Relates to [If it was not possible to generate simulations, it should not be possible to click "Launch simulation" from Data](https://github.com/openbraininstitute/prod-singlecell-simulation/issues/229).

Screenshot 2026-01-28 at 15 27 15

### DIFF
--- a/src/features/scan-config/components/simulations.tsx
+++ b/src/features/scan-config/components/simulations.tsx
@@ -208,7 +208,9 @@ export default function SimulationsTab({
         ctx: { virtualLabId, projectId },
         simulationIds: simIds,
         onInit: () => {
-          simIds.forEach((simId) => void setSimStatus(simId, EntitycoreExecutionStatus.PENDING));
+          simIds.forEach((simId) => {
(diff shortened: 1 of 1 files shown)


src/features/scan-config/components/simulations.tsx
```diff
@@ -208,7 +208,7 @@ export default function SimulationsTab({
         ctx: { virtualLabId, projectId },
         simulationIds: simIds,
         onInit: () => {
-          simIds.forEach((simId) => setSimStatus(simId, EntitycoreExecutionStatus.PENDING));
+          simIds.forEach((simId) => void setSimStatus(simId, EntitycoreExecutionStatus.PENDING));
```
bilalesi: it can be just wrapped in curly brackets  instead of void

bilalesi (dismissed):

PR #1272, opened 2026-01-14
feat-9: add multi species human

bilalesi:
**RELATED TO**: [TICKET](https://github.com/openbraininstitute/prod-explore-functionality/issues/9)

this PR adds support for multiple species (mainly human) in the brain region hierarchy, beforethis, everything was hardcoded to work with one species only, now the user can switch between species and the whole app reacts to that brain regions, atlas viewer, data tables, data contribution forms, everything.

the big change is a new hook called `useWorkspaceHierarchyRegistry` that replaces the old atom-based approach (`useBrainRegionHierarchy, brainRegionBasicCellGroupsRegionsHierarchyAtom, selectedBrainRegionAtom` direct usage, etc)

**⚠️ Important** :

this hook takes care of syncing the selection across URL params, localStorage, remote user preferences, and Jotai atoms. you don't need to worry about any of that manually.
```ts
import { useWorkspaceHierarchyRegistry } from '@/features/brain-region-hierarchy/hooks';

const {
  selectedBrainRegion,      // the current brain region object
  workspaceHierarchyId,     // current hierarchy ID (changes per species)
  workspaceSpecies,         // current species info
  changeBrainRegion,        // call this to change brain region
  changeBulkStoreHierarchySpecies,  // call this to switch species
} = useWorkspaceHierarchyRegistry();
```

**How the state priority works**
when the app loads, the hook figures out which `species/brain-region` to show in this order:

- URL params (so links are shareable)
- Remote user preference from the API
- localStorage (local persistence)
- Default config fallback

> when user switches species, the hook remembers the previous brain region per hierarchy, so if they go back it restores where they were.

**Other changes**
- cell composition is hidden for human species (not supported yet)
- camera behavior adjusted for human brain atlas
- `BrainRegionDropdown` component now uses the workspace registry internally
- all contribution forms migrated to use `useWorkspaceHierarchyRegistry`
- Species selector UI added to the explore/atlas section
- Brain region selection persisted to backend via fire-and-forget API call

**What to test**
- switch between species and verify brain regions update everywhere
- check that data tables filter correctly per species hierarchy
- verify contribution forms pick up the right default brain region
- test that URL sharing works (copy URL with species params, open in new tab)
- check that going back to a previously visited species restores the last selected brain region


g-bar (approved):
LGTM

bilalesi:
thanks @g-bar

PR #1275, opened 2026-01-16
Added animation on reload and login

loris-olivier-obi:


pgetta:
@loris-olivier-obi I've missed this one, is it still relevant?

PR #1289, opened 2026-01-21
Remove credits remaining in chat

WonderPG:
Related to https://github.com/openbraininstitute/prod-explore-functionality/issues/314#event-22096442490

It might be a bit tough to test since staging provides 30 free queries before paying credits. DM me if you need help to test it.


src/components/ai-assistant/state.ts
```diff
@@ -1,14 +1,17 @@
...
+import type { AiAgentRateLimitEndpoint } from '@/services/ai-agent/hooks/chat';
+import React from 'react';
 
 const atomToolsInvertedSelection = atomWithStorage<string[]>(
   'AIAssistant/tools-inverted-selection',
   []
 );
 
+export const atomRateLimit = atom<AiAgentRateLimitEndpoint | null>(null);
+
```
bilalesi: what this is for ?; 
i noticed there is few queries using TQuery

src/services/ai-agent/hooks/chat.ts
```diff
@@ -65,19 +35,17 @@ export function useServiceAiAgentChat(threadId: string) {
...
+      const newRateLimit: AiAgentRateLimitEndpoint = {
         limit: parseInt(resp.headers.get('x-ratelimit-limit') ?? '-1', 10),
         remaining: parseInt(resp.headers.get('x-ratelimit-remaining') ?? '-1', 10),
-        reset: parseInt(resp.headers.get('x-ratelimit-reset') ?? '-1', 10),
+        reset_in: parseInt(resp.headers.get('x-ratelimit-reset') ?? '-1', 10),
       };
-      setRateLimitRemaining(newRateLimit.remaining);
-      setStoredRateLimit(newRateLimit.remaining);
+      setRateLimit(newRateLimit);
       return resp;
```
bilalesi: i don't know why hijacking fetch here, useChat (in v4), already have callbacks, we can use `onResponse`, wdyt ?
- onResponse is designed for that
- overriding fetch mixes network layer logic with ui concerns

also this will help on migration to v6 (if it's in your plan), as fetch does not exists, and `onResponse` will be changed to `onFinish`

src/services/ai-agent/hooks/chat.ts
```diff
@@ -2,52 +2,22 @@
...
 import { type CreateMessage, type Message, useChat } from '@ai-sdk/react';
 import type { ChatRequestOptions } from '@ai-sdk/ui-utils';
-import React from 'react';
-import { useAIActiveTools } from '@/components/ai-assistant/state';
+import { useSetAtom } from 'jotai';
+import { useAIActiveTools, atomRateLimit } from '@/components/ai-assistant/state';
 import { logError } from '@/util/logger';
 import { serviceAiAgentThreadSuggestTitle, serviceAiAgentUrl } from '../api';
 import { useAiAssistant } from '../assistant';
+import type { AiAgentRateLimitEndpoint } from './rate-limit';
```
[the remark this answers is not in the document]
tolokoban: Not everybody uses this convention.

It comes because, in Javascript/Typescript, there is no way of creating packages with modules that are private and only accessible from within this package.

The convention then says that if you have a `index.ts` in a folder, then it is forbidden to import anything else from outside this folder.
You can import the folder (and then the `index.ts` file), but no file that is inside.

This way, you can have files that have code that is only to be used internally and that you can break if you need, without impacting the rest of the code.

src/components/ai-assistant/free-credits-notification/free-credits-notification.tsx
```diff
@@ -0,0 +1,64 @@
...
+    // Check if reset is today or tomorrow
+    const now = new Date();
+    const isToday = resetDate.toDateString() === now.toDateString();
+
+    if (isToday) {
+      return `at ${timeString}`;
+    }
+
+    return `tomorrow at ${timeString}`;
+  }, [resetIn]);
```
WonderPG: I would like to double check with you guys that this piece of code is correct. The goal is to display the time at which free credits will become available again, but in the local timezone of the user. I.e. if someone from the US and from Japan send their last message at the same time, it should show a different replenishment time.

`toLocaleTimeString` is what we want to use here right ?

src/components/ai-assistant/free-credits-notification/free-credits-notification.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export default function FreeCreditsNotification({
+  onDismiss,
+  resetIn,
+}: FreeCreditsNotificationProps) {
+  // Freeze the reset time when component mounts
+  const resetTimeString = useMemo(() => {
+    if (resetIn === null || resetIn <= 0) {
+      return "later";
```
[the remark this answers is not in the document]
WonderPG: It should not happen though

PR #1292, opened 2026-01-22
Fix client config context

pgetta:
The main issue was that the `config` is a Proxy object which can not be passed through the boundary between a Server and a Client component (only plain objects are supported). This PR fixes that.

Relates to https://github.com/openbraininstitute/core-web-app/pull/1283.


jdcourcol:
Can we fast track this PR to reach prod asap ?

PR #1320, opened 2026-01-30
Remove unused status for single cell simulations

pgetta:
It's hardcoded in the simulation entity creation and just displayed in the list view - so doesn't have a real use.

With the recent entitycore update the status has been also removed from the schema.


pgetta:
Thank you all.

This time I actually had to check the documentation if it needs to be updated 😄

PR #1321, opened 2026-01-30
Block union

g-bar:
Simple implementation of 'block_union' UI element, that will be needed for circuit extraction but also other UIs including "reset" functionality.
See vid.

FYI: @bilalesi 

https://github.com/user-attachments/assets/5641a232-1a9b-4f57-92fb-cc40e79edf71


bilalesi:
please, can you keep only the other functionality so it will not make conflicts into the circuit extraction, 
I already have so many conflicts this morning due necessary refactoring (simulation vs extraction)

bilalesi:
> Yes, I know but this follows a simpler approach by not using any extra state like `selectedCategory` or `selectedEntry` and just reading the type directly from the stateAtom, thought might be useful.

i understand, my implementation also not using those

PR #1323, opened 2026-01-30
Migrate microcircuit launch to OBI-ONE API

pgetta:
Instead of using task-launcher API directly now the calls to run circuit simulations (on the microcircuit scale for now - the rest later) will be executed via OBI-ONE API.

Relates to [Capability (P1): Enable large simulation product](https://github.com/openbraininstitute/prod-circuit-simulation/issues/91).


src/api/launch-system/index.ts
```diff
@@ -1,3 +1,5 @@
+// TODO Module to be deprecated/removed. The corresponding functionality to be migrated into OBI-ONE API.
+
```
bilalesi: why not removing the file ?

bilalesi (commented):
I think this should wait until Elef PR merged in obi-one ?

src/api/launch-system/index.ts
pgetta: Good suggestion, it's done.

bilalesi (approved):

PR #1329, opened 2026-02-02
Fix simulation status and extend whitelisted types for expandable table

pgetta:
* Correctly show "failed" status in the extended table view if there is a simulation config is missing (generation failure)
* Fetch simulations in chunks by 30 to avoid triggering AWS load balancer firewall
* Show extended status table for the rest of circuit simulations

Fixes issue related to:
* [If it was not possible to generate simulations, it should not be possible to click "Launch simulation" from Data](https://github.com/openbraininstitute/prod-singlecell-simulation/issues/229)
* [Simulation campaign should have an expandable status pills to inspect individual simulation status](https://github.com/openbraininstitute/prod-virtual-lab/issues/281)


tolokoban (approved):
LGTM

PR #1334, opened 2026-02-03
Fix sim result selection, simplify the logic

pgetta:
The logic to auto-select simulation results has got issues when the react compiler was enabled. This PR simplifies it as well as makes it follow the Rules of React to be react compiler compatible.

Also auto-selected format has been fixed to NWB.

Relates to [[Single neuron beta] DEV (and now in Staging) - not possible to see 'interactive details' of the simulation](https://github.com/openbraininstitute/prod-singlecell-simulation/issues/251).

### DIFF
--- a/src/features/scan-config/components/file-viewer/index.tsx
+++ b/src/features/scan-config/components/file-viewer/index.tsx
@@ -19,16 +19,8 @@ type FileViewerProps = {
 
 export function FileViewer({ file, context, loading = false, className = '' }: FileViewerProps) {
   const [displayFile, setDisplayFile] = useState<File | undefined>(file);
-  const [isFilePreloading, setIsFilePreloading] = useState(false);
 
-  useEffect(() => {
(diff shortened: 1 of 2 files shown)


pgetta:
> I’d be curious to learn:
> 
> 1. What was the issue with the React compiler.
> 2. How do we enable the React compiler locally so that results are deterministic with what we have in the cloud.

Thank you, Gil. These are great questions.

I've noticed that the issue appeared when we enabled the react compiler. Initially I thought it only works in production - but have just read it's part of the build process - so dev is affected as well. Maybe it has something to do also with React strict mode - as the issue was only seen in production builds, but I'm not sure, didn't look deep enough.

At the end I saw that the logic was not well written so I've simplified it and that resolved the issue.

As for the production build - one can run it locally with `make run` - it's almost the same to what we have in production.

g-bar:
Ah I see, thanks for looking into it!

PR #1341, opened 2026-02-04
add user identifier to feedback

jdcourcol:
The user identifiers are missing in the ticket (since the ticket is created by a service account)


loris-olivier-obi (approved):
LGTM

PR #1355, opened 2026-02-06
fix color override for icons in the left menu build/simulate workflosw

bilalesi:
this is not the final solution, as i think the biome configuration is changing the css files order in the imports and lead to class override hell, I will define an import order in another PR for biome so we keep always css files at the end


g-bar:
Can we undo we the formatting changes?  The file was already formatted with biome 🤔 .

PR #1356, opened 2026-02-06
specify imports order in biome and place style files at the end

bilalesi:
biome goes top→down and stops at the first group that matches the order in biome.json, 
and with current config, sometimes biome cann't distinguish between package/path-aliases imports

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -80,7 +80,44 @@
     "enabled": true,
     "actions": {
       "source": {
-        "organizeImports": "on"
+        "organizeImports": {
+          "level": "on",
+          "options": {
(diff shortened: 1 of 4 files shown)


tolokoban:
Biome is faster than eslint, but it's less mature and therefore more buggy.

Just try this little code:

```ts
export class MyClass {
  private readonly myVar: string;

  constructor() {
    this.myVar = "Hello";
  }

  print() {
    const { myVar } = this;
    console.log(myVar);
  }
}
```

Biome complains that `myVar` _is defined but never used_.
And that's obviously wrong.

PR #1414, opened 2026-02-17
Reconnect develop and main branches

pgetta:
That should make it possible to use normal git merge process (and to have clear git history).


pgetta:
Thank you Fabien,

> That's strange, but I don't see any change in Github.

Yeah, that's expected. At some point devolop and main have diverged too much to have a normal merges.
This PR (set of commits) should fix that, but it's true that the diff itself is null for this change.

PR #1416, opened 2026-02-17
Add EM cell mesh skeletonization workflow

pgetta:
This PR adds support to run mesh skeletonization via the UI as well as enables corresponding workflows.

Relates to [Show skeletonization form in GUI](https://github.com/openbraininstitute/Ultraliser/issues/197).

Screenshot 2026-02-20 at 13 22 34

This change requires backend functionality introduced  in [EM Cell Mesh skeletonization batch API with streaming
](https://github.com/openbraininstitute/Bluenaas/pull/115).


tolokoban (dismissed):
I still find that protecting the code from potentially unexpected external types is important, but since the rest looks good to me, I approved.

g-bar (dismissed):

james-isbister:
I think I originally suggested "Process data" but actually @MWolfR  has since said it should be in Build

pgetta:
> Process data is not showed in the first scrollable part, may be order this category to be right after extract ?

I've added ordering so that disabled categories are shown last, preserving their original order inside the group (enabled/disabled).

bilalesi (dismissed):

src/auth/index.ts
```diff
@@ -92,12 +101,10 @@ export async function refreshAccessToken(token: TokenSet) {
...
-    void (async () => {
-      await upsertRefreshTokenInAuthManager({
-        accessToken: refreshedTokens.access_token,
-        refreshToken: refreshedTokens.refresh_token ?? token.refreshToken,
-      });
-    })();
+    await upsertRefreshTokenInAuthManager({
+      accessToken: refreshedTokens.access_token,
+      refreshToken: refreshedTokens.refresh_token ?? token.refreshToken,
+    });
```
bilalesi: is there a reason why you remove the unblocking upsert, this  basically will add some latency for each refresh operation with KC

bilalesi: hhh, i think i comment in the same time of your revert, ignore it

pgetta: For some reason the consent flow doesn't work in preview, even thought it's using the same keycloak client.
I know that AWS lambda based SSR in Amplify apps have some more limits than a node.js server, was checking if somehow the request was killed right after the response was sent.

But it didn't fix the issue, so there is something else.

pgetta: Sorry for all this noise

PR #1425, opened 2026-02-18
CHAT New chat UI

BoBer78:
This branches out of https://github.com/openbraininstitute/core-web-app/pull/1372
This is to separate the bugfixes and the new UI elements.


ayimaok:
looks very good! @BoBer78

loris-olivier-obi (dismissed):
LGTM

src/services/ai-agent/hooks/chat.ts
```diff
@@ -218,15 +216,69 @@ export function useServiceAiAgentChat(threadId: string) {
...
+    [chat, accessToken, threadId, queryClient, virtualLabId, projectId]
+  );
+
+  const stop = useCallback(async () => {
+    chat.stop();
+    queryClient.invalidateQueries({
+      queryKey: keyBuilderAI.messages(threadId, virtualLabId, projectId),
+    });
+    const oldMessages = chat.messages;
+    const messages = await fetchMessagesFromDB(
```
tolokoban, line 234: What does the `fetchMessagesFromDB()` do?
You use `chat.stop()` above, but is there a way to wait for the stop to actually happen? Or do we risk to receive messages arriving after the `fetchMessagesFromDB()` call?

src/services/ai-agent/hooks/chat.ts
```diff
@@ -218,15 +216,69 @@ export function useServiceAiAgentChat(threadId: string) {
...
+          id: `temp-id-${crypto.randomUUID()}`,
+          role: 'assistant',
+          content: '',
+          parts: [],
+        },
+      ]);
+    }
+  }, [chat, queryClient, accessToken, virtualLabId, projectId, threadId]);
+
+  useEffect(() => {
```
tolokoban, line 261: Is there a reason why `assistant` is not part of the dependencies list?

src/services/ai-agent/hooks/chat.ts
```diff
@@ -218,15 +216,69 @@ export function useServiceAiAgentChat(threadId: string) {
...
+    assistant.chat.sync({
+      status: chat.status,
+      error: chat.error,
+      append,
+      stop,
+    });
+  }, [chat.status, chat.error, append, stop]);
+
+  useEffect(() => {
```
tolokoban, line 270: Is there a reason why `assistant` is not part of the dependencies list?

src/components/ai-assistant/panel-content/history/search/search.tsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+export default function Search({ onSelectThread }: SearchProps) {
+  const [searchQuery, setSearchQuery] = React.useState('');
+  const { results, isLoading, hasQuery } = useThreadSearch(searchQuery);
+  const searchRef = React.useRef<HTMLDivElement>(null);
+
+  const highlightText = (text: string, query: string) => {
+    if (!query.trim()) return text;
+    const parts = text.split(new RegExp(`(${query})`, 'gi'));
```
tolokoban: There is a potential security issue here.
`query` comes from the user input and you snap it directly into a `RegExp` constructor.
If the user enters "(", for instance, this line will throw an error that is not catched.

src/components/ai-assistant/suggested-questions/suggested-questions.tsx
```diff
@@ -11,62 +11,46 @@ interface SuggestedQuestionsProps {
...
   suggestions: string[];
   clearSuggestions(): void;
   isLoading: boolean;
 }
 
+const SKELETON_COUNT = 3;
+
 export default function SuggestedQuestions({
   className,
   threadId,
```
tolokoban: 'threadId' is declared but its value is never read.

src/components/ai-assistant/message-item/storage-plots/renderers/plot-renderer/tool-plot-generator.tsx
```diff
@@ -165,44 +503,68 @@ function CustomPlot({
...
+    <div
+      className={compact ? 'px-2 py-0.5 text-center font-bold' : 'px-4 py-2 text-center font-bold'}
+      title={title}
+      style={{
+        fontSize,
+        fontFamily: titleFont.family || 'Arial, sans-serif',
+        fontWeight: titleFont.weight || 'bold',
+        color: titleFont.color || '#333',
+        lineHeight: 1.2,
+        paddingLeft: paddingRight,
```
tolokoban: I think this is intentional, but I tell you in case it is not.
I just saw you affected right to left so I was suspicious  :-)

src/components/ai-assistant/message-item/fullscreen-dialog/fullscreen-dialog.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  const handleBackdropClick = (e: React.MouseEvent<HTMLDivElement>) => {
+    if (e.target === e.currentTarget) {
+      handleClose();
+    }
+  };
+
+  return (
+    <dialog ref={dialogRef} className={styles.dialog}>
+      <div className={styles.backdrop} onClick={handleBackdropClick}>
```
tolokoban: The linter is complaining for a11y (accessibility) here and in the next line:
[rule](https://biomejs.dev/linter/rules/use-key-with-click-events)

tolokoban (dismissed):

src/services/ai-agent/hooks/chat.ts
BoBer78: So when the user stops, the backend has very often a couple more tokens from the LLM than the frontend. 
The idea is to sync both on stop. 
Issue is that in some rare cases (for example the user stop right away) where the DB has not synced yet. 
This is way there is a de-duplication logic. 

In the future, with the typescript re-write of the backend we might have more control on this one.

src/services/ai-agent/hooks/chat.ts
BoBer78: assistant was re-defined on every render. Fixed to make it clearer in : [8cd5806](https://github.com/openbraininstitute/core-web-app/pull/1425/commits/8cd5806f11b585c38000eb42a95313a517f8f420)

src/services/ai-agent/hooks/chat.ts
BoBer78: Same as above.

src/components/ai-assistant/panel-content/history/search/search.tsx
BoBer78: [930910a](https://github.com/openbraininstitute/core-web-app/pull/1425/commits/930910a3a358820bd55f866dcc852ad2704dd4fb)

Fixed here, it is enough ?

src/components/ai-assistant/suggested-questions/suggested-questions.tsx
BoBer78: Fixed here : [a2a4082](https://github.com/openbraininstitute/core-web-app/pull/1425/commits/a2a4082ca82b94ba64b65b4cfc658662d3c38575)

src/components/ai-assistant/panel-content/history/search/search.tsx
tolokoban: That will prevent the exception, indeed.
Is it on purpose that when you split with `query` (after escape, of course) you get the separator in `parts`? That's what adding the parenthesis is doing, I just wanted to know if it was not a mistake.

See [here](https://www.typescriptlang.org/play/?#code/FAMwrgdgxgLglgewgAhgUwM4wBQEcxoBOAngFzJaFwQDmAlMgN7DKvJRJapoAeMyAXmQAiABJE0yAO6SOAWzQAaZABMwKtAEIAOhACChScQRhkhgIYriAfmEs2HCBgQAbNADoXCGtmH4iZMLK-iR09qyOXJhQ5gAOaCqCyCHE7oaxLuZQaNgA9ADa7gBUANTWAHoAJIwAvth0AD752gC62q25NMoA5O2VAGTddADc4eycrh5ePsLRcQmkQchz8SphbONO-LHmhDAYSeh87hgZcDgQaFLIAEpoNACiPLHYAAbY1SsJNXSvPTRwIYjMa5XKbLg7PYHIRHGAnM4XK63e5PF6vT4YGKrGp-ZDdAFA0YbSKTTzebCQ-ZhGrAYDoLC+ERhIA)

BoBer78: It is on purpose, I want to keep the whole text to then highlight it, wrapping the searched parts with html tags. 
Maybe there is a cleaner way to do this.

tolokoban: No, that's perfect. Thanks.

src/components/ai-assistant/message-item/storage-plots/renderers/plot-renderer/tool-plot-generator.tsx
BoBer78: I renamed the variable here : [7528964](https://github.com/openbraininstitute/core-web-app/pull/1425/commits/7528964ac7e4a55b145d4e9e0e4ef2c025efdad7)

Same functionality, just different naming.

src/components/ai-assistant/message-item/fullscreen-dialog/fullscreen-dialog.tsx
BoBer78: Fixed here : [fdf9df8](https://github.com/openbraininstitute/core-web-app/pull/1425/commits/fdf9df83c7855dee1cc15b9202374162f1e62dba)

tolokoban:
All good!

tolokoban (dismissed):

tolokoban (approved):

PR #1428, opened 2026-02-19
Settings for Atlas viewer only in debug mode

tolokoban:
The settings that allow to tweak the look of the altas meshes and points clouds are only shown when the local storage item `debug:atlas-viewer` exists and is not empty.

image

image


bilalesi (approved):
👍

PR #1434, opened 2026-02-20
130/dendrogram with morphoviewer library

tolokoban:
In this PR, we added the Dendrogram view and the transition between 3D and dendrogram, even for 
the synapses.

The important changes are:

1. Upgrade of the morphoviewer library: `tarball/bbp-morphoviewer-0.22.1.tgz`
2. Using it here: `src/components/neuron-viewer/neuron-viewer-with-actions.tsx`
3. Added two files to deal with recordings and synapses: `src/components/neuron-viewer/hooks/electrodes.ts` and `src/components/neuron-viewer/hooks/synapses.ts`.

https://github.com/user-attachments/assets/f657d82d-23fc-40a5-b2a8-c27adc6db763


bilalesi:
@tolokoban i think you run format before you pushed it
I understand this is getting messier, we may need a PR to only format/lint the whole thing but better to do it in different PR to see any breaking things

PR #1437, opened 2026-02-20
Clean up unused CDN setup

pgetta:
Since staging and production deployments of Core Web app are running in Azure where CDN is located in the front of the whole app - there is no need to have old setup / AWS integration.


bilalesi (approved):
thanks pavlo !

PR #1462, opened 2026-02-27
Circuit viz

g-bar:
Implements: 
https://github.com/openbraininstitute/prod-circuit-simulation/issues/74
Requires: 
https://github.com/openbraininstitute/obi-one/pull/609


tolokoban (dismissed):
Cool!

src/features/scan-config/components/circuit-viz/circuit-viz.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  if (isLoading || error) {
+    return (
+      <div className="w-full h-full flex justify-center items-center">
+        {isLoading && <LoadingOutlined className="text-4xl text-primary-8" />}
+        {!!error && (
+          <details className="text-red-500">
+            <summary>
+              <strong>Couldn't load the visualization</strong>
+              <div>{error.message}</div>
```
bilalesi: i think you meant to put the error message in the details
currently everything will be in the "title"

src/features/scan-config/components/circuit-viz/hooks.ts
```diff
@@ -0,0 +1,120 @@
...
+  React.useEffect(() => {
+    nodes?.forEach((n) => {
+      promises.current[n.morphology_file + n.morphology_name] = queryClient.fetchQuery({
+        queryKey: ['morphology', id, n.morphology_file, n.morphology_name],
+        queryFn: async () => {
+          const nameParam = n.morphology_name
+            ? `?name=${encodeURIComponent(n.morphology_name)}`
+            : '';
+          const url = `${config.OBI_ONE_URL}/circuit/viz/${id}/morphologies/${encodeURIComponent(n.morphology_file)}${nameParam}`;
+
```
bilalesi, line 49: why using "fetch query" with useEffect, this is an anti pattern 
useQuery should do all what you need and get rid of the useEffect

src/features/scan-config/components/circuit-viz/hooks.ts
```diff
@@ -0,0 +1,120 @@
...
+    queryFn: async () => {
+      const res = await authFetch(`${config.OBI_ONE_URL}/circuit/viz/${id}/nodes`, {
+        headers: {
+          'Content-Type': 'application/json',
+          Accept: 'application/json',
+          'virtual-lab-id': virtualLabId,
+          'project-id': projectId,
+        },
+      });
+
```
bilalesi, line 97: can we move this to obi api endpoint: src/api/one

src/features/scan-config/components/model-preview/index.tsx
```diff
@@ -18,13 +18,18 @@ export function ModelPreview({ model }: { model: TSupportedEntitiesForScanConfig
...
       />
     ))
-    .with({ type: EntityTypeDict.Circuit, scale: CircuitScaleDictionary.Single }, () => (
+    .with({ type: EntityTypeDict.Circuit, scale: CircuitScaleDictionary.PairNeuron }, () => (
+      <CircuitPreview circuit={model as ICircuit} />
+    ))
+    .with({ type: EntityTypeDict.Circuit, scale: CircuitScaleDictionary.SmallMicrocircuit }, () => (
+      <CircuitPreview circuit={model as ICircuit} />
+    ))
+    .with({ type: EntityTypeDict.Circuit }, () => (
```
bilalesi: this can be 
```ts
  .with(
    {
      type: EntityTypeDict.Circuit,
      scale: P.union(
        CircuitScaleDictionary.PairNeuron,
        CircuitScaleDictionary.SmallMicrocircuit
      ),
    },
    () => <CircuitPreview circuit={model as ICircuit} />
  )
```

bilalesi: `P` from `ts-pattern` too

src/features/scan-config/use-cases/build/in-out-files.tsx
```diff
@@ -38,6 +39,74 @@ type Props = {
...
+  id,
+  path,
+  contentType,
+}: {
+  id: string;
+  path: string;
+  contentType: AssetContentType;
+}): IAsset {
+  return {
+    id,
```
bilalesi: this is not supposed to be used anymore, it was refactored here
`src/features/task-logs-stream/descriptor.ts` and changed to make___

src/features/scan-config/components/circuit-viz/hooks.ts
g-bar: Not sure how to accomplish this without the effect, we require the promises array, which are later awated internally by the MorphoViewer component. the usequery doesn't expose the promises, just the data.

but we still want to benefit from the cache for subsequent calls.

bilalesi: you can use `useQueries` and map through the nodes, smthg like: 
```ts
const morphologyQueries = useQueries({
  queries:
    nodes?.map((node) => ({
      queryKey: [
        "morphology",
        node.morphology_file,
        node.morphology_name,
      ],

      queryFn: async () => {
        return fetchMorphology(node);
      },

      enabled: !!node.morphology_file,
      staleTime: Infinity,
    })) ?? [],
});
```

it also have combine, if you want to do some data aggregation (loading state, final data, error-reduction)

bilalesi: may be i misunderstood the promise inside the morphoviwer thingy

tolokoban: Yes, I agree.

I didn't touch this part of the code because the project is to replace it something Pavlo has done. In this current implementation, we fetch data from a specific entrypoint, but in a near future, we will use a WebWorker in the frontend and just retrieve H% files from EntityCore.

src/features/scan-config/components/circuit-viz/hooks.ts
tolokoban: Same as above. All this fetching code will be soon replaced with Pavlo's WebWorker.

src/features/scan-config/components/circuit-viz/circuit-viz.tsx
tolokoban: Nice catch!
I fixed it.

src/features/scan-config/components/model-preview/index.tsx
tolokoban: That's really good to know. Thanks!
It's fixed.

src/features/scan-config/use-cases/build/in-out-files.tsx
tolokoban: You are right! Good catch!
This branch is here for a long time now and it has been rebased several times.
I think that's why we have such mistakes.

Thanks for the review!

It's fixed.

bilalesi (approved):

tolokoban:
Thanks @bilalesi !

PR #1463, opened 2026-02-27
Show error notifications in ion channel build

pgetta:


pgetta:
Relates to [Update accounting prices and cost models](https://github.com/openbraininstitute/prod-virtual-lab/issues/314).

PR #1474, opened 2026-03-03
Patch sim config

WonderPG:
Contributes to https://github.com/openbraininstitute/prod-ai/issues/5

### DIFF
--- a/src/features/scan-config/components/ui-columns/left.tsx
+++ b/src/features/scan-config/components/ui-columns/left.tsx
@@ -130,7 +130,7 @@ export default function Left({
         })}
       </div>
 
-      {!readOnly && (!aiConfig || (aiConfig && campaignId)) && (
+      {!readOnly &&
         <GenerateConfigButton
           loading={loading}
           campaignId={campaignId}
@@ -142,7 +142,7 @@ export default function Left({
           setLoading={setLoading}
           activity={activity}
         />
-      )}
+      }
     </div>
   );
 }


src/features/scan-config/template.tsx
```diff
@@ -180,11 +180,6 @@ export function ScanConfigTemplate({
               if (!aiConfig) return;
               resetConfig(schema, aiConfig, setAtomsMap);
               setAiConfig(null);
-              updateRequestId();
```
g-bar: We don't we update the requestId?

src/features/scan-config/components/block.tsx
```diff
@@ -129,38 +122,92 @@ export default function Block({
...
+        // Array of NamedTuples
+        value.forEach((namedTuple) => {
+          if (isPlainObject(namedTuple) && Array.isArray(namedTuple.elements)) {
+            elements.push(...namedTuple.elements);
+          }
+        });
+      } else if (isPlainObject(value) && Array.isArray(value.elements)) {
+        // Single NamedTuple
+        elements.push(...value.elements);
+      }
```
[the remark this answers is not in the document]
WonderPG: If that's the case it's a good point. Removing the list would simplify significantly the code but I suggest we let @james-isbister give us his opinion on this one.

ayimaok:
@WonderPG, this is much better than what was before. 

Question, would you know, if I'd launch a simulation, however, am still in the config, usually, one can click "New simulation campaign" and change the parameters to re-launch the simulation. 
However, the chat does not see that state, as of now. Result = if the user asks to change some fields, the chat replies that it did, but in reality nothing happens.

ayimaok:
Just to update this, we've had a discussion with @WonderPG and agreed to continue the discussion this week. 
The question is what to do with state sharing, i.e. what should be the intention eg, once the simulation has been launched, or if the user changes from Workflows to Data.

james-isbister:
IMO, once the simulation has been generated and a subset of the simulations have possibly been run (not essential), it would be cool if the user can prompt again asking for a new campaign or a change to the existing campaign. 

This would take the UI back to the configurations tab and create a new suggestion. The user might even prompt directly from the Configuration tab when its in read only mode

I think this is useful when the user realises they want to change something in the config, or they see a result and want to iterate

src/features/scan-config/components/ui-columns/left.tsx
```diff
@@ -70,11 +69,17 @@ export default function Left({
...
 }) {
   const errors = useValidateSchema({ initialConfig, config, schema });
   const { aiConfig } = useAIConfig();
 
+  // Auto-apply AI-generated configuration changes when available
+  useEffect(() => {
+    if (aiConfig && !campaignId) {
+      handleAcceptAIChanges();
+    }
+  }, [aiConfig, campaignId, handleAcceptAIChanges]);
```
bilalesi: what's this accept changes fn?
if it's some sort of atom update then better do it when you get the data from the AI ? and avoid using useEffect

src/features/scan-config/template.tsx
```diff
@@ -80,7 +80,7 @@ export function ScanConfigTemplate({
 
   const aiEnabled = 'scale' in entity && entity.scale !== 'single';
 
-  const updateRequestId = useAgentState(aiEnabled ? 'smc_simulation_config' : '', config);
+  useAgentState(aiEnabled ? 'smc_simulation_config' : '', config);
```
bilalesi: can we move this `smc_simulation_config` as const elsewhere as in `helpers.ts` 
probably will have more then config 
and may be it can be a map between the activity (simulation, extraction, ...) and the name of the config `smc_simulation_config`

bilalesi (approved):

PR #1484, opened 2026-03-04
update(api-client): api client to use zod schema for validation on se…

bilalesi:
- add `zod` schema validation for api-client
- the validation is optional for the moment
- the validation is on 
    - query parameters 
    - request body
    - response
- i updated the cell morphology to use the schema validation on response for the time being for test
- i would suggest to keep it optional, and update endpoint gradually to not break more the linting, and functionalities 
- this is require the `zod` schema to be in sync with pydantic schemas of entitycore (or whatever service use this client)

cc: @tolokoban

### DIFF
--- a/src/api/api-client.ts
+++ b/src/api/api-client.ts
@@ -1,10 +1,13 @@
 import isNil from 'es-toolkit/compat/isNil';
 import omitBy from 'es-toolkit/compat/omitBy';
 
+import { ValidationError } from '@/api/error';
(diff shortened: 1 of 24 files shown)


bilalesi:
I will close this one just to free some quota for preview links, will back to it when i got time

pgetta:
The issue is that the branch has to be deleted to free the quota. I'll check other branches, maybe some can be deleted as well.

bilalesi:
I deleted it too

PR #1500, opened 2026-03-11
clicking on thumbnail should show the preview

jdcourcol:
We noticed that with Georges and thought it was very annoying.


ayimaok:
I did not understand in which section this is implemented.

PR #1503, opened 2026-03-12
[CHAT] - Small plots and state fix

BoBer78:


BoBer78:
This will be done here : https://github.com/openbraininstitute/core-web-app/pull/1425

PR #1504, opened 2026-03-12
State diffs iteration 3

WonderPG:
Contributes to https://github.com/openbraininstitute/prod-ai/issues/19


g-bar (dismissed):

bilalesi:
@james-isbister that already fixed in `main`, so probably this PR needs to rebase again

WonderPG:
@james-isbister I have merged main but the merge was quite beefy. I see it has introduced a couple of small bugs that I will have to try fixing in the upcoming days but you can test anyway. Also please test that unrelated features are still working correctly on the simulation config page. @g-bar feel free to have a look to make sure that I did not mess up the stuff you merged to main recently

james-isbister:
Very nice. Here are my (subjective) comments, some of which are similar as before:
- I would call it "View changes" rather than "View diffs"
- I prefer that the diffs are not shown in read only mode. Rather the user can quickly start editing a field (and diff mode is exited). Maybe an optimization for later
- I prefer that "View diffs" is at the bottom of the white "speech bubble" response to the last prompt (rather than at the bottom of the chat). It would be ok if the "View diffs" option disappears once the configuration progresses further.
- I prefer that the diff mode shown following the response doesn't dismiss itself automatically, but dismisses when the user starts editing the config, typing a new command, or manually clicking dismiss.
- Maybe a blue border could be used for new blocks (rather than green) to avoid a clash with the green tick

Let's see what the others think @ayimaok @jdcourcol 

But perhaps it's useful to merge this first, as its been a while already

g-bar:
@WonderPG  Can you please document how the data flow goes, I'd be interesting on a step by step description of the api calls and state changes from the moment the user makes a promp, up until the user accepts / rejects a change. Otherwise it's quite difficult to understand the implementation.

jdcourcol:
>But perhaps it's useful to merge this first, as its been a while already

I agree with that. Let 's move on.

WonderPG:
@g-bar 

The flow has three phases: **real-time flashes during streaming**, **global diff view after streaming**, and **restore to a previous state**.

### 1. Streaming → flash animations

Pre-existing: `chat.ts` already watched `chat.messages` for `editstate` tool results and wrote the parsed config to `configStateAtom`.

**Added in this PR:** the same effect now also computes `oldConfig` (tracked incrementally via `lastAppliedConfigRef` to handle rapid successive calls) and writes `{ oldConfig, newConfig, counter }` to the new `lastConfigUpdateAtom` (`src/state/config-highlights.ts`).

`useConfigUpdateFlashes` (`src/features/scan-config/hooks/use-config-update-flashes.ts`), mounted once in `left.tsx`, subscribes to that atom. On each change it:
- Runs `computeLiveDiffs(old, new)` (`src/utils/diff.ts`, uses `fast-json-patch`)
- Groups diffs by root element, builds a `Map<rootElement, ActiveFlash>` with per-entry and per-field flash types
- Writes to `activeFlashesAtom`, auto-expands affected collapsed blocks
- Clears the flashes after 2.1s

The scan-config UI consumes flashes through three new hooks: `useRootElementDiff` → root tabs, `useEntryDiff` → dictionary entry tabs, `useBlockDiff` → individual fields. Each resolves a CSS class via `getDiffClassName()` (`src/utils/diff-class.ts`) that maps to animations in `src/styles/diff-highlights.css` (green=add, red=remove, amber=replace; slow=2s for tabs, fast=1.2s for fields).

### 2. Streaming ends → diff bar + "View Diffs"

`useMessageDiffs` (`src/components/ai-assistant/message-item/use-message-diffs.ts`) runs per message. It computes:
- `firstOldConfig`: walks backward through `allMessages` to find the last `editstate`/`getstate` result *before* this message's calls
- `lastNewConfig`: result of the last `editstate` in this message
- `accumulatedDiffs`: `computeLiveDiffs(firstOldConfig, lastNewConfig)`

When `status` transitions `streaming → ready`, it writes to `diffBarDataAtom`, which makes `chat.tsx` render `<DiffBar>` ("Changes available").

Clicking "View Diffs" toggles `activeDiffMessageIdAtom` in `chat.tsx`. `useMessageDiffs` reacts: when active, it writes `accumulatedDiffs` + `oldConfig` into `diffStateAtom`. The same hook chain (`useRootElementDiff`, etc.) now shows persistent colored borders (the `'highlight'` CSS variant instead of `'flash-*'`). `useDiffPreviewAtom` creates a read-only atom from the AI config so the middle panel can display the new values.

Toggling off clears `diffStateAtom` via `clearDiffStateAtom`.

### 3. Restore to a previous state

Each `CollapsibleMessage` shows a "Restore State" button for messages with completed `editstate` calls. The flow:

1. **Preview** (`handlePreviewRestore`): computes live diffs between current `agentStateAtom` and the message's snapshot, sets `restorePreviewActiveAtom = true` (suppresses the auto-apply effect in `left.tsx`), writes the snapshot to `configStateAtom` for display, and populates `diffStateAtom` with the diffs.

2. **Confirm** (`handleConfirmRestore`): writes the snapshot to `pendingRestoreConfigAtom`. A new effect in `left.tsx` picks it up and calls `resetConfig()` to apply to live atoms. Clears all diff state.

3. **Cancel** (`handleCancelRestore`): clears `configStateAtom`, the preview guard, and diff state.

Edge case: submitting a new message increments `messageSubmittedCounterAtom` (`chat.tsx`), which `CollapsibleMessage` watches to auto-cancel any pending restore confirmation. Same if another message's "Restore State" is clicked (`restorePreviewMessageIdAtom`).

### State atoms summary (all new, in `src/state/config-highlights.ts`)

| Atom | Purpose |
|---|---|
| `lastConfigUpdateAtom` | Bridge from `chat.ts` → flash system |
| `activeFlashesAtom` | Drives temporary flash animations |
| `diffStateAtom` (+ derived: `configHighlightsAtom`, `configDiffsAtom`, `oldConfigAtom`, `expandedRootElementsAtom`) | Drives persistent "View Diffs" highlights |
| `diffBarDataAtom` | Controls diff bar visibility |
| `activeDiffMessageIdAtom` | Which message's diffs are shown |
| `pendingRestoreConfigAtom` | Confirmed restore config, consumed by `left.tsx` |
| `restorePreviewActiveAtom` / `restorePreviewMessageIdAtom` | Guards during restore preview |
| `messageSubmittedCounterAtom` | Auto-cancels pending restores on new message |

I do understand that this PR is very complex. I have tried to simplify it as much as possible but the requested features are inherently complex and custom. Feel free to raise concerns and happy to discuss if needed.

src/services/ai-agent/hooks/chat.ts
```diff
@@ -33,8 +35,40 @@ export function useServiceAiAgentChat(threadId: string) {
...
+  const setLastConfigUpdate = useSetAtom(lastConfigUpdateAtom);
+  const configUpdateCounterRef = useRef(0);
+
+  // Keep a ref to aiAgentState so the React Compiler doesn't add it
+  // to the effect dependency array below (which would cause re-runs
+  // every time the live config changes, creating a stale-data loop).
+  const aiAgentStateRef = useRef(aiAgentState);
+  useEffect(() => {
+    aiAgentStateRef.current = aiAgentState;
+  }, [aiAgentState]);
```
g-bar: This is weird, it suggest to me that aiAgentState should not be an atom, but a ref.

src/services/ai-agent/hooks/chat.ts
```diff
@@ -81,21 +118,61 @@ export function useServiceAiAgentChat(threadId: string) {
...
       }
+    } catch {
+      logError(
+        'Failed to parse tool invocation result as JSON:',
+        // @ts-expect-error - result type is not properly typed
+        editstateResult.toolInvocation.result
+      );
     }
-  }, [chat.messages, setConfig, assistantInitialMessages.length]);
+  }, [chat.messages, setConfig, isLoadingMessages, assistantInitialMessages.length]);
```
g-bar: The linter is throwing this

`This hook does not specify its dependency on setLastConfigUpdate`

src/services/ai-agent/hooks/chat.ts
```diff
@@ -81,21 +118,61 @@ export function useServiceAiAgentChat(threadId: string) {
...
+    // Derive a stable ID for the invocation we found (if any).
+    const invocationId = editstateResult?.toolInvocation?.toolCallId ?? null;
+
+    // If the messages array matches the initial load exactly, mark the
+    // latest invocation as processed so we don't re-handle it, but don't
+    // actually run the editstate logic (these are historical messages).
+    if (assistantInitialMessages.length === chat.messages.length) {
+      if (invocationId) lastProcessedInvocationIdRef.current = invocationId;
+      return;
+    }
```
g-bar, line 127: Wouldn't it be simpler and cleaner to store in the invocation itself whether it has been processed?

src/services/ai-agent/hooks/chat.ts
```diff
@@ -81,21 +118,61 @@ export function useServiceAiAgentChat(threadId: string) {
...
+
+    // Mark as processed before doing work to avoid double-firing.
+    lastProcessedInvocationIdRef.current = invocationId;
+
+    // @ts-expect-error - toolInvocation result type is not properly typed
+    if (!editstateResult?.toolInvocation?.result) return;
+
+    try {
+      // @ts-expect-error - result needs to be parsed as JSON
```
g-bar: Minor but the message is confusing the ts error is the same as above (no type for "result").

src/services/ai-agent/hooks/chat.ts
```diff
@@ -81,21 +118,61 @@ export function useServiceAiAgentChat(threadId: string) {
...
+      // Only flash when the very last part is the editstate result itself
+      const lastPart = lastMessage.parts[lastMessage.parts.length - 1];
+      const isLastPartEditState =
+        lastPart?.type === 'tool-invocation' &&
+        lastPart.toolInvocation.toolName === 'editstate' &&
+        lastPart.toolInvocation.state === 'result';
+
+      if (isLastPartEditState && newConfig && editstateResult.toolInvocation.args) {
+        configUpdateCounterRef.current += 1;
+        setLastConfigUpdate({ oldConfig: oldConfig as Record<string, unknown> | null, newConfig, counter: configUpdateCounterRef.current });
```
g-bar: Why is this needed wouldn't it suffice that whoever consumes this atom, just consumes oldConfig and new Config directly?

g-bar (commented):
Left some initial comments, I'll continue the review later.

src/features/scan-config/hooks/use-config-update-flashes.ts
```diff
@@ -0,0 +1,124 @@
...
+ * Reacts to lastConfigUpdateAtom changes, computes flash maps for ALL
+ * affected root elements at once, writes to activeFlashesAtom, and
+ * auto-expands collapsed blocks. Replaces the per-instance window
+ * CustomEvent listener that was previously in each RootElement.
+ *
+ * Mount once in the parent that renders RootElement instances (left.tsx).
+ */
+export function useConfigUpdateFlashes() {
+  const configUpdate = useAtomValue(lastConfigUpdateAtom);
+  const setActiveFlashes = useSetAtom(activeFlashesAtom);
```
g-bar: If I understand things right, the activeFlashes can always be derived from the old and new configs, so you don't need an atom. The standard practice is to merely have a useMemo and compute the active flashes on the fly when either the old or new config get updated.

src/components/ai-assistant/message-item/use-message-diffs.ts
```diff
@@ -0,0 +1,311 @@
...
+    [message.parts]
+  );
+
+  const hasCompletedEditStateCalls = React.useMemo(
+    () => completedEditStateParts(message.parts).length > 0,
+    [message.parts]
+  );
+
+  /** The config state captured *before* the first editstate call in this message. */
+  const firstOldConfig = React.useMemo(() => {
```
g-bar: Isn't this simply the current config? before applying any edit updates?

g-bar: I don't understand the need for another whole hook.
Aren't the "highlights" the same as the flashes, just the former don't disappear after 2s? Can't we simply reuse those?

src/components/ai-assistant/message-item/message-item.tsx
```diff
@@ -67,6 +72,9 @@ function MessageChild({
   );
   const validStorageIds = useStableArray(memoizedStorageIds);
 
+  const { hasEditStateCalls, handlePreviewRestore, handleConfirmRestore, handleCancelRestore } =
```
g-bar, line 71: The use of this hook is very odd, the hook computes data used at the message level but at the same time has side effects that control what's rendered at the chat panel level?

Looks like some mixing of concerns, functionality should be split, the things that are rendered at the chat panel level should be mounted there.

src/services/ai-agent/hooks/chat.ts
WonderPG: I removed this pattern in 8eed072a2

src/services/ai-agent/hooks/chat.ts
WonderPG: Added `setLastConfigUpdate` to dependency array in 87929e227

src/services/ai-agent/hooks/chat.ts
WonderPG: Well in this case tool invocations are represented by a `ToolInvocationUIPart` which directly comes from the Vercel AI SDK. Adding an extra information to their schemas is very risky imo since we don't own the object, it might be randomly overridden, the types won't match anymore etc...

src/services/ai-agent/hooks/chat.ts
WonderPG: Modified in bc226e404

src/services/ai-agent/hooks/chat.ts
WonderPG: The counter was here to fix a bug related to having multiple `editstate` calls within the same AI response. Iirc I had issues with the way jotai compares the old and new config. Having the counter ensures we don't need to do deduplication on the consumer side.

src/features/scan-config/hooks/use-config-update-flashes.ts
WonderPG: Imo activeFlashes has to be an atom because flashes are time-bounded animations (appear → 2.1s → disappear), not derivable from the current config diff. A useMemo would keep the flash visible permanently as long as the configs differ, which isn't the desired UX. But let me know if i'm mistaken or if there's another way to do it

src/components/ai-assistant/message-item/use-message-diffs.ts
WonderPG: Regarding your first point:
`firstOldConfig` is the config state before the first editstate call in this specific message. It's reconstructed by walking backwards through all messages to find the last editstate or getstate result that isn't from the current message. We have to do this trick as we are storing the state ground truth in the frontend, it is the only place where this info is accessible. This is needed because by the time the user clicks "Show Diffs" on a message at the end of the stream, the live config has already been updated — you can't just read the current config. 

Regarding your second point:
- Flashes (activeFlashesAtom): Temporary 2.1s animations triggered in real-time as the AI streams `editstate` calls. They show "what just changed" and auto-clear.
- Highlights (configHighlightsAtom via diffStateAtom): Persistent diff overlays shown when the user clicks "Show Diffs" on a message, or during a restore preview. They show the accumulated diff between the pre-message config and the post-message config, and stay visible until dismissed.

So we have two different concepts: Immediate diff when `editstate` triggers and accumulated diffs across all of the `editstate` calls within one message. Both hold potentially very different information.

I think this is the crux of the complexity of this PR. A lot of features have been requested together, which seem similar on the outside but end up being very different in terms of how to manage them in the UI.

src/features/scan-config/hooks/use-config-update-flashes.ts
g-bar: `not derivable from the current config diff`, they HAVE to be derived from a given 2 versions of the config, That's what the flashes are, diffs, aren't they? 

It seems to me than only one of the two is needed, either you store in an atom the configs from which the flashes are computed, or you store the flashes, but not both.

`time-bounded animations`, that doesn't necessitate making them an atom, you could only store a flag whether to display them or not, for example.

src/components/ai-assistant/message-item/use-message-diffs.ts
g-bar: `firstOldConfig` Couldn't you make a copy before applying the updates?

But aren't the highlights, just the "accumulation" of the same diffs used for the flashes? Therefore derivable from them? or from the same configs. Whether they persist or are temporary is irrelevant as you can just have flags whether to display them.

src/components/ai-assistant/message-item/message-item.tsx
WonderPG: I did a refactor in 04ec68a57. It is quite a refactor so I need to triple check that it did not introduce regressions.

src/services/ai-agent/hooks/chat.ts
g-bar: Are you sure agentStateAtom needs to be an atom at all? from the code it seems to me it isn't reactive, ie it doesn't drive UI updates, it's merely sent to the agent. so it can merely be a `const`.

src/components/ai-assistant/message-item/use-message-diffs.ts
```diff
@@ -0,0 +1,237 @@
...
+    setDiffState,
+  ]);
+
+  const handleConfirmRestore = React.useCallback(() => {
+    const latestState = getLatestState();
+    if (latestState) {
+      setConfig(null);
+      setRestorePreviewActive(false);
+      setRestorePreviewMessageId(null);
+      setPendingRestoreConfig(latestState as Record<string, any>);
```
g-bar, line 173: I think this is another case of redundant state, `pendingRestorCconfig` can ultimately be derived from the latest tool invocation, so we can just compute it. Not sure about the other atoms here.

src/components/ai-assistant/message-item/collapsible-message/collapsible-message.tsx
```diff
@@ -1,22 +1,63 @@
...
+  const [isConfirmingRestore, setIsConfirmingRestore] = React.useState(false);
   const previousPartsLength = React.useRef(0);
 
+  // Cancel pending restore when a new message is submitted (counter increments)
+  const submittedCounter = useAtomValue(messageSubmittedCounterAtom);
+  const isChatReady = useAtomValue(isChatReadyAtom);
+  const restorePreviewMessageId = useAtomValue(restorePreviewMessageIdAtom);
+  const prevCounterRef = React.useRef(submittedCounter);
+  const onCancelRestoreRef = React.useRef(onCancelRestore);
+  onCancelRestoreRef.current = onCancelRestore;
```
g-bar: Since you're re-setting it on every re-render, thsi ref effectively does nothing.

src/components/ai-assistant/message-item/collapsible-message/collapsible-message.tsx
```diff
@@ -1,22 +1,63 @@
...
+
+  React.useEffect(() => {
+    if (submittedCounter > prevCounterRef.current) {
```
g-bar, line 49: Is this comparison needed? You would only need to detect if the counter has changed, given it cannot decrease.

src/components/ai-assistant/message-item/collapsible-message/collapsible-message.tsx
```diff
@@ -1,22 +1,63 @@
...
+  status, 
+  children,
+  onPreviewRestore,
+  onConfirmRestore,
+  onCancelRestore,
+  hasEditStateCalls = false,
+}: CollapsibleMessageProps) {
   const [collapsedIndices, setCollapsedIndices] = React.useState<Set<number>>(new Set());
   const [animatingIndex, setAnimatingIndex] = React.useState<number | null>(null);
+  const [isConfirmingRestore, setIsConfirmingRestore] = React.useState(false);
```
g-bar, line 39: Seems this state is also redundant.

g-bar (commented):
.

src/features/scan-config/hooks/use-config-update-flashes.ts
WonderPG: You are right, I tried to go with your approach in 3411dc7d8

The flash map is not stored as an atom anymore, only a boolean flag to show it or not is.

src/services/ai-agent/hooks/chat.ts
WonderPG: If I understood correctly Jotai, you are right that most of its usage isn't reactive except for this function
```
export function useAIConfig() {
  const [aiConfig, setAiConfig] = useAtom(configStateAtom);
  const [aiAgentState] = useAtom(agentStateAtom);
  const [isChatReady] = useAtom(isChatReadyAtom);

  const aiCircuitId = (aiConfig as any)?.initialize?.circuit?.id_str;
  const agentCircuitId = (aiAgentState as any)?.smc_simulation_config?.initialize?.circuit?.id_str;
  const guardPassed = aiCircuitId === agentCircuitId;

  return {
    aiConfig: guardPassed ? aiConfig : null,
    setAiConfig,
    isChatReady,
  };
}
```
which ensures the id of the circuit in the config we want to apply is the same as the one we have in the UI. As soon as a new aiAgentState arrives, we want the subscription to this atom to trigger and validate the `circuit_id`. Let me know if I got it wrong and changes are indeed needed here.

src/components/ai-assistant/message-item/use-message-diffs.ts
WonderPG: It is not purely an accumulation of all of the diffs that have happened inside a single AI message. Imagine the following scenario:
- First call to `editstate`: Add a new `Neuron Set`
- Second call to `editstate`: Delete another `Neuon Set`

For the root node `Neuron Set`, there has been one addition (green) and one deletion (red) so there is a conflict if you simply accumulate both diffs, you would need to show two colors. Diffing the final config with the initial one is the most straight forward way to remove conflicts.

Regarding `firstOldConfig`, I remember that there was a reason to do it this way. However I cannot remember it so I have made the changes you suggested in f81deb701. Thanks !

src/components/ai-assistant/message-item/use-message-diffs.ts
WonderPG: The config to restore isn't always the latest — it's from whichever message the user clicked "Restore State" on. And the restore only happens when the user explicitly confirms. You can't derive "the user confirmed a restore" from the tool invocation data.

There elements are used for the button `restore state` on the collapsible component that appears whenever there is an `editstate` tool call as one of the step in the message. It is something that entirely depends on a user action, not on the LLM stream/tool calling.

Just to add a bit of context, we originally had the accept/reject button for pending changes from the LLM. We have decided to remove that to improve the flow of setting up a simulation config from the chat. However, people were afraid that the LLM could mess up and break your config. We have decided to add this layer of "safety" which allows the user to restore previous LLM generated states. By restoring the state of an older message, you are setting back the output of the very last `editstate` tool call from the message you are restoring.

src/components/ai-assistant/message-item/collapsible-message/collapsible-message.tsx
WonderPG: Good catch, removed the ref and added `onCancelRestore` in the `useEffect`'s dependency array in f6756e1b6

src/components/ai-assistant/message-item/collapsible-message/collapsible-message.tsx
WonderPG: It could be `!==` instead of `>`. Or do you have something else in mind ?

src/components/ai-assistant/message-item/collapsible-message/collapsible-message.tsx
WonderPG: `isConfirmingRestore` is local UI state that controls which buttons are shown in this specific message's collapsible header. It's toggled by user clicks and auto-dismissed by several external events. It can't be cleanly derived from the global atoms because the dismiss timing is controlled locally.

WonderPG:
@g-bar Thank you very much for the thorough review btw. I understand it is an involved one. Appreciate it.

src/services/ai-agent/hooks/chat.ts
g-bar: What's really driving the UI updates is the aiConfig, not the aiAgentState, since the latter always "trails" the former it can't go stale, so a mere `const` would suffice. However I don't see any major problem in this either.

src/components/ai-assistant/message-item/collapsible-message/collapsible-message.tsx
g-bar: submittedCounter can never decrease and even if set with the same value it won't trigger a state update, so simply this should be safe:

```ts
React.useEffect(() => {

      setIsConfirmingRestore(false);
      onCancelRestore?.();
    
    
  }, [submittedCounter]);
```

src/components/ai-assistant/message-item/collapsible-message/collapsible-message.tsx
g-bar: Yes, you're right.

src/components/ai-assistant/message-item/collapsible-message/collapsible-message.tsx
g-bar: Actually, the above doesn't apply to react strict mode, therefore we can go with your original comparison.

src/services/ai-agent/hooks/chat.ts
g-bar: Now `jotaiStore.get` is missing from the deps.

g-bar:
There are several linter errors.

src/components/ai-assistant/panel-content/chat/use-last-message-diff-bar.ts
```diff
@@ -0,0 +1,113 @@
...
+ * message-level and panel-level concerns separate.
+ *
+ * Mount once in the Chat component.
+ */
+export function useLastMessageDiffBar(
+  messages: UIMessage[],
+  status: 'submitted' | 'streaming' | 'ready' | 'error',
+) {
+  const [activeDiffMessageId] = useAtom(activeDiffMessageIdAtom);
+  const [, setDiffBarData] = useAtom(diffBarDataAtom);
```
g-bar: This doesn't need to be an atom, you can use local state in the hook and return it for the consumer component to use.

src/components/ai-assistant/message-item/use-message-diffs.ts
g-bar: Yes, I meant the latest from the message.
could we not simply store the message Id the user wishes to restore, then derive the actual config instead of storing the full config?

WonderPG: `left.tsx` has no access to the messages, so it can't derive the config from a message ID. 

Storing just the message ID would require left.tsx to access the chat messages array and understand the editstate tool invocation format to derive the config. Since left.tsx is in the scan-config tree (not the chat tree), it doesn't have access to messages, and adding that dependency would tightly couple the two domains. The atom acts as a handoff, the chat side resolves the config, the scan-config side just applies it.

g-bar: Yes, good points.

src/components/ai-assistant/panel-content/chat/use-last-message-diff-bar.ts
WonderPG: You are right. Done in bbc8dbc0a

WonderPG:
Let me fix the linter still

src/services/ai-agent/hooks/chat.ts
WonderPG: I added `jotaiStore` in 9c2f2b8e8

WonderPG:
I ran the linter in 9c2f2b8e8 on my files. Note that the whole repo is heavily nested with >1k linting issues
image

g-bar (approved):
Great job!

g-bar:
> I ran the linter in [9c2f2b8](https://github.com/openbraininstitute/core-web-app/commit/9c2f2b8e88627084ec834be507b9bf4a8b1868e4) on my files. Note that the whole repo is heavily nested with >1k linting issues image

Yeah, we still have some things to fix since the change to biome, but we try not to introduce new errors. Thanks!

PR #1535, opened 2026-03-24
Sonata voltage report viewer

pgetta:
## Summary

  Add a new Sonata voltage report viewer for visualizing SONATA HDF5 simulation results. The viewer parses `.h5` voltage report files (soma reports) in a Web Worker using `h5wasm` and
  renders interactive Plotly charts.

Screenshot 2026-03-24 at 17 05 23

  ### New feature: `src/features/sonata-viewer/`

  - **Overview mode** — thumbnail grid of all node traces per population, with lazy loading (intersection observer) and container queries for responsive layout. Clicking a thumbnail
  navigates to the interactive view for that node.
  - **Interactive details mode** — full Plotly charts with zoom/pan support. Zoom triggers re-downsampling from the worker for detail on demand. Population and multi-select node selectors
   let users filter which traces to display.
  - **Web Worker (`comlink` + `h5wasm`)** — HDF5 parsing and data extraction run off the main thread. The worker reads the SONATA `/report/<population>/` structure, extracts per-node time
   series from the flat `[timesteps × compartments]` matrix, and applies LTTB downsampling before transferring data back.
  - **LTTB downsampling** — Largest Triangle Three Buckets algorithm for perceptually accurate downsampling of large time series (overview: 100 points, interactive: 1000 points).
  - **Dynamic import** — `SonataViewer` is loaded with `next/dynamic` (SSR disabled) since it depends on browser-only APIs (Web Workers, h5wasm/Emscripten FS).

  ### Integration

  - Wired into the file viewer (`src/features/scan-config/components/file-viewer/index.tsx`) — H5 assets with `voltage_report` label are routed to the new `SonataViewer`.

  ### Minor improvements to existing ephys viewer

  - Extracted `CHART_LINE_COLOR` constant (`#0050b3`) shared between ephys and sonata viewers.
  - Made overview thumbnails clickable as full `<button>` elements instead of a separate icon button, improving hit target and simplifying markup.
  - Lowercased "Select Cell" → "Select cell" labels for consistency.

  ## Architecture

  ```
  sonata-viewer/
  ├── index.tsx                    # Dynamic import entry point (SSR disabled)
  ├── sonata-viewer.tsx            # Main component (overview ↔ interactive toggle)
  ├── types.ts                     # Shared types (metadata, trace data, downsample request)
  ├── constants.ts                 # Chart line color
  ├── components/
  │   ├── report-overview.tsx      # Thumbnail grid with lazy-loaded plots
  │   ├── report-details-view.tsx  # Interactive plots with population/node selectors
  │   ├── interactive-plot.tsx     # Single interactive Plotly chart with zoom-on-demand
  │   ├── population-select.tsx    # Population dropdown
  │   └── node-selector.tsx        # Multi-select node picker with "All" toggle
  ├── hooks/
  │   ├── use-sonata-report.ts     # Data fetching + worker lifecycle hook
  │   └── config-hooks.ts          # Plotly layout/config hooks (overview + interactive)
  └── worker/
      ├── sonata-worker.ts         # h5wasm HDF5 parser + LTTB downsampler (runs in Worker)
      └── worker-api.ts            # Comlink wrapper for worker instantiation/teardown
  ```

  ## Test plan

  - [x] Open a simulation result entity that has an H5 asset with `voltage_report` label
  - [x] Verify the overview loads with thumbnail traces for all nodes
  - [x] Verify thumbnails lazy-load as they scroll into view
  - [x] Click a thumbnail → confirm it switches to interactive view with that node pre-selected
  - [x] In interactive view, zoom into a region → verify the chart re-renders with higher-resolution data
  - [x] Double-click to reset zoom → verify full trace is restored
  - [x] Switch populations via the dropdown → verify node list and charts update
  - [x] Use the multi-select node picker: select "All", deselect "All", select individual nodes
  - [ ] ~~Test with files containing multiple populations~~ - to be tested once we have such files
  - [ ] ~~Test with a file >100MB → verify the size limit error is shown~~ - to be tested once we have such files
  - [x] Verify no main-thread jank during HDF5 parsing (worker offloading)
  - [x] Verify responsive layout at different viewport widths


bilalesi (dismissed):
thank you Pavlo

src/features/scan-config/use-cases/simulations/results.tsx
```diff
@@ -439,12 +467,14 @@ export default function SimulationsTab({
 
       {/* Preview for selected file */}
       <div className="relative pl-4">
-        <FileViewer
-          file={selectedFile}
-          className="h-full"
-          context={context}
-          loading={filesLoading}
-        />
+        <SimulationReportsProvider reports={simConfig?.reports ?? null}>
```
bilalesi, line 470: is this `reports` shared across all the files in the viewer, 
if not i would suggest to split the viewer responsibility, this way we are sure not mixing the logic and the input/output props

src/features/sonata-viewer/sonata-viewer.tsx
```diff
@@ -0,0 +1,106 @@
...
+import type { ISimulationResult } from '@/api/entitycore/types/entities/simulation-result';
+import type { WorkspaceContext } from '@/types/common';
+
+const ErrorFallback = SimpleErrorComponent as unknown as ComponentType<FallbackProps>;
+
+enum VIEW {
+  OVERVIEW = 'overview',
+  INTERACTIVE = 'interactive',
+}
+
```
bilalesi, line 24: this is just my preference, but also to harmonize the app, 
probably using 
```ts
const View =  { Overview: "overview", ...} as const
type TView = (typeof View)[keyof typeof View]

if(View.Overview) ....
```
I agree with this argument [here ](https://www.totaltypescript.com/why-i-dont-like-typescript-enums) and also there is double conversion in types which is bad

bilalesi (approved):
I approve, I left few comments, can be addressed later after production depl

src/features/scan-config/use-cases/simulations/results.tsx
pgetta: Good point, I've noted and will address separately.

src/features/sonata-viewer/sonata-viewer.tsx
pgetta: That's a hilarious read, thank you! I'll implement your suggestion.

PR #1547, opened 2026-03-30
feat-17: migrate extraction to the new task config/activity

bilalesi:
1. the main migration is: extraction no longer uses its own campaign/config/execution flow. It now uses generic `task_config` and `task_activity`

2. the old dedicated extraction campaign api path is being removed from the main flow, and all the logic will be deprecated in another PR,
the new pattern is:
campaign `task_config` -> config-generation `task_activity` -> generated `task_config` -> execution `task_activity`

3. for the new task management, we have now 
```ts
src/api/entitycore/types/entities/task-activity.ts
src/api/entitycore/types/entities/task-config.ts // `ITaskConfig` is now generic and typed by meta
```
`src/api/entitycore/queries/task` that contains everything related to the new task api

4. there is a new shared helper layer for task workflows. It builds campaign rows, links generations/configs/executions, and computes latest status / aggregated status counts. this is the reusable part other simulation modules can use
`src/entity-configuration/domain/task-helpers.ts`: `buildTaskCampaignRows` it return the campaign and all the executions in this format
```ts
export type TTaskCampaignExecutionRow<
  TMeta extends Record<string, unknown> = Record<string, unknown>,
> = {
  execution: ITaskActivity | null;
  provenance: {
    config: ITaskConfig<TMeta>;
    generation: ITaskActivity;
  };
};
```
this is used in `resolveExtractionCampaigns` and it can be copied in other simulations in the way to unifying the API,  next step will be having one function to resolve any campaign and can be extended for special use cases or exposing specific data.

5. launching and polling were refactored into shared `task-runner` utilities, instead of extraction-specific launch/query logic, the UI now uses generic helpers to load configs, load executions, poll active runs, and launch Obi One tasks
you can find the implementation here `src/features/scan-config/task-runner.ts` 
this is also can be used across all simulations

6. workflow activity table was updated to treat new extraction campaigns as TaskConfig records. Expanded rows, status badges, duplicate flow, and detail page resolution now read from task data and we have a shared expandable table implementation here `src/entity-configuration/definitions/list-expanded-view-defs/task-activity.tsx`.

7. a new module, src/api/entitycore/selectors/assets.ts, has been introduced to provide asset retrieval capabilities across multiple formats, the api exposes a well defined interface backed with comprehensive tests.
the api allow to provide both `label` and `contentType` (at least one should be provided)
```ts
getOneOrThrow
getAllOrThrow
getOneOrNull
getAllOrNull
````

fyi: @pgetta this should help to avoid duplicate work and also unifying the modules for all activities from now on


pgetta (approved):
Nice work!

PR #1555, opened 2026-04-01
Course vlabs

g-bar:


src/api/entitycore/types/entities/analysis-notebook-template.ts
```diff
@@ -105,5 +105,7 @@ export interface IAnalysisNotebookTemplate
     EntityCoreOwnership,
     TAnalysisNotebookTemplateBase {
   description: string;
+  scale: 'subcellular' | 'cellular' | 'circuit' | 'system';
+  exercise_id?: string | null;
```
bilalesi: i think exercise_id is deprecated ?

src/api/entitycore/types/index.ts
```diff
@@ -65,4 +66,10 @@ export type EntityCoreObjectTypes =
   | IonChannelModel
   | IIonChannelModelingCampaign
   | IEMCellMesh
+  | IAnalysisNotebookTemplate
   | ISimulatableExtracellularRecordingArray;
+
+export function isNotebook(entity: EntityCoreObjectTypes): entity is IAnalysisNotebookTemplate {
+  if (entity.type === 'analysis_notebook_template') return true;
+  return false;
+}
```
bilalesi, line 76: it can be just `return (entity.type === 'analysis_notebook_template')` 
and may be using the extendedType dict is better `ExtendedEntitiesTypeDict.AnalysisNotebookTemplate` (it helps to use the single source of truth of available types)

src/features/notebooks/components/notebooks-layout.tsx
```diff
@@ -2,71 +2,105 @@
 
 import { LoadingOutlined, PlusOutlined } from '@ant-design/icons';
 import { useQuery } from '@tanstack/react-query';
-import { domAnimation, LazyMotion, m } from 'framer-motion';
```
bilalesi, line 5: why removing the animation ?

src/features/notebooks/components/notebooks-layout.tsx
```diff
@@ -2,71 +2,105 @@
...
-function handleUploadData() {
-  makeSelectContributionEntityClickEvent({
-    display: true,
-    entityType: ExtendedEntitiesTypeDict.AnalysisNotebookTemplate,
-    sessionId: crypto.randomUUID(),
+export async function createNotebook({
+  payload,
+  context,
+}: {
+  payload: IAnalysisNotebookTemplate;
```
bilalesi: this can move to the api side src/api/entitycore/queries/analysis-notebook-template.ts

src/features/notebooks/components/notebooks-layout.tsx
```diff
@@ -92,24 +126,66 @@ export function NotebooksLayout({ children }: Props) {
...
-    <LazyMotion features={domAnimation}>
-      <div
-        id="notebooks-layout"
-        data-testid="notebooks-layout"
-        className="bg-background grid h-full w-full grid-cols-[22rem_1fr] grid-rows-[auto_1fr] gap-2 overflow-hidden [grid-template-areas:'header_header''main_main'] pr-1"
-      >
-        <div className="flex w-full items-center justify-between gap-4 pl-3 [grid-area:header]">
-          <div className="flex min-w-0 max-w-1/2 items-center gap-2">
-            <TabsSelector activeTab={scope} onValueChange={onScopeChange} />
-          </div>
```
bilalesi, line 109: why remove the `TabsSelector` ?
this is unique and shared across all use cases in the app, also we don't have public/private in the URLs
this is a ticket that was resolved by this changes, to use the same logic across the whole application  as data and workflows (query param instead of param path)

src/features/notebooks/components/notebooks-layout.tsx
```diff
@@ -127,58 +203,120 @@ export function NotebooksLayout({ children }: Props) {
...
-    </LazyMotion>
+async function _syncNotebook({
+  notebook,
+  virtualLabId,
+  projectId,
+  targetProjectId,
+}: {
+  notebook: IAnalysisNotebookTemplate;
+  virtualLabId: string;
+  projectId: string;
```
bilalesi: why have business logic in the presentation layer better to move is into `/services` folder 
(valid for all functions below)

src/features/notebooks/components/notebooks-layout.tsx
g-bar: sorry i made a mistake when rebasing, it's back

src/features/notebooks/components/notebooks-layout.tsx
g-bar: it was a mistake when rebasing, it's back

src/api/entitycore/types/entities/analysis-notebook-template.ts
g-bar: I don't know about this, is it deprecated @pgetta ?

bilalesi: i checked entitycore: https://github.com/openbraininstitute/entitycore/blob/32666e617b522dec6b8c3e34901fd9fd6f9a56e9/app/schemas/analysis_notebook_template.py#L37 (swagger: https://staging.openbraininstitute.org/api/entitycore/docs#/analysis-notebook-template/read_many_analysis_notebook_template_get)

pgetta: It has got replaced by `assignment_id`, both are optional strings.

pgetta: https://github.com/openbraininstitute/core-web-app/blob/82a8dbba369d0d33db91113a2e1cb452eaed92f1/src/api/entitycore/types/entities/analysis-notebook-template.ts#L84

src/app/app/course-enrolment/page.tsx
```diff
@@ -0,0 +1,203 @@
...
+export default function CourseEnrolmentPage() {
+  const config = useConfig();
+  const router = useRouter();
+  const searchParams = useSearchParams();
+  const enrolmentId = searchParams.get('enrolment_id');
+  const [error, setError] = useState<ClaimError | null>(null);
+  const [success, setSuccess] = useState<ClaimSuccessData | null>(null);
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
```
bilalesi, line 36: can't be this `useQuery`?

src/app/app/course-enrolment/page.tsx
```diff
@@ -0,0 +1,203 @@
...
+  start_date?: string;
+}
+
+const ERROR_MESSAGES: Record<string, string> = {
+  'Enrolment not found': 'The enrolment could not be found. Please check your link.',
+  'This enrolment has been dropped': 'This enrolment has been dropped and cannot be claimed.',
+  'This enrolment has already been claimed':
+    'This enrolment has already been claimed by another user.',
+  'Cannot claim enrolment: course is in':
+    'The course is no longer active and cannot accept new claims.',
```
bilalesi: why do you think to have code_error for the keys (probably need to change in vlab-svc) instead of human readable strings ?

src/api/virtual-lab-svc/queries/project.ts
```diff
@@ -122,6 +122,33 @@ export async function listProjects({
   });
 }
 
+export async function listAllProjectIds(virtualLabId: string) {
```
bilalesi, line 126: does this  `src/utils/pagination.ts` help in this case?

src/components/activate-enrolments.tsx
```diff
@@ -0,0 +1,29 @@
+'use client';
+
+import { useEffect } from 'react';
+
+import { authFetch } from '@/auth-fetch';
+import { useConfig } from '@/config';
+
+export function ActivateEnrolmentsOnLoad() {
+  const config = useConfig();
```
bilalesi, line 13: this is will triggered every time the user visit the page ?

src/features/notebooks/components/notebooks-layout.tsx
```diff
@@ -96,6 +100,27 @@ export function NotebooksLayout({ children }: Props) {
     setLoading(false);
   }
 
+  const course = virtualLabData?.course;
+
+  const onNotebookCreateSuccess = useCallback(
+    async (notebook: EntityCoreObjectTypes) => {
+      if (!isNotebook(notebook) || course?.template_project_id !== projectId) {
```
bilalesi: may be passing the same logic to `ContributionModal` to not open modal/display the button only if this condition is met ?

src/services/notebooks/sync-template-notebooks.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (studentProjectIds.length === 0) return;
+
+  const allNotebooks = [];
+  let page = 1;
+  const pageSize = 100;
+  let hasMore = true;
+
+  while (hasMore) {
+    const response = await getAnalysisNotebookTemplates({
+      filters: { page, page_size: pageSize },
```
bilalesi: this `src/utils/pagination.ts` can be helpful may be ?

src/services/notebooks/sync-template-notebooks.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  if (allNotebooks.length === 0) return;
+
+  await Promise.all(
+    allNotebooks.map((notebook) =>
+      syncNotebook({
+        notebook,
+        virtualLabId: context.virtualLabId,
+        projectId: templateProjectId,
```
bilalesi: what happen if the sync failed ?

src/ui/segments/contribute/analysis-notebook-template/steps/setup.tsx
```diff
@@ -18,7 +22,39 @@ const SCALE_OPTIONS = ScaleEnum.options.map((value) => ({
...
+  const validateNameUniqueness = useCallback(
+    async (_: unknown, value: string) => {
+      if (!value) return;
+
+      return new Promise<void>((resolve, reject) => {
+        if (debounceTimerRef.current) {
+          clearTimeout(debounceTimerRef.current);
+        }
+
+        debounceTimerRef.current = setTimeout(async () => {
```
bilalesi: you don't have to debounce here, the `Form.Item` already have a debounce prop that can help clean up the logic

src/ui/segments/project/course-assign-seats-modal.tsx
```diff
@@ -0,0 +1,319 @@
...
+    setFilteredStudents([]);
+    setExistingStudents([]);
+    setError('');
+    setResults([]);
+  };
+
+  const assignMutation = useMutation({
+    mutationFn: async (students: Student[]) => assignSeats(courseId, students),
+    onSuccess: async (data) => {
+      setResults(data.results || []);
```
bilalesi: you can have the data from the mutation results, rather then having another state
```TS 
{ mutateAsync, isPending, data } = useMutation(...)
```

src/components/activate-enrolments.tsx
g-bar: Yes, basically they get access to the course  after a predefined "start_date", at first i put it on log-in, but that misses the case where they are already logged in when the course starts.

The backend is already a no-op in case enrolment is already activated, and there will be only very few courses per semester so very little overhead.

src/features/notebooks/components/notebooks-layout.tsx
g-bar: That's already the case, the button doesn't show up if condition is met, the modal doesn't need it if it can't be triggered.

src/services/notebooks/sync-template-notebooks.ts
g-bar: for right now it fails silently (logged), I might add some button to "resync" in case of failure but it's out of scope for now.

src/app/app/course-enrolment/page.tsx
g-bar: I don't understand the question, they seem to be pretty readable to me?

bilalesi: sorry a typo, ~why~ (what) ,
it's not about understandability, it's about best practices, usually an api send code_error, and a message, 
the `code_error` is what's used to determine what message to show, not a message

src/app/app/course-enrolment/page.tsx
g-bar: No, it's a post request and there is no user action, it runs on page load.

src/app/app/course-enrolment/page.tsx
g-bar: ah right, the api already returns error codes, I'll change the map to use those instead of the message.

g-bar: Since multiple messages share same code, (ie 409 conflict), we can't map 1-1, but we can just use the api message which is pretty user friendly. I changed it to that.

src/ui/segments/contribute/analysis-notebook-template/steps/setup.tsx
g-bar: Done

src/services/notebooks/sync-template-notebooks.ts
g-bar: Yes, changed to use it.

src/api/virtual-lab-svc/queries/project.ts
g-bar: Done

.github/workflows/deploy-preview.yml
```diff
@@ -205,7 +206,7 @@ jobs:
 
             let status = 'PENDING';
             const startTime = Date.now();
-            const maxWaitTime = 10 * 60 * 1000; // 10 minutes
+            const maxWaitTime = 20 * 60 * 1000; // 20 minutes
```
pgetta: We should have different values for this entry and the action's `timeout-minutes`.

You can think of these as a soft and a hard limit.

When the script times out (soft limit) - it proceeds to update the build comment status to `Failed`.
While the hard limit (`timeout-minutes`) just kills the job and the build comment status is getting stuck in the "Building" state.

I've checked Amplify and so far most of the builds are below 10 minutes, with some exceptions going up to ~12 minutes. 

Maybe 15 minutes is a reasonable value for the `maxWaitTime` for now?

.env.development
```diff
@@ -36,11 +36,12 @@ OBI_ONE_URL="https://staging.cell-a.openbraininstitute.org/api/obi-one"
 SMALL_SCALE_SIMULATOR_URL="https://staging.cell-a.openbraininstitute.org/api/small-scale-simulator"
 THUMBNAIL_API_URL="https://staging.cell-a.openbraininstitute.org/api/thumbnail-generation"
 VIRTUAL_LAB_API_URL="https://staging.cell-a.openbraininstitute.org/api/virtual-lab-manager"
+# VIRTUAL_LAB_API_URL="http://localhost:8000"
```
pgetta: Best to clean that up.

For the local overrides you can create `.env.development.local`, it is git ignored.

src/app/app/course-enrolment/page.tsx
```diff
@@ -0,0 +1,184 @@
...
+        if (!response.ok) {
+          const errorData = await response.json().catch(() => ({}));
+          const displayMsg = errorData.message || 'Failed to claim enrolment';
+
+          setError({ message: displayMsg, display: displayMsg });
+          setLoading(false);
+          return;
+        }
+
+        const responseData = await response.json();
```
pgetta: The claim response is untyped, and `data.course?.virtual_lab_name` (used below) does not exist on any `Course` type in the codebase. Please define an interface for this response so the field accesses are checked.

src/app/app/course-enrolment/page.tsx
```diff
@@ -0,0 +1,184 @@
...
+              },
+            });
+          } catch (error) {
+            console.error('Failed to activate enrolments', error);
+          }
+          router.push(`/app/virtual-lab/${data.course.virtual_lab_id}/${data.project_id}`);
+        } else {
+          // Course hasn't started yet - show success message
+          setSuccess({
+            virtual_lab_id: data.course.virtual_lab_id,
```
pgetta: `data.course` is accessed with optional chaining elsewhere (`data.course?.start_date`) but unguarded here. If the claim response does not include `course`, this throws, and the user is shown "An unexpected error occurred" even though the claim succeeded. Please guard this branch.

src/app/app/course-enrolment/page.tsx
```diff
@@ -0,0 +1,184 @@
...
+      } catch (err) {
+        setError({
+          message: err instanceof Error ? err.message : 'Unknown error',
+          display: 'An unexpected error occurred. Please try again.',
+        });
+        setLoading(false);
+      }
+    };
+
+    claim();
```
pgetta, line 108: This effect performs a mutating POST. In development, StrictMode runs effects twice, so the claim endpoint is called twice; if it is not idempotent, the second call may surface an error for an already-claimed seat. Consider an in-flight/completed guard (e.g. a ref) or moving the claim behind an explicit user action.

src/app/app/course-enrolment/page.tsx
```diff
@@ -0,0 +1,184 @@
...
+        <div className="mx-auto w-full max-w-md text-center">
+          <ErrorComponent
+            error={new Error(error.display)}
+            customError={error.display}
+            showButtons={false}
+          />
+          <div className="mt-6">
+            <Button
+              onClick={() => {
+                router.push('/app/virtual-lab/sync');
```
pgetta: Hardcoded path; the rest of the app builds these from `config.ROOT_ROUTE`. Same on line 173.

src/components/activate-enrolments.tsx
```diff
@@ -0,0 +1,29 @@
...
+          headers: {
+            'content-type': 'application/json',
+          },
+        });
+      } catch (error) {
+        console.error('Failed to activate enrolments', error);
+      }
+    };
+
+    activate();
```
pgetta: This fires an authenticated POST on every `/app` mount for every user, although almost none will have pending enrolments, and the claim page already calls the same endpoint. Consider invoking it only from the claim flow, or gating it on a signal that the user actually has an enrolment to activate. Note it also runs twice in development due to StrictMode.

src/services/notebooks/index.ts
```diff
@@ -213,3 +222,91 @@ export async function startEmptyNotebook(
...
+  virtualLabId,
+  projectId,
+  targetProjectIds,
+}: {
+  notebook: IAnalysisNotebookTemplate;
+  virtualLabId: string;
+  projectId: string;
+  targetProjectIds: string[];
+}) {
+  const promises = targetProjectIds.map((id) => {
```
pgetta: `Promise.all` over all target projects, combined with the per-notebook `Promise.all` in `sync-template-notebooks.ts`, produces unbounded concurrency (students x notebooks x assets requests fired at once). It also short-circuits on the first failure while leaving already-created notebooks in place, so a retry would create duplicates (name uniqueness is only enforced client-side in the form). Suggest batching with a small concurrency limit, using `Promise.allSettled`, and surfacing which targets failed.

src/api/entitycore/queries/analysis-notebook-template.ts
```diff
@@ -87,3 +87,21 @@ export async function deleteAnalysisNotebookTemplate({
     },
   });
 }
+
+export async function createNotebook({
```
pgetta: The payload here is the full existing entity, including `id`, `assets`, and other server-managed fields; this works only while the backend silently ignores them. Consider building an explicit creation payload. Also, for consistency with the neighbouring functions, `createAnalysisNotebookTemplate` would be a better name.

src/services/notebooks/sync-template-notebooks.ts
```diff
@@ -0,0 +1,50 @@
...
+          authorized_public: false,
+        },
+        context: { ...context, projectId: templateProjectId },
+      }),
+    pageSize: 100,
+  });
+
+  if (allNotebooks.length === 0) return;
+
+  await Promise.all(
```
pgetta: See the comment in `services/notebooks/index.ts`: this multiplies the fan-out and shares the same partial-failure behaviour. `Promise.allSettled` plus failure reporting would be safer.

src/features/notebooks/components/mini-detail-actions.tsx
```diff
@@ -98,6 +111,47 @@ export function NotebookActions<T extends EntityCoreObjectTypes>({
     setPendingDownload(false);
   };
 
+  const canDelete =
```
pgetta, line 134: While the virtual lab query is still loading, `virtualLabData` is `undefined`, so `!virtualLabData?.course` is `true` and the delete button renders for students in course projects until the data arrives, then disappears. Consider gating on the query having resolved.

pgetta (commented):
## Review summary

This PR introduces course-based virtual labs: enrolment claiming (`/app/course-enrolment`), a course management page (seats, enrolments, CSV-based seat assignment, seat dropping with recoverability rules), template-to-student notebook syncing, and a notebook delete action. It also changes the contribution pipeline contract (`createEntity` now returns the created entity instead of its id) and includes some unrelated changes (preview-deploy timeout increase, `parseApiError` improvement).

File-specific issues are in the inline comments. The following points are not attached to a diff line:

**Regression in `src/ui/segments/contribute/experimental-neuron-density/pipeline.ts` (file not touched by this PR).** `IPipelineHookResult.createEntity` was changed to return `Promise<EntityCoreObjectTypes>`, and `contribution-form.tsx` now calls `setCreatedEntityId(entity.id)`. The experimental-neuron-density pipeline was not migrated and still returns the id string, so `entity.id` evaluates to `undefined` at runtime. After creating an experimental neuron density, the form never enters its created state (step navigation stays enabled, the details/done button never appears). CI does not catch this because the build sets `ignoreBuildErrors: true`. All other pipelines were updated; this one needs the same change.

**Backend authorization.** The course page and the notebook delete action rely entirely on server-side authorization; the client only hides UI. Please confirm that virtual-lab-manager restricts `/seats/courses/*` and `/courses/{id}/enrolments` to lab admins (the responses contain student emails and ids), and that entitycore rejects unauthorized notebook deletions.

**Tests.** No tests accompany the new logic. `parseCSV`, the `SeatRecoverability` branching (period-1 cutoff, previously-dropped seat, credit threshold), and the enrolment sorting are pure and inexpensive to unit-test; the seat/credit rules in particular would benefit from coverage.

**PR description.** The PR body is empty. Given the size (44 files) and the number of behavioural changes, please add a description covering the scope, the pipeline contract change, and the notebook sync semantics.

**Scope.** The preview-deploy timeout change and the `parseApiError` change are unrelated to the feature and would be easier to track in separate PRs.

P.S. Obviously this code review is LLM generated, but I think it's pretty good.

src/features/notebooks/components/mini-detail-actions.tsx
```diff
@@ -98,6 +111,47 @@ export function NotebookActions<T extends EntityCoreObjectTypes>({
...
+  const deleteMutation = useMutation({
+    mutationFn: () =>
+      (isTemplate ? deleteAnalysisNotebookTemplate : deleteAnalysisNotebookResult)({
+        id: record.id,
+        context: { virtualLabId, projectId },
+      }),
+    onSuccess: async () => {
+      await queryClient.invalidateQueries({
+        predicate(query) {
+          const first = query.queryKey[0] as
```
pgetta: This predicate couples to the internal shape of the query key. Since the keys are produced by the builders in `src/ui/use-query-keys/`, a builder-based predicate (or invalidating by the builder-produced key) would be less brittle.

src/app/app/virtual-lab/[virtualLabId]/[projectId]/course/page.tsx
```diff
@@ -0,0 +1,446 @@
...
+import { Skeleton } from '@/ui/molecules/skeleton';
+import { Tooltip, TooltipContent, TooltipTrigger } from '@/ui/molecules/tooltip';
+import { AssignSeatsModal } from '@/ui/segments/project/course-assign-seats-modal';
+import { DropSeatButton } from '@/ui/segments/project/drop-seat-button';
+import { SeatRecoverability } from '@/ui/segments/project/seat-recoverability';
+
+type SortField = 'email' | 'student_id' | 'status' | 'activated' | 'dropped' | 'created' | 'seat';
+type SortOrder = 'asc' | 'desc';
+
+export default function CoursePage() {
```
Comment (pgetta) [inline src/app/app/virtual-lab/[virtualLabId]/[projectId]/course/page.tsx:76]: This route has no client-side role gate: the nav link is hidden for non-admins, but anyone with the URL renders the page and triggers the seats/enrolments fetches. Even if the backend enforces authorization, applying the same `isVlabAdmin` check used in `top-menu-nav.tsx` here would provide defence in depth and a better failure mode than raw query errors.

src/app/app/virtual-lab/[virtualLabId]/[projectId]/course/page.tsx
```diff
@@ -0,0 +1,446 @@
...
+  const [sortField, setSortField] = useState<SortField>('email');
+  const [sortOrder, setSortOrder] = useState<SortOrder>('asc');
+  const queryClient = useQueryClient();
+
+  const handleAssignSuccess = () => {
+    queryClient.invalidateQueries({ queryKey: ['seats', courseId] });
+    queryClient.invalidateQueries({ queryKey: ['enrolments', courseId] });
+  };
+
+  const handleDropSuccess = () => {
```
Comment (pgetta) [inline src/app/app/virtual-lab/[virtualLabId]/[projectId]/course/page.tsx]: `handleAssignSuccess` and `handleDropSuccess` are identical; they can be a single handler. Both also close over `courseId`, which is declared further down, so declaring the handler after `courseId` would read more clearly.

src/app/app/virtual-lab/[virtualLabId]/[projectId]/course/page.tsx
```diff
@@ -0,0 +1,446 @@
...
+  const handleSort = (field: SortField) => {
+    if (sortField === field) {
+      setSortOrder(sortOrder === 'asc' ? 'desc' : 'asc');
+    } else {
+      setSortField(field);
+      setSortOrder('asc');
+    }
+  };
+
+  const SortableHeader = ({ field, label }: { field: SortField; label: string }) => {
```
Comment (pgetta) [inline src/app/app/virtual-lab/[virtualLabId]/[projectId]/course/page.tsx]: Defining `SortableHeader` inside the component creates a new component type on every render, so React unmounts and remounts each header cell on every sort click or query update. Please hoist it to module level and pass the state via props. The clickable `th` also lacks keyboard support (`tabIndex`/`onKeyDown`).

src/app/app/virtual-lab/[virtualLabId]/[projectId]/course/page.tsx
```diff
@@ -0,0 +1,446 @@
...
+            />
+          </svg>
+        </div>
+      </th>
+    );
+  };
+
+  // Fetch virtual lab to get course ID
+  const labQuery = useQuery({
+    queryKey: ['virtualLab', virtualLabId],
```
Comment (pgetta) [inline src/app/app/virtual-lab/[virtualLabId]/[projectId]/course/page.tsx]: This ad-hoc key caches `getVirtualLab` separately from `keyBuilder.getOneLab(...)` used in `top-menu-nav.tsx`, so the same request is fetched and cached twice, and invalidations miss one of the consumers. Query keys are kept in `src/ui/use-query-keys/`; please use `getOneLab` here (and in `browse-entity.tsx`), and add builders for the `seats`/`enrolments` keys as well.

src/ui/segments/workspaces/top-menu-nav.tsx
```diff
@@ -139,7 +159,44 @@ export function TopMenuNavigation() {
...
+    enabled: Boolean(virtualLabId),
+  });
+
+  const { data: userGroups } = useQuery({
+    queryKey: userKeyBuilder.groups(),
+    queryFn: () => getUserGroups(),
+    gcTime: 0,
+    staleTime: 0,
+    refetchOnMount: 'always',
+    refetchOnWindowFocus: 'always',
```
pgetta: `gcTime: 0`, `staleTime: 0`, and both refetch options set to `'always'` mean group membership is refetched on every mount and on every window focus, indefinitely, for a nav item. Role data changes rarely; a normal `staleTime` should be sufficient. Is there a specific reason for the aggressive settings?

src/ui/segments/project/seat-recoverability.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <span className="text-xs text-gray-600">(Seat has been vacated before)</span>
+      </div>
+    );
+  }
+
+  if (isPending) {
+    return <Skeleton className="h-4 w-24" />;
+  }
+
+  const creditThreshold = (course.credits_per_seat ?? 50) - 50;
```
pgetta: The displayed reason hardcodes "more than 50 credits" while the threshold is derived from `credits_per_seat`. Also, when `credits_per_seat` is undefined the threshold becomes 0, making the seat effectively always recoverable unless the balance is negative. Please confirm the `?? 50` default matches the business rule, and derive the message from the same values.

src/ui/segments/project/course-assign-seats-modal.tsx
```diff
@@ -0,0 +1,307 @@
...
+      setFile(selectedFile);
+      setError('');
+      setFilteredStudents([]);
+      setExistingStudents([]);
+      assignMutation.reset();
+
+      try {
+        const content = await selectedFile.text();
+        const students = parseCSV(content);
+        const uniqueInFile = Array.from(new Map(students.map((s) => [s.student_id, s])).values());
```
pgetta: Deduplication is by `student_id` only; two rows with the same email but different ids pass this filter and will only fail server-side. Separately, the duplicate notice below is stored in the `error` state and rendered in the red error box although the assignment proceeds; a distinct warning style would avoid signalling a failure.

src/ui/segments/project/drop-seat-button.tsx
```diff
@@ -0,0 +1,106 @@
...
+            setShowConfirm(false);
+          }
+        }}
+        title="Drop seat"
+        size="sm"
+        footer={
+          <div className="flex justify-end gap-3">
+            <Button
+              variant="outline"
+              onClick={() => setShowConfirm(false)}
```
pgetta: This cancel button does not call `mutation.reset()` (the modal `onClose` handler does), so reopening the dialog after a failed drop shows the stale error message.

src/ui/segments/contribute/shared/components/submit-button.tsx
```diff
@@ -1,7 +1,7 @@
 'use client';
 
-import { useRouter } from '@bprogress/next';
 import { Form } from 'antd';
+import { useRouter } from 'next/navigation';
```
pgetta: Switching from `@bprogress/next` to `next/navigation` drops the progress-bar integration used across the app for this navigation. If this is intentional (e.g. it fixes an issue), a short note would help; otherwise it should probably stay on the wrapped router.

src/api/virtual-lab-svc/queries/types.ts
```diff
@@ -101,6 +101,16 @@ export interface IProjectExpandedResponse extends IProject, IProjectExtra {
   virtual_lab: TVirtualLab | null;
 }
 
+export type Course = {
```
pgetta: `Course` is now defined both here and in `queries/course.ts` (which additionally has `credits_per_seat`). The two are used interchangeably and are only compatible through structural typing; please consolidate into a single definition.

src/ui/segments/contribute/shared/components/contribution-form.tsx
```diff
@@ -88,8 +92,9 @@ function FormContent<
       await form.validateFields();
       setIsSubmitting(true);
       const values = form.getFieldsValue(true) as TFormValues;
-      const id = await createEntity({ values });
-      setCreatedEntityId(id);
+      const entity = await createEntity({ values });
+      setCreatedEntityId(entity.id);
+      onCreateSuccess?.(entity);
```
pgetta: The async callback is invoked without `await` or error handling. The current callers catch internally, but any future callback that throws becomes an unhandled rejection. Consider catching here, or documenting that `onCreateSuccess` must not throw.

src/features/notebooks/components/notebooks-layout.tsx
```diff
@@ -96,6 +100,27 @@ export function NotebooksLayout({ children }: Props) {
...
+        return;
+      }
+      try {
+        const projectIds = (await listAllProjectIds(virtualLabId)).filter((id) => id !== projectId);
```
pgetta: This syncs to every other project in the virtual lab, assuming all non-template projects are student projects. If a course lab can contain other projects, notebooks will be copied there as well. If enrolment project ids are available, targeting those explicitly would be safer.

src/services/notebooks/index.ts
pgetta: Just to note - in addition to the above, migrating this logic to the server is still a major improvement (but as we discussed - maybe later if the time is tight)

pgetta:
There are bunch of comments that I've posted, but please see what actually makes sense to address.

src/components/activate-enrolments.tsx
g-bar: Flow is like this
- Claim after start_date call activate immediatly
- Claim before start_date, can't activate yet course hasn't started.

The on mount usecase is for when activating previously claimed course, the double call in strict mode is not a concern, the second call is a no-op.

src/ui/segments/project/course-assign-seats-modal.tsx
g-bar: That's fine IMO, the BE doesn't assign at all if there are duplicate emails or ids, this one is just so that we filter out the already existing ids in the file.

src/components/activate-enrolments.tsx
pgetta: I just realised that this will be called for *all* the platform pages because this component is being used in the providers file. Seems a bit excessive. I know it's only a small request, but for all pages and all users - when only a very small subset will actually require that.

Does it make sense to limit the scope to some pages where the course perhaps is being used/launched?

g-bar: Optimizing out this single no-op call will yield no measurable performance gain. Given the existing volume of global requests (session, groups, vlab, entitycore), the overhead of this addition is statistically insignificant, not excessive, IMHO.

Also there is no way to know a priori if the user has courses to activate and to find out we have to make a call to virtual lab api any way. 

Maybe just the home page, I think that's a reasonable compromise?

src/features/notebooks/components/notebooks-layout.tsx
g-bar: Done

src/ui/segments/contribute/shared/components/contribution-form.tsx
g-bar: Done

src/components/activate-enrolments.tsx
pgetta: We've discussed (in Teams) moving the logic to the vlab/project layout and triggering requests only for course vlabs. Potentially to be addressed later given the tight deadline.

src/services/notebooks/index.ts
g-bar: This one I deleted as it had duplicated logic see below.

src/services/notebooks/sync-template-notebooks.ts
g-bar: Yep, good catch

The simplest way to "batch" was to simply process one notebook at a time.

The calls for the projects for each notebook still happen in a Promise.all, there is no use to report per project failures since there is no mechanism to attempt to sync to some student projects, this would require some specialised "sync" UI that shows which projects are in sync and which aren't, not in scope for now and probably needs some entitycore changes. So if only one project fails we need to attempt the sync to all, but at least it's per notebook not for all notebooks

The new function avoids duplicated notebooks from prev runs since it first checks if a notebook of that name already exists in the project.

I report failures for each notebook that failed, and there is a "sync" notebook in the notebooks UI, so that they can retry those that failed manually.

src/features/notebooks/components/mini-detail-actions.tsx
g-bar: Using buildQueryKey directly won't work here because TanStack Query's queryKey-based invalidation does exact structural match on each array element, since the key is a single-element tuple [{ workspace, context, queryParameters, ... }], you'd have to match every field exactly, which defeats the purpose of invalidating all queries for a given entity type regardless of filters/pagination.

The predicate is the correct pattern for this case. Building the predicate from the builder would need passing a bunch of irrelevant (dummy) data that's unused here to the builder (like brain region, species).

What I've done instead is replace the inline ad-hoc type cast with the ExtendedEntityTypeQueryKey type from the builder module, so the predicate is at least structurally coupled to buildQueryKey's output

src/ui/segments/project/seat-recoverability.tsx
g-bar: Changed the message. 

`credits_per_seat` can never be undefined, that was a typing error.

src/ui/segments/project/course-assign-seats-modal.tsx
g-bar: I'm now checking duplicate emails too.
Error is correct backend doesn't proceed to assign anything if there are duplicates.

.github/workflows/deploy-preview.yml
g-bar: changed to 15 minutes

g-bar:
I addressed the rest of the feedback in this other PR to avoid accidentally introducing regressions to this preview env shoul Natali need it:

https://github.com/openbraininstitute/core-web-app/pull/1819

.github/workflows/deploy-preview.yml
pgetta: Did you push your changes? The diff still shows 20 minutes.

src/api/entitycore/queries/assets/index.ts
```diff
@@ -50,8 +50,8 @@ export async function getAssets({
   entityId: string;
   ctx?: WorkspaceContext;
 }): Promise<EntityCoreResponse<IAsset>> {
-  const api = await authApiClient(config.ENTITY_CORE_URL);
-  return await api.get<EntityCoreResponse<IAsset>>(entityAssetsPath(entityType, entityId), {
+  const api = await entityCoreApi();
+  return await api.get<EntityCoreResponse<IAsset>>(`/${kebabCase(entityType)}/${entityId}/assets`, {
```
bilalesi, line 54: `entityAssetsPath` do the same thing as the change

src/api/entitycore/queries/assets/index.ts
```diff
@@ -223,8 +223,8 @@ export async function createJsonAsset({
   if (label) formData.append('label', label);
   if (meta) formData.append('meta', JSON.stringify(meta));
 
-  const api = await authApiClient(config.ENTITY_CORE_URL);
-  return await api.post<IAsset>(entityAssetsPath(entityType, entityId), {
+  const api = await entityCoreApi();
+  return await api.post<IAsset>(`/${kebabCase(entityType)}/${entityId}/assets`, {
```
bilalesi, line 227: same here

bilalesi (approved):

src/api/entitycore/queries/assets/index.ts
bilalesi: can be done later

PR #1561, opened 2026-04-02
Download previously selected entities (unfiltered)

g-bar:


pgetta:
> @pgetta I'm getting this error in the deploy job for the preview env, any idea what might be wrong?
> 
> ```
> aws: [ERROR]: An error occurred (NotFoundException) when calling the ListJobs operation: Branch download-select not found.
> ```

Apparently we've hit the "Branches per app" quota in AWS Amplify which is [50 and unfortunately not adjustable](https://docs.aws.amazon.com/amplify/latest/userguide/quotas-chapter.html).

We need to clean up some branches from the repo to make it operational again. Some manual interventions will be needed for this branch as well. I'll get back to you.

bilalesi (dismissed):

bilalesi (approved):

PR #1570, opened 2026-04-14
feat-144: filter by `generation_type` and `protocol_design` for `cell-morphology`

bilalesi:


tolokoban (approved):
That's cool to have added tests.

PR #1575, opened 2026-04-20
fix loading state stale for counting entities

bilalesi:


tolokoban (approved):

bilalesi:
@tolokoban thanks for both !

PR #1578, opened 2026-04-21
Prod build emodel/131/spike activity 2

tolokoban:
# Use new data_type attribute in simulation stream messages

## Summary

Refactors the single-neuron simulation runner to support the new `data_type` field on streamed simulation messages, enabling differentiation between trace and spikes data types.

## Changes

- `src/services/small-scale-simulator/types.ts`: Added optional `data_type` field to DataMessage type to support categorizing incoming stream data (e.g. trace, spikes).
- `src/ui/segments/workflows/simulate/single-neuron/shared/runner.tsx`:
  - Extracted the NDJSON message parsing callback into a standalone makeMessageParser function.
  - Extracted stream data appending logic into makeAppendStreadData.
  - Extracted initial plot data setup into setInitialPlotData.
  - Extracted argument validation into checkArguments.
  - Added hasCause type guard for safer error handling (replaces error.cause access on any-typed catch variable).

Message parser now inspects `data_type` to distinguish between trace and spikes messages (spikes handling is a TODO for a follow-up PR, see phase 2 of [this ticket](https://github.com/openbraininstitute/prod-build-emodel/issues/131)).

## Motivation

Prepares the codebase for handling spike activity data from the small-scale simulator.

We can now merge the [PR for small-scale-simulator](https://github.com/openbraininstitute/Bluenaas/pull/141). The frontend will be compatible with the new format.

Improves readability and maintainability by breaking a ~200-line inline callback into discrete, testable functions.

Removes usage of any in the catch clause for type-safe error handling.

### DIFF
--- a/src/services/small-scale-simulator/types.ts
+++ b/src/services/small-scale-simulator/types.ts
@@ -11,22 +11,22 @@ export enum MessageType {
   DATA = 'data',
 }
 
-export type Message<T> = StatusMessage | DataMessage<T>
+export type Message<T> = StatusMessage | DataMessage<T>;
 
 type MessageCtx = {
-  [key: string]: any
-}
+  [key: string]: any;
+};
 
 type StatusMessage = {
-  message_type: MessageType.STATUS
-  status: JobStatus
-  extra?: string
-  ctx?: MessageCtx
-}
+  message_type: MessageType.STATUS;
+  status: JobStatus;
+  extra?: string;
+  ctx?: MessageCtx;
+};
 
 type DataMessage<T> = {
-  message_type: MessageType.DATA
-  data_type?: string
-  data: T
-  ctx?: MessageCtx
-}
+  message_type: MessageType.DATA;
+  data_type?: string;
+  data: T;
+  ctx?: MessageCtx;
+};
--- a/src/services/small-scale-simulator/types.ts
+++ b/src/services/small-scale-simulator/types.ts
@@ -11,21 +11,22 @@ export enum MessageType {
   DATA = 'data',
 }
 
-export type Message<T> = StatusMessage | DataMessage<T>;
+export type Message<T> = StatusMessage | DataMessage<T>
 
 type MessageCtx = {
-  [key: string]: any;
-};
+  [key: string]: any
+}
 
 type StatusMessage = {
-  message_type: MessageType.STATUS;
-  status: JobStatus;
-  extra?: string;
-  ctx?: MessageCtx;
-};
+  message_type: MessageType.STATUS
+  status: JobStatus
+  extra?: string
+  ctx?: MessageCtx
+}
 
 type DataMessage<T> = {
-  message_type: MessageType.DATA;
-  data: T;
-  ctx?: MessageCtx;
-};
+  message_type: MessageType.DATA
+  data_type?: string
+  data: T
+  ctx?: MessageCtx
+}
--- a/src/ui/segments/workflows/simulate/single-neuron/shared/runner.tsx
+++ b/src/ui/segments/workflows/simulate/single-neuron/shared/runner.tsx
@@ -1,6 +1,7 @@
-'use client';
+/** biome-ignore-all lint/performance/noAccumulatingSpread: useless rule */
+'use client'
 
-import { captureException } from '@sentry/nextjs';
+import { captureException } from '@sentry/nextjs'
 import {
   delay,
   isNil,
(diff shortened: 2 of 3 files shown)


src/ui/segments/workflows/simulate/single-neuron/shared/runner.tsx
```diff
@@ -496,54 +420,261 @@ export const launchSimulationAtom = atom<
...
+    weight_scalar: number;
+    color?: string | undefined;
+  }[],
+  plotDataAtom: PlotDataAtom,
+  stimulusGraphResult: PlotData
+): ((data: Message<SimulationStreamData>) => void) | undefined {
+  const appendStreamData = makeAppendStreadData(get, set, plotDataAtom);
+  return (message) => {
+    match(message)
+      .with({ message_type: MessageType.DATA }, (msg) => {
```
pgetta, line 517: With this `match` function you can provide both message_type and data_type at the same time, so that there is no need to have nested comparison.

Something ilke this:
```
match(msg)
  .with(){ message_type: MessageType.DATA, data_type: 'trace' }, () => {})
  .with(){ message_type: MessageType.DATA, data_type: 'spikes' }, () => {})
```

Proabbly worth to create some kind of enum with those `trace` and `spikes` values not to have hardcoded values.

tolokoban: It seems a bit more complicated, because I have actually three cases. Can we do some OR logic with this `with()`?

```ts
match(msg)
  .with(){ message_type: MessageType.DATA, data_type: 'spikes' }, () => {
    processSpikes()
  })
  .with(){ message_type: MessageType.DATA, data_type: 'trace' }, () => {
    processTrace()
  })
  .with(){ message_type: MessageType.DATA, data_type: undefined }, () => {
    processTrace()
  })
```

tolokoban: Another point for the "nested conditions" is that one of them will disappear once the backend will go in production.

pgetta: > It seems a bit more complicated, because I have actually three cases. Can we do some OR logic with this `with()`?
> 
> match(msg)
>   .with(){ message_type: MessageType.DATA, data_type: 'spikes' }, () => {
>     processSpikes()
>   })
>   .with(){ message_type: MessageType.DATA, data_type: 'trace' }, () => {
>     processTrace()
>   })
>   .with(){ message_type: MessageType.DATA, data_type: undefined }, () => {
>     processTrace()
>   })

Yeah, that will work. But that's up to you, really minor thing, especially given it's going to be cleaned up.

pgetta (dismissed):

pgetta:
@tolokoban please ping me once you resolve the merge conflict - will re-approve.

src/ui/segments/workflows/simulate/single-neuron/shared/runner.tsx
tolokoban: You are right, in the cleanup process, we won't have any nesting of conditions.

tolokoban:
@pgetta I have fixed the merge conflicts.

src/services/small-scale-simulator/types.ts
```diff
@@ -11,21 +11,22 @@ export enum MessageType {
   DATA = 'data',
 }
 
-export type Message<T> = StatusMessage | DataMessage<T>;
+export type Message<T> = StatusMessage | DataMessage<T>
```
pgetta: Just noticed, where are these formatting changes coming from, perhaps you have some conflicting formatter? From what I remember Biome is configured to always use semicolons.

tolokoban: Yes, I have seen a lot of this semi colons during merge conflict resolution.
I noticed that, sometimes, Biome is failing and VSCode switches to anoher linter.
I have to execute "Biome: Restart" to fix that.

I have fixed this semi colon in this file, but I don't dare executing the formatter on the whole code, unless we get too many changes.

pgetta: Could you also run format against `src/ui/segments/workflows/simulate/single-neuron/shared/runner.tsx`? That one has been reformatted (w/o semi colons) as well.

bilalesi: may it can help, you can run this snippet before any commit 
It will only format/lint the unstaged files
```ts
./node_modules/.bin/biome check --write \
$(git diff --name-only) \
$(git ls-files --others --exclude-standard)
```

pgetta: Thank you Fabien, that's way better.

pgetta (dismissed):

src/services/small-scale-simulator/types.ts
tolokoban: Thanks for the trick @bilalesi , that will come in handy.

In the meantime, I have updated the Lefthook config to force the formatting off commited files.
It seems to work.

tolokoban: I have projects at home that use another linter.
And, sadly, VSCode cannot define a default formatter per project.
So, I guess, I have changed the formatter at some point, and since VSCode is synchronised, I got the config on my Mac too.

Anyway, thanks to Bilal's snippet and the Lefthook config, I hope I will avoid this pitfall next time.

pgetta (approved):

PR #1581, opened 2026-04-21
Refactor landing pages

pgetta:
## Summary

Refactor landing pages to leverage Next.js server components for faster initial content delivery, enforce kebab-case file naming, and consolidate components into `src/ui/segments/landing/`.

## Landing pages refactor

- **Client → Server components**: `LandingPage`, `HeroServer`, `FooterServer`, `SectionRouter`, and `SectionGeneric` are now async server components that fetch Sanity data at render time — eliminates the loading delay before CMS content appears
- **Relocated** entire `src/components/LandingPage/` → `src/ui/segments/landing/` to match project conventions
- **File naming**: renamed PascalCase/mixed-case files to kebab-case throughout
- **Removed ~80 redundant barrel `index.ts`** files that only re-exported a single default
- **Extracted 16 new Sanity API functions** into `src/services/sanity/api/` (e.g. `get-news-list`, `get-hero-content`, `get-foundations-content`) — data fetching is now co-located with the service layer instead of scattered across widget components
- **Composable page shell**: split monolithic `LandingPage` into `LandingPageShell` (client, scroll + error dialog) and `LandingPage` (server, orchestrates Menu → Hero → Section → Footer)
- **SectionRouter**: server component that routes by `EnumSection` and pre-fetches section-specific data (e.g. news list + count for the News section)
- **SFN 2025 page**: converted from `'use client'` to a server component using `LandingPageShell`
- **News detail page** (`/news/[slug]`): data is now fetched server-side via `getNewsItem()`, falls back to the news listing if the item is not found

## Other cleanup

- Removed dead code: unused `useFullHeight` hook (moved to `src/utils/`), `social-media-links/hooks.ts` (replaced by `get-social-media` API), announcement popup assets, `VerticalRuler`, `select-currency` styles
- Deduplicated GROQ projections that were previously copy-pasted across content files
- Added `.playwright-mcp` to `.gitignore`

## Performance gains

Lighthouse shows performance change on `/` (in comparison to the main.preview deployment): 74 -> 90

Before:
image

After:
image

**P.S.** I've run Lighthouse perf tests couple of times more and the difference isn't that dramatic, but still very good.
The performance for the rest of the pages was not measured, but I expect similar gains.


tolokoban:
> Removed ~80 redundant barrel index.ts files that only re-exported a single default

Sometimes, these `index.ts` are a convention that help preventing the spaghetti nightmare.

There is no way in Javascript/Typescript to define a private module. So you can import any file from anywhere.
But it can be inconvenient when you want modules that should be private to other modules.

So, the convention that says "it is forbidden to import a module if the path crosses a folder with an index.ts", can be checked by automated scripts and could help reducing unwanted dependencies to private modules.

Most of the time, we don't use a private module on purpose, but just by using the auto-completion that leads to a file with the same name, for instance.

That's said, I don't know how many of these 80 folders used this convention.

tolokoban (dismissed):

pgetta:
Thank you @tolokoban for the discussion, agreed to merge this for now and look into enforcing module boundaries by using [noPrivateImports](https://biomejs.dev/linter/rules/no-private-imports/) rule from Biome - if it satisfies our needs, to be confirmed.

pgetta:
@loris-olivier-obi I also wanted you to be aware of this change. Maybe you have any comments?

pgetta:
@tolokoban I've checked the noPrivateImports rule and indeed it can be used for our purposes.

I've created a PR to enable it on the error level: https://github.com/openbraininstitute/core-web-app/pull/1585

loris-olivier-obi (dismissed):
LGTM

pgetta:
Relates to [Hero text flicker, layout shift and lost scroll position on landing pages](https://github.com/openbraininstitute/core-web-app/issues/1637).

loris-olivier-obi (approved):
LGTM

PR #1595, opened 2026-04-27
Handle calibration absence in ME-model

pgetta:
The `calibration_result` is normally registered with a short delay after an ME-model has been created, within that time window accessing `model.calibration_result.rin` throws an error due to `model.calibration_result` being `null`.

This change ensures the hook returns `RIN` as `undefined` while the calibration hasn't been registered.


pgetta:
Relates to [[Simulate new me-model] Not possible to simulate freshly built me-model](https://github.com/openbraininstitute/prod-singlecell-simulation/issues/268).

bilalesi (approved):

pgetta:
Thank you Bilal

PR #1597, opened 2026-04-27
Fix neurite show/hide issue

tolokoban:
# Fix missing colors & dendrite show/hide regression

## Summary

Fixes a regression where dendrite colors were missing and the show/hide dendrites functionality was broken in the MorphoViewer.

### MorphoViewer colors & dark mode

- Removed the `mode` prop from `MorphoViewerComponent`; background color is now derived from the `isDarkMode` setting hook instead of being passed externally.
- Added a fallback default (`?? false`) for `darkMode` in `useMorphoViewerSettings` to prevent undefined state.

### ColorsLegend reactivity fix

- Introduced a `useLabels` hook that listens to `eventColorsChange` so the legend re-renders when morphology parts (soma, dendrites, axon) become available or are toggled.
- Previously, `getProperLabels` was called inline during render without reacting to painter events, causing stale labels.

### Import cleanup

- Reorganized imports across MorphoViewer files to follow project conventions (packages → `@/` aliases → relative → type imports).
- Switched `MorphologyCanvas` and `ColoringType` to type-only imports where applicable.

### Dependency updates

- Upgraded `@bbp/morphoviewer` from `0.24.13` → `0.24.15`.
- Upgraded Biome schema from `2.3.10` → `2.4.13`.


src/components/MorphoViewer/hooks/settings.ts
```diff
@@ -17,7 +17,9 @@ import {
   LIGHT_BASAL_DENDRITE,
   LIGHT_SOMA,
 } from '../constants';
-import { atomFamilyWithExpiration } from '@/util/atoms';
+
+/* eslint-disable no-param-reassign */
```
bilalesi: we are not using eslint anymore

package.json
```diff
@@ -132,6 +134,7 @@
     "@types/tar-stream": "^3.1.4",
     "@types/three": "^0.159.0",
     "autoprefixer": "^10.4.27",
+    "check-barrel-import-violation": "^0.3.0",
```
bilalesi, line 137: this is cool!

bilalesi (dismissed):

src/components/MorphoViewer/hooks/settings.ts
tolokoban: You are right!
I think it's the context menu "Fix" that has added this.
Let me remove it.
Good catch!

tolokoban: Sorry @bilalesi I just pushed the fix for this es-lint comment, but you already approved.
Can you do it again?

bilalesi (approved):

tolokoban:
Thanks @bilalesi !

PR #1598, opened 2026-04-27
feat: refactor listing tasks with status use on-demand fetching

bilalesi:
we noticed a significant slowdown when fetching data while opening a simulation type from either the listing page or the details page this mainly comes from the public simulations uploaded by Armando, where a large amount of configuration-related data is being fetched upfront, 
and I do the following: 
- grouping extraction/skeletonization (only shared part) together which basically will be the same for all future task-config/activity 

- refactors the scan-config “results” screens so the lists are fully rendered  and each row/card owns its own status loading

**Changes:**
- **simulations**: `src/features/scan-config/use-cases/simulations/results.tsx` now **lists all simulations** at once and each **simulation card polls/loads its own status** 
- **extraction + skeletonization**: `src/features/scan-config/use-cases/extraction/results.tsx` and `.../skeletonization/results.tsx` now **load all task configs** up front (using the task runner hook with `loadExecutions: false`) and each **card loads its own latest execution/status**
  - add task namespace that will be used across all new task (it contains shared methods, `Task.many`, `Task.one`, `Task.status`, `Task.count`)

  - added a **single-id simulation status helper** (`statusById`) so row-level queries don’t rely on array-of-ids fetching
  - added/used a **per-config execution hook** so each card can poll only when needed (pending/running)

- **ui deduplication**: 
    - refactor ui shared components, we had many issues as we were duplicating `extraction/simulations/skeletonization/build (coming soon)` this also prepare for migrating to the task config/activity
   - extracted the “select all” checkbox into a shared component and reused it in both the **task config** list and the **simulation** list, 
   - results layout 
   - in/output layout
   - status badge
   - listing page status column


tolokoban (approved):

bilalesi:
**NOTE:** 
I will have 2 PRs for 
- currently we are having 2 big loading state, one to load the circuit (entity), campaign (if passed in query param)
and another one in the `template.tsx` to fetch the schema, entity, ... which leads to have two loading-skeleton 
moving all the data fetching (only for lightweight requests) to the server will result only one skeleton and also some speed
- secondly when opening the scan-config details page from listing page, it resolve the full campaign (campaign, config, base-entity, configs) which is not necessary needed for example configs can be async until the user enter the results tab

fyi: @pgetta @g-bar

bilalesi:
thanks @tolokoban

PR #1599, opened 2026-04-28
(DO_NOT_MERGE) feat: support multi-reference types in scan-config block references

bilalesi:
- rename reference_type (string) to reference_types (Array<string>) in types
- update `useReferenceTypeDict` to build many-to-many lookup map
- rewrite `Reference` dropdown to collect options from all matching dictionaries
- fix `block.tsx` visibility gate to check any reference type has a default label
- skip fields without ui_element (e.g. population with type: null)
- fix `EntityPropertyDropdown` onChange writing to wrong key (node_set -> [k])
- add single-select mode to `EntityPropertyDropdown` for string-typed fields

### DIFF
--- a/src/features/scan-config/components/ui-blocks/block.tsx
+++ b/src/features/scan-config/components/ui-blocks/block.tsx
@@ -94,7 +94,9 @@ export default function Block({
                 !isType(paramSchema) &&
                 !paramSchema.ui_hidden &&
                 (paramSchema.ui_element !== ScanConfigUIElementDict.Reference ||
-                  Boolean(schema.default_block_reference_labels?.[paramSchema.reference_type]))
+                  paramSchema.reference_types.some(
+                    (refType) => !!schema.default_block_reference_labels?.[refType]
+                  ))
             )
             .map(([k, blockElementSchema]) => {
               if (isType(blockElementSchema)) return null;
--- a/src/features/scan-config/components/ui-elements/index.tsx
+++ b/src/features/scan-config/components/ui-elements/index.tsx
@@ -194,24 +194,44 @@ export function UIElementRender({
         entity: P.nonNullable,
       },
       ({ paramSchema }) => {
-        const getValue = (): string[] => {
-          if (Array.isArray(value) && value.every((v) => typeof v === 'string')) {
-            return value;
-          }
-          if (typeof value === 'string') return [value];
-          return [];
-        };
+        // detect if the field supports multiple values by checking for anyOf with an array type
+        const isMultiple = 'anyOf' in paramSchema;
+
+        if (isMultiple) {
+          const getValue = (): Array<string> => {
+            if (Array.isArray(value) && value.every((v) => typeof v === 'string')) {
+              return value;
+            }
+            if (typeof value === 'string') return [value];
+            return [];
+          };
+
+          return (
+            <EntityPropertyDropdown
+              multiple
+              disabled={disabled}
+              schemaMappingConfig={schemaMappingConfig}
+              value={getValue()}
+              onChange={(newV: string | Array<string>) => setState({ ...state, [k]: newV })}
+              property={paramSchema.property}
+            />
+          );
+        }
+
+        // single-value mode (e.g. population field with type: "string")
+        const singleValue = typeof value === 'string' ? value : undefined;
 
         return (
           <EntityPropertyDropdown
-            schemaMappingConfig={schemaMappingConfig}
+            multiple={false}
             disabled={disabled}
-            value={getValue()}
-            onChange={(newV: string[]) =>
+            schemaMappingConfig={schemaMappingConfig}
+            value={singleValue ?? ''}
+            onChange={(newV: string | string[]) =>
               setState({
                 ...state,
                 // NOTE: this is requested by James for IT'IS collaboration
-                node_set: Array.isArray(newV) && newV.length === 1 ? newV[0] : newV,
+                [k]: Array.isArray(newV) && newV.length === 1 ? newV[0] : newV,
               })
             }
             property={paramSchema.property}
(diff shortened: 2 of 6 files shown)


src/features/scan-config/components/ui-blocks/block.tsx
```diff
@@ -93,16 +87,18 @@ export default function Block({
               ([_, paramSchema]) =>
                 !isType(paramSchema) &&
                 !paramSchema.ui_hidden &&
+                'ui_element' in paramSchema &&
```
g-bar: Why? AFAIK all params should have an ui_element as per the spec. If a param doesn't it's not spec compliant and shoudl be added.

Or did that change?

src/features/scan-config/components/ui-elements/index.tsx
```diff
@@ -186,20 +186,40 @@ export function UIElementRender({
...
-          if (Array.isArray(value) && value.every((v) => typeof v === 'string')) {
-            return value;
-          }
-          if (typeof value === 'string') return [value];
-          return [];
-        };
+        // detect if the field supports multiple values by checking for anyOf with an array type
+        const isMultiple = 'anyOf' in paramSchema;
+
+        if (isMultiple) {
```
g-bar, line 200: IMHO it would be simpler to have a separate `ui_element` for the multiple case to comply with the strict non-conditional schema in the spec for each `ui_element`.

thoughts @chr-pok @james-isbister

james-isbister: So the idea was to specify the "reference_types" as a list of strings in the json_extra_schema for all references parameters (both when there is a single reference or multiple references are supported), as in this [example](https://github.com/openbraininstitute/obi-one/blob/164ee1592552fa80a53390663f919fb9faf4cf39/obi_one/scientific/blocks/stimuli/stimulus.py#L64). 

So even the single case is a special case of the multi case, so maybe there is no need to use conditional logic in the frontend? I see that the "anyOf" is being accessed though, but maybe the types can be taken from the reference_types list? 

Alternatively, it looks like the typing of all references variable is always `BlockReference(s) | None` so I think the anyOf should always be present for the single and multi case if its better to use the types rather than the strings in reference_types

g-bar: Yes, makes sense. If it's true that `anyOf` is always present then this logic will fail @bilalesi .

bilalesi: this is basically handle this case: 
```json
"population": {
    "type": "string",
    "minLength": 1,
    "title": "Population",
    "description": "Name of the biophysical node population to select from.",
    "property": "BiophysicalNeuronalPopulation",
    "property_group": "Circuit",
    "ui_element": "entity_property_dropdown"
},
```
in other cases we have : 
```json
"node_set": {
    "anyOf": [
        {
            "type": "string",
            "minLength": 1
        },
        {
            "items": {
                "type": "string",
                "minLength": 1
            },
            "type": "array"
        }
    ],
    "title": "Node Set",
    "description": "Name of the node set to use.",
    "property": "NodeSet",
    "property_group": "Circuit",
    "ui_element": "entity_property_dropdown"
}
```

so i think the schema need to change @james-isbister ?

src/features/scan-config/components/ui-blocks/block.tsx
bilalesi: you are right, fixed it, i missed the `!isType` check

src/features/scan-config/components/ui-elements/index.tsx
g-bar: Ay, yes I think we briefly discussed this in the past @james-isbister 

Basically the list was meant for 'parameter sweep' but that wasn't supported on the FE for this ui_element, so I asked for it's removal, we decided to keep it but ignore it on the FE side.

But then with the introduction of the AI feature, since it uses the schema to check if the configs it produces are valid it's possible for the ML to produce lists thereby breaking the FE.

We have a similar problem for the IDNeuronSet.

james-isbister: Ah sorry, in my last comment I thought we were looking at logic for the reference types - I didn't see this was for EntityPropertyDropdown.

Yes, I see and agree - we should make two different entity_property_dropdown ui_elements indeed to handle this case (basic dropdown - no scan) and the node_set selection (scan over selected dropdown elements). 

I can make an update in new_neuron_sets obi-one branch soon and we can update the frontend logic in this PR?

james-isbister: I'll remind myself of the IDNeuronSet at the same time

james-isbister: Have added entity_property_dropdown_sweep for the second case in the new_neuron_sets branch @bilalesi

bilalesi: thanks @james-isbister !
fyi: i could not find any use of this `entity_property_dropdown_sweep` in many different schemas, can you point me to one please ?

i've also noticed another ui_element named `neuron_property_filter`, what's this for ?, and should we  included it in the scope of this ticket?

james-isbister: Ah yes, I've added the old PredefinedNeuronSet. You should see it when you click "Add neuron set" now

The neuron_property_filter ui element is part of the neuron sets refactor but was assigned to @g-bar in another ticket: https://github.com/openbraininstitute/prod-circuit-simulation/issues/167

So maybe you want a single base branch for the neuron sets refactor

I'm about to update the property neuron set ticket now that I've updated the specs

bilalesi:
closed by this https://github.com/openbraininstitute/core-web-app/pull/1781

PR #1603, opened 2026-04-29
fix: wire brain-region simulation entities and preview updates

bilalesi:
- register brain-region view/domain mappings
- align simulate workflow config/types so region circuit simulations resolve correctly


tolokoban (approved):

bilalesi:
thanks @tolokoban

PR #1607, opened 2026-04-30
Make UX for spike viewer consistent with the rest of plots

pgetta:
## Summary

Aligns the spike-viewer raster plot's interaction model with the Plotly charts used elsewhere in the app, makes the Y-axis tick rendering more appropriate for neuron IDs, fixes inconsistent spike tick sizes caused by subpixel rasterization, and tunes point sizing + selection visuals.

## Interaction changes (`interaction.ts`, `raster-plot.tsx`)

Replaces the divergent gestures (scroll-zoom + drag-to-pan) with Plotly defaults:

- **Drag inside plot** → box-zoom (5px threshold).
- **Shift+drag inside plot** → pan.
- **Drag in x-tick band** → x-only zoom; **drag in y-tick band** → y-only zoom.
- **Scroll wheel** → no longer zooms.
- **Double-click** → resets axes: on plot resets both, on a single axis resets only that axis.
- Cursor reflects the region/modifier (`crosshair` / `grab` / `col-resize` / `row-resize`).
- Help text updated to `Drag to zoom · Shift+drag to pan · Double-click to reset`.

The selection rectangle is rendered as a DOM overlay: the active drag region shows as a light-bordered rect while everything outside is dimmed via a large `box-shadow` clipped by a wrapping mask (so axis labels stay legible). Opacity fades in/out (300ms) to avoid a hard pop on click. The WebGL/canvas render path is untouched.

## Axis changes (`axis-overlay.ts`)

- Y-axis labels are integers — `computeTicks` accepts an `integer` flag that floors the step to `Math.max(1, round(step))`, preventing fractional Neuron IDs and duplicate-rounded labels.
- Y-axis filters out negative ticks — neuron IDs are non-negative, so negative gridlines/labels would never be meaningful.
- Tick counts are dynamic from `plotRect`: ~80px per tick on X, ~30px per tick on Y (min 2). Larger viewports surface more IDs without crowding small ones.

## Point-size consistency fix (`webgl-points.ts`, `raster-renderer.ts`)

Spike ticks were rendering at visibly different widths/heights because `gl_POINTS` rasterization is sensitive to subpixel position and the fragment shader used a hard `discard` cutoff.

- `gl_PointSize` is rounded to an integer before being passed to the GPU, eliminating size jitter from fractional sizing.
- Fragment shader replaces the hard cutoff with a `smoothstep` alpha falloff. The feather sits *outside* the solid core (`smoothstep(0.2, 0.2 + feather, dx)`) so the body stays fully opaque and dense overlapping ticks read as solid colour rather than accumulating a translucent haze.
- Blend func switched to `blendFuncSeparate` so destination alpha accumulates toward 1.0 in dense regions — fixes pale-blue compositing over the white page when feathered edges overlap.
- `u_pointSize` uniform is qualified `mediump` in the vertex shader to match the fragment shader's precision.

## Dynamic point sizing (`raster-renderer.ts`)

`computePointSize` now multiplies a fixed `baseSize = 8` by a count-based factor and adds the existing zoom bonus, capped at 14 px. The factor scales the dot size down as the dataset grows so dense rasters stay readable while sparse ones get bigger ticks:

- 1.0× at ≤ 1k spikes → 0.5× at ≥ 100k spikes, log-interpolated between (count spans orders of magnitude, so log-space puts the visible transition where users feel it).
- Recalculates on population toggle — hiding a dense population enlarges the remaining ticks.

## Test plan

- [x] Drag inside plot draws rectangle with dimmed surround; release zooms; <5px drag does nothing.
- [x] Drag in x-tick band zooms X only; drag in y-tick band zooms Y only.
- [x] Shift+drag pans; scroll over plot does not zoom.
- [x] Double-click on plot resets both axes; on x-axis resets X only; on y-axis resets Y only.
- [x] Hover tooltip still works when not dragging.
- [x] Y-axis labels are non-negative integers at all zoom levels; tick density scales with window size.
- [x] Cross-check gestures against another Plotly chart (e.g. ephys-viewer) — feel matches.
- [x] Spike ticks render at uniform width/height across the plot at multiple zoom levels.
- [x] Dot size visibly shrinks for >100k-spike populations and grows for sparse ones; toggling a dense population off enlarges the rest.

### DIFF
--- a/src/features/spike-viewer/components/raster-plot.tsx
+++ b/src/features/spike-viewer/components/raster-plot.tsx
@@ -71,7 +71,7 @@ export default function RasterPlot({ data }: RasterPlotProps) {
           ))
         )}
         <span className="ml-auto text-xs text-gray-400">
-          Scroll to zoom · Drag to pan · Double-click to reset
+          Drag to zoom · Shift+drag to pan · Double-click to reset
         </span>
       </div>
       <div ref={containerRef} className="min-h-0 flex-1" />
--- a/src/features/spike-viewer/renderer/axis-overlay.ts
+++ b/src/features/spike-viewer/renderer/axis-overlay.ts
@@ -62,7 +62,7 @@ export class AxisOverlay {
     const xTickCount = Math.max(2, Math.floor(plotRect.width / 80));
     const yTickCount = Math.max(2, Math.floor(plotRect.height / 30));
     const xTicks = computeTicks(bounds.xMin, bounds.xMax, xTickCount);
-    const yTicks = computeTicks(bounds.yMin, bounds.yMax, yTickCount, true);
+    const yTicks = computeTicks(bounds.yMin, bounds.yMax, yTickCount, true).filter((v) => v >= 0);
 
     ctx.clearRect(0, 0, ctx.canvas.width, ctx.canvas.height);
     ctx.save();
--- a/src/features/spike-viewer/renderer/axis-overlay.ts
+++ b/src/features/spike-viewer/renderer/axis-overlay.ts
@@ -14,8 +14,8 @@ export type PlotRect = {
 
 export const MARGIN = { top: 20, right: 20, bottom: 60, left: 70 };
 
-/** Compute nice tick values for an axis range. */
-function computeTicks(min: number, max: number, targetCount: number): number[] {
+/** Compute nice tick values for an axis range. When integer is true, step is constrained to an integer >= 1. */
+function computeTicks(min: number, max: number, targetCount: number, integer = false): number[] {
   const range = max - min;
   if (range <= 0) return [min];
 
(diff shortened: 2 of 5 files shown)


pgetta:
Relates to [Consistent UX for spike viewer](https://github.com/openbraininstitute/prod-circuit-simulation/issues/161).

tolokoban (dismissed):

pgetta:
@tolokoban sorry I've pushed a small extra change, could you please re-approve?

tolokoban (dismissed):
No problem Pavlo.

pgetta:
@tolokoban I've asked for the review too quick. Again, sorry for the noise. I'll request another review once this is good with QA / PO.

tolokoban (approved):

PR #1611, opened 2026-04-30
feat: integrate realtime logs to circuit extraction activity

bilalesi:


g-bar (dismissed):

src/features/scan-config/use-cases/build/in-out-files.tsx
tolokoban: Here, you are looping on all files in `outputFiles` but only use the first one (`outputFiles[0]`). Is that on purpose?

bilalesi: yes, good catch, i fixed it

tolokoban (approved):

bilalesi:
thanks both !

PR #1612, opened 2026-04-30
fix em-synapse-mapping redirection after build

bilalesi:


tolokoban (approved):
LGTM. But you have an issue with the preview.

bilalesi:
thanks @tolokoban

PR #1619, opened 2026-05-01
Large neuronsets

g-bar:


src/features/scan-config/types.ts
```diff
@@ -29,13 +29,13 @@ interface Object {
   [key: string]: Primitive | Primitive[] | Object;
 }
 
-export type ConfigValue = Primitive | Primitive[] | Object;
+export type ConfigValue = Primitive | Primitive[] | Object | ConfigValue[];
 export type Config = Record<string, Object | string>;
 
 export interface AtomsMap {
   [key: string]:
-    | ReturnType<typeof atom<Record<string, ConfigValue | Array<ConfigValue>>>>
```
bilalesi, line 38: I know it's draft but just to let you know, 
this should accept Array<ConfigValue> too, it was added in the em-synapse-mapping schema

g-bar: Yes, this was causing a type error in the `<Block/>`component.

But this PR changes the ConfigValue to be 

```ts
export type ConfigValue = Primitive | Primitive[] | Object | ConfigValue[];
```
so it still works as an array, and fixed the other type error.

but i'll double check the types before making it ready for review, thanks!

src/features/scan-config/types.ts
```diff
@@ -29,13 +29,13 @@ interface Object {
   [key: string]: Primitive | Primitive[] | Object;
 }
 
-export type ConfigValue = Primitive | Primitive[] | Object;
+export type ConfigValue = Primitive | Primitive[] | Object | ConfigValue[];
```
g-bar: Here @bilalesi

bilalesi: missed this one, thanks !

tolokoban (dismissed):

src/services/ai-agent/hooks/chat.ts
```diff
@@ -4,6 +4,7 @@ import { useChat } from '@ai-sdk/react';
 import { useQueryClient } from '@tanstack/react-query';
 import { DefaultChatTransport, getToolName, isToolUIPart } from 'ai';
 import { atom, useAtom, useSetAtom, useStore } from 'jotai';
+import { isPlainObject } from 'node_modules/zod/v4/core/util.cjs';
```
tolokoban: Is it normal that you import from `node_modules`?

tolokoban: I see that we already have the function `isPlainObject` defined here: `@/features/scan-config/components/utils`.

src/features/scan-config/types.ts
```diff
@@ -15,28 +15,19 @@ import {
...
 import type { ActivityStatus } from '@/api/entitycore/types/shared/activity';
 import type { AssetContentType, IAsset } from '@/api/entitycore/types/shared/global';
 import type { Prettify } from '@/utils/type';
 
-export type SetAtom<Args extends unknown[], Result> = (...args: Args) => Result;
-
 type Primitive = null | boolean | number | string;
-interface Object {
-  [key: string]: Primitive | Primitive[] | Object;
+export interface Object {
```
tolokoban: Isn't this shadowing the global `Object` javascript type?

src/services/ai-agent/hooks/chat.ts
g-bar: Good catch, thanks, fixed.

src/features/scan-config/types.ts
g-bar: You're right, I renamed it to `ConfigObject`

tolokoban (dismissed):
Good job Gil!

package.json
```diff
@@ -27,6 +27,7 @@
     "@apache-arrow/es2015-esm": "^19.0.1",
     "@apidevtools/json-schema-ref-parser": "^13.0.5",
     "@bprogress/next": "^3.2.12",
+    "@monaco-editor/react": "^4.7.0",
```
bilalesi, line 31: are we using this ?
this is too heavy, and i do not see it used in this PR ?

package.json
```diff
@@ -137,6 +138,7 @@
     "check-barrel-import-violation": "^0.3.0",
     "knip": "^5.88.1",
     "lefthook": "^2.1.6",
+    "monaco-editor": "^0.55.1",
```
bilalesi, line 142: goes with the above

src/features/scan-config/index.tsx
```diff
@@ -154,6 +154,7 @@ export function ScanConfiguration({
   if (entity || usedType) {
     return (
       <ScanConfigTemplate
+        key={schemaName + entityId}
```
bilalesi: while I really like this approach, it means re-rendering the whole scan config, technically, that’s probably the cleanest and most correct solution, but another part of me thinks that resetting only the affected parts could give us a small performance win and avoid unnecessary re-renders.

src/features/scan-config/components/hooks/schema.ts
```diff
@@ -167,206 +131,126 @@ const ModelIdentifierSelector = {
...
+  initialConfig,
+  model,
+}: {
+  schema: ConfigSchema;
+  initialConfig?: Config;
+  model: TSupportedEntitiesForScanConfiguration | Nullish;
+}) {
+  const [configState, setConfigState] = useState<Config>(() =>
+    buildInitialConfigState(schema, initialConfig, model)
+  );
```
bilalesi, line 251: 👍

package.json
g-bar: It's used in the new ui element: 

https://github.com/openbraininstitute/core-web-app/blob/e171bf4e4ece122ae504cbba3a09c7c5de43bb0b/src/features/scan-config/components/ui-elements/neuron-ids.tsx#L479

src/features/scan-config/index.tsx
bilalesi: this is just a thought, no need to change anything

g-bar: I like this because it's very simple,
Imo the performance gains are almost negligible

The Middle and right panels always have to be re-rendered anyway when you change screen since we reset the config state.

The only thing you're not re-rendering is a few root keys in the left panel in case there are shared.

src/features/scan-config/components/ui-elements/neuron-ids/number-editor.tsx
```diff
@@ -0,0 +1,108 @@
+import Editor, { type OnChange, type OnMount } from '@monaco-editor/react';
+import dynamic from 'next/dynamic';
+import { useRef } from 'react';
+
```
bilalesi, line 4: we probably need to refactor this component a bit  and use it instead of code-blocks for `file-viewer` for `json` files, 
currently microcircuit/region simulations are freezing the ui when loading the `node_sets.json` file due large number of content. 
I just wanna mention it, this is out scope of this PR

bilalesi (approved):

src/features/scan-config/components/ui-elements/neuron-ids/number-editor.tsx
g-bar: I implemented truncating the file (just for display) in this PR:

Not as elegant but does the job for now, maybe we can indeed use sth like this on that panel too later.

Capture d’écran 2026-05-21 à 15 07 07
Capture d’écran 2026-05-21 à 15 07 19

g-bar:
Thanks @bilalesi @tolokoban for the review

PR #1624, opened 2026-05-05
Scan config action and form fixes

pgetta:
Fixes the **Duplicate** action on **Simulate → Single neuron (beta)** and **Simulate → Ion channel (beta)** campaigns, plus a latent re-render loop in scan-config form initialization.

  ## Changes

  - **`workflow-activity.tsx`** — add missing `return` after the `IonChannelModelSimulation` navigate so the catch-all `SimulationCampaign` branch no longer overrides it with `/configure/circuit/${entity_id}`.
  - **`memodel-circuit-simulation.ts` resolver** — replace `getCircuit` with `getMEModel` for the entity branch. The campaign’s `entity_id` points to a MEModel, so `GET /circuit/{id}` was 404-ing and crashing the resolver.
  - **`memodel/[id]/page.tsx`**:
    - import `resolveSimulationByCampaignId` from the memodel domain instead of `small-microcircuit-simulation` (stale copy/paste).
    - pass `populate: ['config']` (entity is already fetched via `getMEModel` on the page).
    - gate `<ScanConfiguration>` rendering on the campaign query resolving when `initialCampaignId` is present, mirroring the ion-channel page. Without this, the form mounted with `initialConfig=undefined`, and `useAtomsMap`’s init-once guard ignored the late-arriving config — the form stayed empty even though the
  campaign config loaded.
  - **`scan-config/components/hooks/schema.ts`** — in the `initialConfig` branch of `useAtomsMap`, backfill schema-defined root keys that `initialConfig` omitted, so `hasInitializedMapForSchema` becomes true on the next render. Prevents an effect re-fire loop ("Maximum update depth").

### DIFF
--- a/src/features/scan-config/components/hooks/schema.ts
+++ b/src/features/scan-config/components/hooks/schema.ts
@@ -215,6 +215,22 @@ export function useAtomsMap({
             map[k][subK] = atom<Record<string, ConfigValue | Array<ConfigValue>>>(subV);
           });
         });
+
+      // TODO: Consider implementing schema versioning
+      // Fill in any schema-defined root keys that initialConfig omitted, so the
+      // hasInitializedMapForSchema guard above passes on the next render. Without
+      // this, the effect re-fires every render and triggers Maximum update depth.
+      for (const [k, v] of Object.entries(schema.properties)) {
+        if (isType(v) || k in map) continue;
+        if (
+          v.ui_element === ScanConfigUIElementDict.BlockSingle ||
+          v.ui_element === ScanConfigUIElementDict.BlockUnion
+        ) {
+          map[k] = atom<Record<string, ConfigValue | Array<ConfigValue>>>({});
+        } else {
+          map[k] = {};
+        }
+      }
     } else {
       Object.entries(schema.properties).forEach(([k, v]) => {
         if (isType(v)) return;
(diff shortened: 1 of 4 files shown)


src/features/scan-config/components/hooks/schema.ts
```diff
@@ -215,6 +215,22 @@ export function useAtomsMap({
...
+      for (const [k, v] of Object.entries(schema.properties)) {
+        if (isType(v) || k in map) continue;
+        if (
+          v.ui_element === ScanConfigUIElementDict.BlockSingle ||
+          v.ui_element === ScanConfigUIElementDict.BlockUnion
+        ) {
+          map[k] = atom<Record<string, ConfigValue | Array<ConfigValue>>>({});
+        } else {
+          map[k] = {};
+        }
```
bilalesi, line 232: 👍

bilalesi (dismissed):

pgetta:
Relates to [[Staging][Chrome] “View configuration” and “Duplicate” actions fail or lose data in Simulate workflows](https://github.com/openbraininstitute/obi-one/issues/745).

src/features/scan-config/components/hooks/schema.ts
pgetta: Thanks for the clue Bilal. The way it was failing made it not very obvious.

bilalesi (approved):

PR #1625, opened 2026-05-05
Manage Amplify branches in GH actions

pgetta:
## Summary

Manage the AWS Amplify preview branch + subdomain lifecycle from GitHub Actions instead of relying on Amplify's auto-branch-creation. Auto-creation consumes the 50-branch app quota with non-PR branches; with ~25 open PRs we're already at the limit.

## Why disabling auto-branch-creation needs more than just `delete-branch`

Auto-branch-creation in Amplify also gates `enable_auto_sub_domain` — turning the first off disables the second. So the workflow has to own **both** lifecycles:

- **Branch lifecycle**: `create-branch` on PR open if missing, `delete-branch` on close.
- **Subdomain lifecycle**: read-modify-write on the app's `domain-association.subDomainSettings` to attach the prefix on deploy and detach it on close.

## Job structure

```
on PR open/sync/reopen:
  deploy ─┬─→ domain-attach   (cross-PR queue)
          └─→ monitor

on PR close:
  domain-detach ─→ cleanup    (detach queued cross-PR; delete-branch unconstrained)
```

Subdomain edits live in their own jobs — not inlined into `deploy`/`cleanup` — so the cross-PR mutex only covers the few seconds of domain editing rather than the multi-minute build path.

## Why a cross-PR concurrency group

`update-domain-association` is a *replace* operation against a **single shared resource** (one association per app+domain). Amplify does **not** appear to enforce optimistic concurrency on it, so concurrent edits silently last-write-wins and drop entries.

The reliable fix is GitHub-side serialization. Both `domain-attach` and `domain-detach` use:

\`\`\`yaml
concurrency:
  group: amplify-domain-\${{ vars.AMPLIFY_PREVIEW_APP_ID }}
  cancel-in-progress: false
\`\`\`

`cancel-in-progress: false` so PR B's edit *queues* behind PR A's instead of cancelling it — every PR must get its entry applied. The existing workflow-level \`preview-\${{ github.head_ref }}\` group still cancels superseded runs of the *same* PR.

## Other design choices

- `--stage PULL_REQUEST` on `create-branch` matches what auto-branch-creation produced for PR branches.
- `--no-enable-auto-build` since we explicitly invoke `start-job`; we don't want Amplify also auto-building on push.
- No `--environment-variables` on `create-branch`: app-level env vars apply automatically; only per-branch overrides need explicit values.
- Idempotency: attach/detach are no-ops if the entry is already in the desired state; `delete-branch` tolerates a missing branch via `|| echo`.
- Detach refuses to write an empty `subDomainSettings` list (Amplify rejects it, and it would orphan cert/DNS config).
- "Cancel running builds" remains *after* "Ensure Amplify branch exists" — `list-jobs` against a non-existent branch errors out.

## Required AWS-side changes (must land together with this PR)

The workflow alone won't fix the quota — these will have to be applied to the Amplify app + IAM role:

- [ ] Disable branch auto-creation on the Amplify app - **can be done after this PR is merged**
- [x] Extend the role behind `PREVIEW_DEPLOY_AWS_ROLE_ARN` with:
  - `amplify:GetBranch`, `amplify:CreateBranch`, `amplify:DeleteBranch`
  - `amplify:GetDomainAssociation`, `amplify:UpdateDomainAssociation`

## Test plan

- [ ] Open a fresh PR from a branch never seen by Amplify → `deploy` logs `Branch not found in Amplify, creating...`; `domain-attach` logs `Attached <prefix> -> <branch>`; `get-domain-association` shows the new prefix.
- [ ] After build completes → `https://<domain-safe-branch>.<PREVIEW_DOMAIN>` resolves and serves the preview.
- [ ] Push a follow-up commit → both `Branch already exists in Amplify` and `Subdomain <prefix> already attached` (idempotent).
- [ ] Close the PR → `domain-detach` logs `Detached <prefix>`; `cleanup` logs `Deleting Amplify branch`. Branch and prefix both gone in `list-branches` / `get-domain-association`.
- [ ] Reopen the PR → branch and subdomain recreated cleanly.
- [ ] (Optional) Open two PRs nearly simultaneously → in the Actions UI, one `domain-attach` runs while the other shows "Waiting for status" on the `amplify-domain-*` group, then runs sequentially. Both entries land in the final domain association (no lost writes).
- [ ] After a week of normal traffic, `aws amplify list-branches | jq '.branches | length'` tracks open-PR count, not lifetime branch count.

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -2,7 +2,7 @@ name: Preview
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, closed]
+    types: [ opened, synchronize, reopened, closed ]
 
 permissions:
   id-token: write
@@ -175,6 +175,8 @@ jobs:
           fi
 
           SHORTHAND=$(echo "$UPDATED" | jq -r '.[] | "prefix=\(.prefix),branchName=\(.branchName)"' | tr '\n' ' ')
+
+          echo $SHORTHAND
 
           aws amplify update-domain-association \
             --app-id "$AMPLIFY_APP_ID" \
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -139,6 +139,50 @@ jobs:
               });
             }
 
+  domain-attach:
+    needs: deploy
+    if: github.event.action != 'closed'
+    runs-on: ubuntu-latest
+    concurrency:
+      group: amplify-domain-${{ vars.AMPLIFY_PREVIEW_APP_ID }}
+      cancel-in-progress: false
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.PREVIEW_DEPLOY_AWS_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Attach preview subdomain
+        env:
+          BRANCH_NAME: ${{ needs.deploy.outputs.branch-name }}
+          PREFIX: ${{ needs.deploy.outputs.domain-safe-branch }}
+        run: |
+          set -euo pipefail
+
+          CURRENT=$(aws amplify get-domain-association \
+            --app-id "$AMPLIFY_APP_ID" \
+            --domain-name "$PREVIEW_DOMAIN" \
+            --query 'domainAssociation.subDomains[].subDomainSetting' \
+            --output json)
+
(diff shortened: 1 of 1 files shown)


pgetta:
Ok, I give up on this. Facing too many roadblocks: changing domain association fails with a very generic / vague message and there is no clear explanation in AWS Amplify documentation why. LLMs have no clue as well.

So far managed to delete some unused branches and brought down AWS Amplify branches to 35 which gives more 15 available. Will be cleaning up those once we are again close to the quota limit.


@danifr we can keep the current deployment role as is, even having those extra updates should be fine. Please let me know if you think otherwise.

pgetta:
FYI, @bilalesi, please see the comment above.

PR #1631, opened 2026-05-06
feat: implement multi-species 

bilalesi:
squash all current branch commits into a single commit for PR review


.env
```diff
@@ -1,16 +1,15 @@
...
-
 LEGACY_DEFAULT_CIRCUIT_ID="https://bbp.epfl.ch/data/bbp/mmb-point-neuron-framework-model/2b29d249-6520-4a98-9586-27ec7803aed2"
+
+APP_DEFAULT__BRAIN_REGION_HIERARCHY_ID="e3e70682-c209-4cac-a29f-6fbed82c07cd" # mouse
+MOUSE_DEFAULT__SELECTED_BRAIN_REGION_ID="5c60bf3e-5335-4971-a8ec-6597292452b2" # Cerebrum
+MOUSE_ATLAS__ID="d2fe7b47-257c-4068-9d15-ed12fd7d0510"
+HUMAN_DEFAULT__SELECTED_BRAIN_REGION_ID="125e1b98-96f1-4e60-bcb7-5eef1744f2da" # telencephalon
+RAT_DEFAULT__SELECTED_BRAIN_REGION_ID="a55e2b73-0ad1-49a0-848a-429c98f0ace4"
+
+EXCLUDED_HIERARCHY_IDS="e0e05eb2-421f-4133-a008-f63e7197e309"
```
pgetta, line 14: For later: this file should contain only default values that are common (e.g. valid for all: local dev, preview, staging and production).

I see that the same values are set in the `env.development` so probably this can be just cleaned up.

pgetta (approved):

.env
bilalesi: good catch!, will do after

PR #1638, opened 2026-05-07
Switch nav pills from Button to Badge component to prevent pill shift

danifr:
This is a small thing but I noticed that currently when you hover over the nav bar pills there is small shift in the components.

See the bug in action:
Image

Another issue I found is that some navbar icons get clipped at 100% size. 
Hard to notice but once you see it you cannot unsee it :) 

Screenshot 2026-05-07 at 10 57 50

This PR fixes both of those issues. All the code was done with Claude Sonnet


danifr:
This is how the fix looks like once deployed: 

Screenshot 2026-05-07 at 11 17 52


CC: @ayimaok

ayimaok:
~~I don't see it in my browser, will reach out in teams @danifr~~

danifr:
Addressed @ayimaok's feedback in https://github.com/openbraininstitute/core-web-app/pull/1638/changes/ae25e13d204dd87dc9896bfb11b0f752486320ed

This is the new behavior in action: 
Kapture 2026-05-07 at 11 49 03

ayimaok:
thanks a lot @danifr

pgetta:
Nice work, Daniel!

It seems the background and the spacing "label <-> icon" are different now, before:
image

After:
image

Is that expected?

danifr:
Ah true, I will make the background white again. Thanks!!

For the spacing it is a matter of personal taste but I think this new version uses the space better than the old one. 
Also another fix that this change includes is that when you zoom in the page at 150% the old version was deforming the nav bar, the new code works as expected.

danifr:
@pgetta shall we merge this one?

pgetta:
Ah, sorry, @danifr.

I've just checked and there is one small discrepancy:
An already active element (e.g. home page) on hover changes it's bg to white, while in staging it stays dark:

#### Current branch
image

#### Staging
image

Is that on purpose? I'll approve this PR in case it is, otherwise - please ping when to re-approve.

pgetta (approved):

danifr:
To me it looks nicer in the new implementation but I guess this is very personal :)

pgetta:
Yeah, I was not sure which one is better. Checked our [style guide](https://www.figma.com/design/akGPTH0WwNFDfSWSs3qAnh/OBI---UI-UX-2025?node-id=872-14278&t=wDKQhzwomxJXmnZz-4) in Figma: there is no design for active/hover combination. Checked few other websites and it's pretty common to explicitly highlight even the current/active menu item on hover.

So I presume it's good to go, thanks again for these fixes!

danifr:
Thanks, I will merge it as-is, but I'd be happy if someone else applies a more consistent style to all our site elements

ayimaok:
@danifr., this is in main.preview - looks good, is there a ticket for this (so that this can be moved it to staging)?

danifr:
hi @ayimaok I just created this one https://github.com/openbraininstitute/core-web-app/issues/1677 using the same description. We can continue there! thanks

PR #1640, opened 2026-05-07
fix when change schema, old selected block keep stored cause ui_element in undefined

bilalesi:
this is basically happen when the user in a specific scan configuration (let's say microcircuit sim)
and the user select a root block, then the user switch to another type of scan configuration and the previously selected root block is not available 

**TODO**: reset all atoms after the user quit a specific page of scan config


bilalesi:
@g-bar can i have your blessing here please

tolokoban (approved):

PR #1600, opened 2026-04-28
Vercel v6

WonderPG:
Related to https://github.com/openbraininstitute/prod-ai/issues/4

Can be reviewed by pointing to `AI_AGENT_URL = "https://staging.cell-a.openbraininstitute.org/api/agent-ts/api"` in your `.env.local`

PR #1647, opened 2026-05-11
New states

WonderPG:
Branches out of #1600 . Merge it first before this one.


ayimaok:
@WonderPG, several questions: 

- why is there this error message before the config is filled in. 
image
<br></br>
- initially, I was on the Workflows > simulation > ion channel model list view, asked to fill in the config, then changed the page to the config page and asked to generate the config, but still had this reply: 
image
<br></br>
- there were 2 issues, it successfully generated a config, all the fields are filled, but there is still an error message, don't understand. 2nd issue - I could not generate simulations with the error of calibration (a bit strange)
Screenshot 2026-05-21 at 14 38 01
<br></br>
- question: why the diff changes in the INFO tab are yellow and the other green? 
Screenshot 2026-05-21 at 15 14 04

WonderPG:
>why is there this error message before the config is filled in.

The error messages indicate that the tool call failed. The `Validate state` fails when the state is not valid. So it makes sense that it fails on a vanilla config.

> initially, I was on the Workflows > simulation > ion channel model list view, asked to fill in the config, then changed the page to the config page and asked to generate the config, but still had this reply:

This is a mistake of the agent. It should have tried again to read the state after you changed page. It incorrectly assumed you were still on the same page. We can eventually twist it a bit if this becomes an issue

> there were 2 issues, it successfully generated a config, all the fields are filled, but there is still an error message, don't understand. 2nd issue - I could not generate simulations with the error of calibration (a bit strange)

I am not sure what error message you are referring to. But I see on this screenshot that you are on the wrong preview. 
When it comes to the calibration issue, I quote Darshan:
```
In staging, please DON'T use public me-models created by Gil Barrios. Those were accidently made public. Only trust the service admin .  
```

> question: why the diff changes in the INFO tab are yellow and the other green?

Changes to existing fields are displayed in yellow. Addition of new fields are displayed in green. As a rule of thumb:
- Green = addition
- Yellow = modification
- Red = Removal

Feel free to reach out on teams if there are persisting issues.

ayimaok:
@WonderPG, 


What do you mean by 'wrong preview' ? This model is a public model, not the ones created by Gil. 

> I am not sure what error message you are referring to. But I see on this screenshot that you are on the wrong preview.
When it comes to the calibration issue, I quote Darshan

What I don't understand is that, it fails initially and then starts generating the config again
>The error messages indicate that the tool call failed. The Validate state fails when the state is not valid. So it makes sense that it fails on a vanilla config.

WonderPG:
>What do you mean by 'wrong preview' ? This model is a public model, not the ones created by Gil.

The URL in your screenshot is https://indicate-chat-can-do-simu.preview.openbraininstitute.org/... which belongs to another PR
Also the public models created by Gil don't work. You need to scroll down a bit in the list and take one that is created by 
`service-admin` for ME model simulation

> What I don't understand is that, it fails initially and then starts generating the config again

Exactly, the agent first checks what are the errors in your config, before fixing them. Or do you mean something else ?

ayimaok:
They are not Gil's models, they are mine and they have been validated and have calibration 
i am confused this link is not to be able to generate configs in the rest of the simulations in workflows
https://new-states.preview.openbraininstitute.org/

WonderPG:
>They are not Gil's models, they are mine and they have been validated and have calibration

Ah then this is a more general issue related to the ME model, most likely unrelated to my PR. I assume you have the same error when you click "Generate simulation(s)" right ?

>i am confused this link is not to be able to generate configs in the rest of the simulations in workflows
https://new-states.preview.openbraininstitute.org/

No this link is correct, but this screenshot 
image
that you linked in your previous message is not on the correct URL. Not sure if it is related to the issue you had.

ayimaok:
all right, thanks. All good. 
just to note that the yellow = modification - > these fields were empty initially, that's why I was a bit confused that they are in yellow.

WonderPG:
>just to note that the yellow = modification - > these fields were empty initially, that's why I was a bit confused that they are in yellow.

That is fair. However the two fields (Campaign Name and Campaign Description) already exist in the simulation config json. However when you add a nodeset for example, the field doesn't exist in the json and is added from scratch. Hence why the former is yellow and the latter is green.

src/components/ai-assistant/message-item/use-message-diffs.ts
```diff
@@ -114,6 +124,37 @@ export function useMessageDiffs({ message }: UseMessageDiffsArgs): MessageDiffAc
...
+    if (messageKey === null || messageKey !== activeKey) return false;
+
+    // Check that the message config belongs to the same entity as the current page.
+    // The URL's last path segment is the entity ID on entity-specific pages.
+    // Only check if it looks like a UUID (entity-specific pages).
+    const urlSegments = globalThis.location?.pathname?.split('/').filter(Boolean) ?? [];
+    const urlEntityId = urlSegments[urlSegments.length - 1];
+    const isUuid =
+      urlEntityId &&
+      /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i.test(urlEntityId);
```
bilalesi: this can be simply, 
```ts
const isUuid = z.uuid().safeParse(urlEntityId).success
```

src/features/scan-config/helpers.ts
```diff
@@ -6,15 +11,54 @@ export {
...
+  entityType: TExtendedEntitiesTypeDict,
+  activity: TScanConfigActivity,
+  entity?: { scale?: string }
+): TAIConfigKey {
+  if (activity === ScanConfigActivity.Process) return 'skeletonization_config';
+
+  // Simulate activity — resolve by entity type
+  if (entityType === ExtendedEntitiesTypeDict.MemodelCircuit) {
+    return 'me_model_simulation_config';
+  }
```
bilalesi, line 49: this better be moved later to the config registry [here](https://github.com/openbraininstitute/core-web-app/blob/7eb3a275c30d57c0b8863c91e03cb43dfd9db1c6/src/ui/segments/workflows/config/activities/simulate.ts)
may be adding 
```ts
scanConfig: {
      // current  config example
      definition: simulateIonChannelWorkflow,
      schemaName: SchemaNameDict.IonChannelModelSimulationScanConfig,
      configureBinding: ionChannelSimulationConfigureBinding(),
      // ai related
      aiStateKey: 'ion_channel_model_simulation_config'
}
```

src/features/scan-config/template.tsx
```diff
@@ -98,7 +98,11 @@ export function ScanConfigTemplate({
   const previousCampaignId = usePrevious(campaignId);
   const isCampaignIdChanged = previousCampaignId !== campaignId;
 
-  useAgentState(aiEnabled ? ACTIVITY_AI_CONFIG_MAP[activity] : '', config);
+  useAgentState(
+    aiEnabled ? getConfigKeyForEntity(entityType, activity, entity as { scale?: string } | undefined) : '',
```
bilalesi, line 102: you can use 
```ts
const { scanConfig, definition } = useScanConfigWorkflow();
// later
scanConfig.aiStateKey
````

File: src/components/ai-assistant/message-item/use-message-diffs.ts (hunk 1 above)
Comment (WonderPG) [reply src/components/ai-assistant/message-item/use-message-diffs.ts]: Nice idea, done in 08e4184c8

File: src/features/scan-config/helpers.ts (hunk 2 above)
Comment (WonderPG) [reply src/features/scan-config/helpers.ts:49]: Adding `aiStateKey` to the workflow registry works for `IonChannelModel`, `MemodelCircuit`, and `EMCellMesh` (skeletonization) since their sourceType matches the entity type that reaches the scan-config template.

For circuits (all scales), there's a mismatch: the registry uses specific subtypes as sourceType (`SingleNeuronCircuit`, `SingleNeuronSynaptome`, `Microcircuit`, etc.), but `getSupportedEntityTypesForScanConfiguration` resolves them to Circuit or `MEModelWithSynapses` before it reaches the template. So the lookup `workflows.find(w => w.sourceType === entityType)` doesn't find a match and deeper modifications are required to extend this registry option to every scale.

Also this restricts the `aiStateKey` to simulation workflows. In the future, I am pretty sure that `Build` workflows or any kind of workflow will also need to be AI enabled. Not sure if there is a similar mechanism for other types of workflows.

Comment (bilalesi) [reply src/features/scan-config/helpers.ts:49]: the fn: `getSupportedEntityTypesForScanConfiguration` will no longer exists in the new design, 
the registry have both the `sourceType` and `targetType`,
also the registry is configured for any workflows (build/simulate/process/extract) and, as any future workflow should be handled by obi-one's scan-config then the configurations are pretty similar, and extendable 

in this use case, each simulation page should be aligned with specific extended type so having the aiStateKey per each config will be easily resolved on any stage of the workflow components using the context

e.g:
```ts
 {
    ...WorkflowBrowseDefaults,
    ...WorkflowStagePresets.ScanConfig,
    sourceType: ExtendedEntitiesTypeDict.SmallMicrocircuit,
    targetType: ExtendedEntitiesTypeDict.SmallMicrocircuitSimulation,
    scanConfig: {
      definition: simulateSmallMicrocircuitWorkflow,
      schemaName: SchemaNameDict.CircuitSimulationScanConfig,
      configureBinding: circuitSimulationConfigureBinding(
        ExtendedEntitiesTypeDict.SmallMicrocircuit
      ),
    },
    configurationInputs: [{ type: ExtendedEntitiesTypeDict.SmallMicrocircuit }],
    order: 7,
    disabled: false,
},
```

WonderPG: So I agree that the `aiStateKey` has it's place there, but with the current state of the code, it would take a significant refactor to make it compatible with every workflow no ? Given that `getSupportedEntityTypesForScanConfiguration` will disappear I assume we should make this change once the new design is out. Or is it simpler than that ?

bilalesi: yeah yeah, my comment was suggested for later, after my PR gets merged 🙂

bilalesi: just wanted to give a quick heads-up

WonderPG: Ah I see, I will keep that in mind for the future then thanks !

src/features/scan-config/template.tsx
WonderPG: Will keep in mind as soon as the new simulation design is merged, thanks !

BoBer78 (approved):
LGTM, works fine for me locally and in the preview.

PR #1648, opened 2026-05-11
Fix SwipeableCardsList logic

pgetta:
## Summary

Fixes two bugs in `SwipeableCardsList` that broke its navigation controls.

### 1. Arrow / pagination buttons did nothing & active dot never updated on scroll

`ref.current` was read during render and passed into the `useMoveTo` / `useScrollWatcher` hooks. On first render the ref is still `null`, so:

- `useMoveTo`'s `useCallback` captured `div = null` → click handlers hit the early `return` and did nothing.
- `useScrollWatcher`'s effect bailed out before attaching the `scroll` listener, so `newsIndex` never updated as the user scrolled.

The ref attached after commit, but since reading `ref.current` during render doesn't subscribe to changes, no re-render ever happened to feed the real node back into the hooks.

**Fix:** replace the `useRef` with a `useState`-backed callback ref (`setScrollEl`). The component re-renders when the node attaches, and both hooks receive the real DOM element. Also memoized `makeId` so effect deps are stable.

### 2. Clicking a pagination dot scrolled to the wrong card

`useMoveTo` used `card.offsetLeft` as the target scroll position. `offsetLeft` is relative to the nearest *positioned* ancestor — neither `.scroll` nor `.swipeableCardsList` is positioned, so the value included the container's own offset on the page, not the card's offset inside the scroller.

Symptoms:
- From index 1, clicking dot 0 was a no-op (target scroll equalled current scroll).
- From the last index, clicking dot 0 landed on dot 1.

**Fix:** compute the scroll target from bounding rects, which is independent of the offsetParent:

```ts
const scrollPosition =
  div.scrollLeft + card.getBoundingClientRect().left - div.getBoundingClientRect().left;
```

### DIFF
--- a/src/components/LandingPage/components/swipeable-cards-list/swipeable-cards-list.tsx
+++ b/src/components/LandingPage/components/swipeable-cards-list/swipeable-cards-list.tsx
@@ -1,12 +1,11 @@
-/* eslint-disable react/no-array-index-key */
-/* eslint-disable no-param-reassign */
-import React, { CSSProperties, type JSX } from 'react';
+import { type CSSProperties, type JSX, useCallback, useEffect, useId, useState } from 'react';
+
+import { classNames } from '@/util/utils';
 
 import { IconChevronLeft } from '../../icons/IconChevronLeft';
 import { IconChevronRight } from '../../icons/IconChevronRight';
 import { styleButtonSquare } from '../../styles';
 
-import { classNames } from '@/util/utils';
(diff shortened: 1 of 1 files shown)


pgetta:
FYI, @ayimaok

ayimaok:
thank you @pgetta, this works well.

loris-olivier-obi (approved):
LGTM

PR #1649, opened 2026-05-11
Circuit nodes table

pgetta:
## Summary

Adds a **Circuit Nodes Table** for browsing SONATA node populations directly in the UI. Renders below the circuit image preview (scan-config).

image
(The bottom right corner)

***

Data flow: H5 asset is streamed via `CacheStorage` into a Web Worker, opened with `h5wasm`, and surfaced to ag-grid's infinite row model through Comlink. The main thread never touches the bytes.

## What's included

- **`src/features/circuit-nodes/`** — new feature module:
  - `circuit-nodes-table.tsx` — top-level component (view/population/display-mode state)
  - `components/` — toolbar, ag-grid wrapper, custom column header, categorical set-filter, column chooser
  - `hooks/use-circuit-config.ts` — fetches & parses `circuit_config.json`
  - `hooks/use-nodes-worker.ts` — manages worker lifecycle + ag-grid `IDatasource`
  - `worker/` — `nodes.worker.ts` (Comlink entry), `nodes-h5.ts` (`NodesSession`: column metadata, filtering, sorting, LRU-bounded loaded-column & view caches), `fetch-and-cache.ts` (streaming download into Emscripten FS via
`CacheStorage`)
- **`src/api/entitycore/queries/assets`** — new `buildAssetDownloadRequest()` that returns `{ url, headers }` for handing asset downloads off to environments outside the main `ApiClient` (e.g., workers).
- **Integration**: `scan-config/components/model-preview/circuit-preview.tsx` and `ui/segments/explore/circuit/elements/visualization.tsx`.
- **Deps**: `ag-grid-community`, `ag-grid-react` (`^32.3.5`).

## Capabilities

- Server-side row model — only visible rows are read from the H5 file.
- Categorical (set), text, and numeric filters; multi-column sort with order badges.
- Column chooser with drag-and-drop reordering.
- Three display modes: collapsed / half / full (resizes the image preview accordingly).
- Cross-session cache: H5 file is downloaded once per `(circuit, asset, population)` and reused.

## Performance & optimizations

**Memory**

- Worker-isolated: bytes never cross to the main thread. Comlink only transfers row pages (≤ `cacheBlockSize`).
- Streamed download — `response.body.getReader()` writes chunks directly into the Emscripten FS file in the worker; peak memory is ~chunk size, not the full asset size.
- `CacheStorage` (`obi-circuit-h5-v1`) holds the response across sessions — second visit is FS-only, no network.
- LRU caps on the hot caches in `NodesSession`:
  - `LOADED_COLUMN_CACHE_SIZE = 12` — full-column arrays (typed where possible: `Uint32Array` for categorical codes, `Float32/64Array` for numeric).
  - `VIEW_CACHE_SIZE = 8` — computed `(sort + filter) → Uint32Array of row indices`.
- Categorical columns stored as integer codes + shared `library` string table — no per-row string allocations.

**Read path**

- **Hyperslab fast-path**: when there's no sort and no filter, rows are read with `dataset.slice([[start, end]])` — a narrow HDF5 hyperslab read instead of materializing the full column. Lazy until the column is needed.
- **Gather path** (with sort/filter): the full column is loaded once into the LRU, and pages are gathered by index. Subsequent pages of the same view are O(page size).
- **View cache key canonicalization**: set-filter values are sorted before keying, so reordered selections hit the same cache entry.
- ag-grid config tuned for infinite scroll: `cacheBlockSize: 200`, `maxBlocksInCache: 10`, `rowBuffer: 50`, `blockLoadDebounceMillis: 75` — coalesces rapid scroll into batched worker calls.
- Synthetic `node_id` column is generated on the fly (no HDF5 read); typed `Uint32Array` cached once.

**Sort & filter**

- Set-filter compiled into a `Uint8Array` mask indexed by library code — O(1) per row vs. `Set.has(string)` lookups.
- Filter loop short-circuits on first failing predicate per row.
- Sort is stable (`a − b` tiebreaker on the original row index).
- Categorical sort uses a precomputed alphabetic `librarySortRank: Uint32Array` — sort comparator is two array reads + a subtract.

**Network & lifecycle**

- `buildAssetDownloadRequest` resolves URL+headers without firing the request, so the worker (which has its own auth context) can do the fetch directly.
- Generation counter + cancellation flag in `useNodesWorker` discards results from superseded `open()` calls instead of racing them into state.
- Worker is torn down (`Comlink.releaseProxy` + `worker.terminate`) on population change and unmount; the previous H5 file is `FS.unlink`-ed before opening the next one.
- Table renders only when `mode !== 'collapsed'` and `view === 'nodes'`, so collapsed mode is zero-cost beyond the toolbar.
- `CategoricalFilter` debounces model emission (`APPLY_DEBOUNCE_MS = 150`) so rapid checkbox toggling collapses into one `refreshInfiniteCache` call.
- Lazy-loaded entry point — `index.tsx` uses `next/dynamic({ ssr: false })` so ag-grid and `h5wasm` are kept out of the initial bundle and never run on the server.

## Test plan

- [x] Open a circuit with a sonata asset; verify default population (biophysical) loads.
- [x] Toggle collapsed → half → full; image preview resizes correctly.
- [x] Sort multiple columns (shift-click); verify order badges and stable ordering.
- [x] Apply categorical + numeric + text filters; verify filtered count matches.
- [x] Reorder & hide/show columns via drag handle and chooser modal.
- [x] Switch populations; verify worker is torn down and rebuilt without leaks.
- [x] Reload the page; verify H5 is served from `CacheStorage` (no re-download).

***

Relates to [Implement nodes table - circuit simulation designer](https://github.com/openbraininstitute/prod-circuit-simulation/issues/164).


src/features/circuit-nodes/hooks/use-nodes-worker.ts
```diff
@@ -0,0 +1,170 @@
...
+        setError(e instanceof Error ? e : new Error(String(e)));
+        setStatus('error');
+        setOpenResult(null);
+      }
+    })();
+
+    return () => {
+      cancelled = true;
+    };
+  }, [enabled, circuitId, circuitAssetId, population, ctx]);
```
tolokoban, line 97: I don't know if `useWorkspace()` will create a new `ctx` everytime.
If this is the case, maybe we can replace the dependency `ctx` with `ctx.virtualLabId, ctx.projectId`?

tolokoban: Yes. I have checked the implementation of `useWorkspace` and it will return a new object every render.

Maybe we can prevent this by changing a bit its implementation:

```ts
export function useWorkspace() {
  const { virtualLabId, projectId } = useParams<WorkspaceContext>();
  return React.useMemo(
    () => ({ virtualLabId, projectId }),
    [virtualLabId, projectId]
  );
}
```

src/features/circuit-nodes/circuit-nodes-table.tsx
```diff
@@ -0,0 +1,191 @@
...
+    if (!config) return;
+    if (populationName && config.nodes.some((p) => p.name === populationName)) return;
+    const next = pickDefaultPopulation(config.nodes);
+    setPopulationName(next?.name);
+  }, [config, populationName]);
+
+  const population = useMemo(
+    () => config?.nodes.find((p) => p.name === populationName),
+    [config, populationName]
+  );
```
g-bar: This pattern is a bit weird could we not simply do:

```ts
const [populationName, setPopulationName] = useState<string | undefined>(pickDefaultPopulation(config.nodes);
```

src/features/circuit-nodes/hooks/use-nodes-worker.ts
pgetta: Thank you, that is a good find Fabien. I've added memoization to the `useWorkspace` hook.

src/features/circuit-nodes/circuit-nodes-table.tsx
pgetta: Thank you Gil,

The config is coming **async**, so at first render it's undefined.
When it arrives we need to select the default population and its name.
And after when the user selects a different population (name) - the population object needs to reflect that.

So I'm not sure it's going to work like you've suggested or maybe I'm missing something?

g-bar: Not sure if I understand the useEffect, but it's a code smell since useEffect is for external side effects not internal state, as per: 

https://react.dev/learn/you-might-not-need-an-effect

The AI is suggesting this:

```ts
// 1. Keep state only for EXPLICIT user overrides
const [selectedName, setSelectedName] = useState<string | undefined>();

// 2. Derive the actual population during render
const population = useMemo(() => {
  if (!config) return undefined;

  // Use the user's selection if it exists and is still valid in the current config
  const current = config.nodes.find((p) => p.name === selectedName);
  if (current) return current;

  // Fallback to default if no selection or selection is invalid
  return pickDefaultPopulation(config.nodes);
}, [config, selectedName]);

// 3. Derived name for convenience
const populationName = population?.name;
````

Comment (pgetta) [reply src/features/circuit-nodes/circuit-nodes-table.tsx]: Ok, thank you, got it, that is a nice one. I need to teach my AI to follow react best practices ))

Should be addressed now.

**Review by g-bar — DISMISSED** [review]
awesome

File: package.json [hunk 3]
```diff
@@ -53,6 +53,8 @@
     "@tanstack/react-query-devtools": "^5.100.9",
     "@tanstack/react-virtual": "^3.13.24",
     "@tolokoban/tgd": "2.0.78",
+    "ag-grid-community": "^32.3.5",
+    "ag-grid-react": "^32.3.5",
```
bilalesi: nice, i was planning to use it for csv contribution, i will definitely migrate to it

pgetta: Yeah, JDC mentioned that he wants to eventually use this also in the explore tables to improve UX.

bilalesi (approved):

PR #1653, opened 2026-05-12
allow immediate running of notebook after upload

dkeller9:
### Problem

After contributing a new analysis notebook template, clicking **Run** in the action popover would silently do nothing. A full page refresh was required before the button worked.

### Root cause

`createNotebookAsync.onSettled` fired a query invalidation immediately after the notebook entity was created — before `uploadAssetsAsync` had run. The table re-fetched at that point and rendered the new row with `notebook.assets = []`.

In `ActionPopover`, `handleRunNotebook` does:

```ts
const asset = notebook.assets.find((n) => n.label === 'jupyter_notebook');
if (!asset) return;
```

With an empty `assets` array, `asset` was always `undefined` and the handler returned early with no error and no notification. A page refresh forced a fetch *after* the uploads had completed on the server, which is why it worked around the issue.

### Fix

Extracted the invalidation logic into a shared `invalidateNotebookQueries` helper and added `onSettled: invalidateNotebookQueries` to `uploadAssetsAsync`. This triggers a second query invalidation once all asset uploads have settled, ensuring the table row is re-fetched with a populated `assets` array before the user can interact with it.

No changes were needed in `ActionPopover` — its logic was correct throughout.

### Files changed

- `pipeline.ts` — added post-upload invalidation; extracted shared invalidation helper

### DIFF
--- a/src/ui/segments/contribute/analysis-notebook-template/pipeline.ts
+++ b/src/ui/segments/contribute/analysis-notebook-template/pipeline.ts
@@ -33,20 +33,22 @@ export function useAnalysisNotebookTemplatePipeline({
   const { projectId, virtualLabId } = useWorkspace();
 
   const invalidateNotebookQueries = () =>
-    Promise.all([
-      queryClient.invalidateQueries({
-        predicate: (query) =>
-          query.queryKey[0] === `data-entity-count-${ExtendedEntitiesTypeDict.Notebook}`,
-      }),
-      queryClient.invalidateQueries({
-        predicate: (query) =>
+    queryClient.invalidateQueries({
+      predicate: (query) => {
+        const firstKeySegment = query.queryKey[0];
+
+        const matchesEntityCount =
+          firstKeySegment === `data-entity-count-${ExtendedEntitiesTypeDict.Notebook}`;
+
+        const matchesExtendedEntity =
           get(
-            (query.queryKey as ExtendedEntityTypeQueryKey)[0],
+            firstKeySegment as ExtendedEntityTypeQueryKey[0],
             'context.extendedEntityType'
-          ) === ExtendedEntitiesTypeDict.Notebook,
-      }),
-      queryClient.invalidateQueries(),
-    ]);
+          ) === ExtendedEntitiesTypeDict.Notebook;
+
+        return matchesEntityCount || matchesExtendedEntity;
+      },
+    });
 
   const createNotebookAsync = useMutation({
     mutationFn: (values: TAnalysisNotebookTemplateForm) =>
--- a/src/ui/segments/contribute/analysis-notebook-template/pipeline.ts
+++ b/src/ui/segments/contribute/analysis-notebook-template/pipeline.ts
@@ -32,6 +32,22 @@ export function useAnalysisNotebookTemplatePipeline({
   const queryClient = useQueryClient();
   const { projectId, virtualLabId } = useWorkspace();
 
+  const invalidateNotebookQueries = () =>
+    Promise.all([
+      queryClient.invalidateQueries({
+        predicate: (query) =>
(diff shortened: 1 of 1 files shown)


src/ui/segments/contribute/analysis-notebook-template/pipeline.ts
```diff
@@ -32,6 +32,22 @@ export function useAnalysisNotebookTemplatePipeline({
...
+          query.queryKey[0] === `data-entity-count-${ExtendedEntitiesTypeDict.Notebook}`,
+      }),
+      queryClient.invalidateQueries({
+        predicate: (query) =>
+          get(
+            (query.queryKey as ExtendedEntityTypeQueryKey)[0],
+            'context.extendedEntityType'
+          ) === ExtendedEntitiesTypeDict.Notebook,
+      }),
+      queryClient.invalidateQueries(),
```
g-bar: This is invalidating every query in the cache. This wipes out all data, not only notebooks related.

g-bar (changes requested):

src/ui/segments/contribute/analysis-notebook-template/pipeline.ts
```diff
@@ -32,6 +32,22 @@ export function useAnalysisNotebookTemplatePipeline({
...
+        predicate: (query) =>
+          query.queryKey[0] === `data-entity-count-${ExtendedEntitiesTypeDict.Notebook}`,
+      }),
+      queryClient.invalidateQueries({
+        predicate: (query) =>
+          get(
+            (query.queryKey as ExtendedEntityTypeQueryKey)[0],
+            'context.extendedEntityType'
+          ) === ExtendedEntitiesTypeDict.Notebook,
+      }),
```
g-bar: You don't need an array with two calls to `invalidateQueries`, you should be able to invalidate both in the same call.

g-bar (changes requested):

src/ui/segments/contribute/analysis-notebook-template/pipeline.ts
g-bar: For example:

```ts
queryClient.invalidateQueries({
  predicate: (query) => {
    const firstKeySegment = query.queryKey[0];

    const matchesEntityCount =
      firstKeySegment === `data-entity-count-${ExtendedEntitiesTypeDict.Notebook}`;

    const matchesExtendedEntity =
      get(
        firstKeySegment, 
        'context.extendedEntityType'
      ) === ExtendedEntitiesTypeDict.Notebook;

    return matchesEntityCount || matchesExtendedEntity;
  },
});
```

g-bar: You probably need a key for the assets related to the notebook?

src/ui/segments/contribute/analysis-notebook-template/pipeline.ts
dkeller9: Good catch, removed. After investigating the codebase, getNotebook (singular) is never used in a useQuery call, so assets don't have a separate cache entry — they're embedded in the list response. The two targeted predicates are sufficient.

src/ui/segments/contribute/analysis-notebook-template/pipeline.ts
dkeller9: > You don't need an array with two calls to `invalidateQueries`, you should be able to invalidate both in the same call.

Fixed, merged into a single call with the predicate pattern you suggested.

dkeller9: > You probably need a key for the assets related to the notebook?

Traced this through — getNotebook is only called imperatively in the download handler, never via useQuery, so there's no separate asset query key in the cache. Assets come back embedded in INotebook from the list query, which is already covered by matchesExtendedEntity.

g-bar (approved):
nice job

PR #1656, opened 2026-05-13
Hide non-simulatable circuits in simulate workflows

pgetta:
## Summary

Circuit entities with `has_electrical_cell_models: false` (purely anatomical / EM reconstructions, e.g. `MICrONS_v1`, `H01_v1`) were appearing as selectable in Simulate workflow pickers and exposing the "Simulate" action on the Circuit detail view, even though launching a simulation against them fails.

## Changes

- **`ICircuitFilter` / `CircuitBase`** (`src/api/entitycore/types/entities/circuit.ts`) — surface `has_electrical_cell_models` on the entity and add it as an optional filter (mirrors the `has_segmented_spines` precedent in `CellMorphologyFilter`).
- **`EntityCoreTypeConfig.isSimulatable` contract** (`src/entity-configuration/domain/types.ts`) — broadened from `boolean | ((scale: TCircuitScaleDictionary) => boolean)` to `boolean | ((entity: T) => boolean)`, so each entity type can express its own simulatability rule against its full entity shape rather than just the circuit scale.
- **Per-entity simulatability rules colocated in entity config** (`src/entity-configuration/domain/model/*.ts`):
  - `SingleNeuronCircuit`, `SmallMicrocircuit`, `Microcircuit` — `(entity) => entity.has_electrical_cell_models` (previously hard-coded `false`).
  - `Circuit` (base) — `(entity) => entity.has_electrical_cell_models && scale ∈ {Single, PairNeuron, SmallMicrocircuit, Microcircuit}`.
- **Action menu** (`src/ui/segments/action-menu.tsx`) — function-form `isSimulatable` now receives the entity directly; the simulatability rule lives in entity config rather than at the call site.
- **Simulate workflow pickers** (`src/ui/segments/workflows/config/activities/simulate.ts`) — shared `simulatableCircuitFilters = { has_electrical_cell_models: true }` constant applied as `configurationInputs.filters` on the 5 Circuit-backed workflows: Single Neuron, Paired Neuron, Small Microcircuit, Microcircuit, Region Circuit (BrainRegion).
- **Typo fix** — `domain/model/mirocircuit.ts` → `microcircuit.ts`.

## Scope

Synaptomes are intentionally untouched — `ISingleNeuronSynaptome` has no `has_electrical_models` field; synaptomes are always simulatable per the team. The anatomical-looking entries flagged in the issue's synaptome reproducer are expected to be Circuits surfacing in the wrong picker, which the Circuit-side filter resolves.

## Test plan

- [x] Open each Simulate workflow (Single Neuron / Paired / Small Microcircuit / Microcircuit / Region) — request to `/circuit` includes `has_electrical_cell_models=true`; structural circuits (`MICrONS_v1`, `H01_v1`) no longer appear; simulatable circuits still appear.
- [x] On a Circuit detail page with `has_electrical_cell_models: false` — Simulate action is hidden.
- [x] On a Circuit detail page with `has_electrical_cell_models: true` (Small Microcircuit / Pair Neuron) — Simulate action is visible and navigates to the configure page.
- [x] General Models-library browses of Microcircuit / SmallMicrocircuit — structural circuits still appear (filter must not apply outside Simulate).

***

Relates to [Synaptome beta - models without electrical models are selectable](https://github.com/openbraininstitute/prod-singlecell-simulation/issues/279).

### DIFF
--- a/src/ui/segments/action-menu.tsx
+++ b/src/ui/segments/action-menu.tsx
@@ -16,7 +16,6 @@ import NextLink from 'next/link';
 import { notFound, useRouter } from 'next/navigation';
 
 import { deleteCellMorphology } from '@/api/entitycore/queries/experimental/cell-morphology';
-import { EntityTypeDict } from '@/api/entitycore/types/entity-type';
 import {
   ExtendedEntitiesTypeDict,
   type TExtendedEntitiesTypeDict,
@@ -128,14 +127,10 @@ export default function ActionMenu({
     },
   });
 
-  const isAnatomicalCircuit =
-    entityType.type === EntityTypeDict.Circuit &&
-    (entity as ICircuit).has_electrical_cell_models === false;
-
   const isSimulatable =
-    (typeof entityType.isSimulatable === 'boolean'
+    typeof entityType.isSimulatable === 'boolean'
       ? entityType.isSimulatable
-      : 'scale' in entity && entityType.isSimulatable(entity.scale)) && !isAnatomicalCircuit;
+      : (entityType.isSimulatable as (e: typeof entity) => boolean)(entity);
 
   return (
     <div className="text-primary-9 mt-10 flex flex-col gap-5 px-5 text-base font-bold">
--- a/src/ui/segments/action-menu.tsx
+++ b/src/ui/segments/action-menu.tsx
@@ -128,14 +128,14 @@ export default function ActionMenu({
     },
   });
 
+  const isAnatomicalCircuit =
+    entityType.type === EntityTypeDict.Circuit &&
+    (entity as ICircuit).has_electrical_cell_models === false;
+
   const isSimulatable =
     (typeof entityType.isSimulatable === 'boolean'
       ? entityType.isSimulatable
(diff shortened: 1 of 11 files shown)


tolokoban (dismissed):

src/ui/segments/action-menu.tsx
```diff
@@ -127,10 +128,14 @@ export default function ActionMenu({
     },
   });
 
+  const isAnatomicalCircuit =
+    entityType.type === EntityTypeDict.Circuit &&
+    (entity as ICircuit).has_electrical_cell_models === false;
```
bilalesi: Why not moving this directly to the issimulatable fn per config ?

pgetta: Yeah, I was thinking maybe later on we will have more checks which will require access to assets which are not available in the definition, so didn't do that, but maybe that is too far in the future.

I've moved the check to the circuit definition as you've suggested.

pgetta:
This also addresses [Simulate button in detailed page for all physiological circuits](https://github.com/openbraininstitute/prod-circuit-simulation/issues/186).

bilalesi (approved):

PR #1663, opened 2026-05-19
Enable SWR caching for heavy reused payloads

pgetta:
## Why

A number of payloads were being re-fetched on every page load:

- **Point cloud** - Apache Arrow `ArrayBuffer` per circuit/region; large, recomputable backend response
- **Brain meshes** - GLB files sometimes having the size of up to 20 MBs.
- **Cell composition summary asset** - bulky JSON describing the full mouse atlas composition tree

Both are stable enough that aggressive client-side caching is safe, and large enough that re-downloading on every reload is a real cost (bandwidth + time-to-render). Persisting them in Cache Storage with a stale-while-revalidate window lets the next reload render instantly while a background fetch refreshes silently.

The SWR machinery already existed inside `ApiClient` (used for mesh GLBs), but `fetchPointCloud` bypasses `ApiClient` entirely because the cell-api needs a `nexus-token` header instead of `Authorization: Bearer`. Extracting the cache logic into its own module lets both `ApiClient`-driven and raw-`fetch` call sites share the same SWR implementation.

## Summary

- Extract Cache Storage SWR from `ApiClient` into a standalone `src/api/cache-storage.ts` module
- Add `swrCacheConfig(name, ttl?, swr?)` factory — 1d fresh / 6d stale-while-revalidate defaults match the existing mesh cache windows; per-call overrides for future tuning
- Cache point cloud responses (`fetchPointCloud`) and the cell composition summary asset (`downloadAsset` opt-in `cache` param)

## Behavior preserved

- `ApiClient` delegates to the extracted functions — tri-state (fresh/stale/expired), inflight-revalidation dedupe, header sanitisation, and `ignoreVary` on lookup are unchanged
- `downloadAsset`'s three overloads gain optional `cache?: CacheConfiguration`; the ~15 existing callers are untouched
- Mesh GLB caching continues to work identically

## Test plan

- [x] Hard-refresh a brain-atlas-viewer page → `brain-atlas-meshes`, `brain-atlas-point-clouds` populate in DevTools → Application → Cache Storage; second load shows `[cached:fresh]` logs with no network calls
- [x] Visit a page triggering `useCellCompositionQuery` → new `cell-composition-summary-asset` bucket populates; reload shows `[cached:fresh]`
- [x] Temporarily lower `ttlInSeconds` to force a stale read → verify exactly one background `/circuit` revalidation per URL
- [x] Point cloud error mapping (`BRAIN_REGION_DOES_NOT_EXIST`) surfaces unchanged
- [x] Other `downloadAsset` callers (NWB traces, ion-channel recording, thumbnails) still hit the network and don't appear in any Cache Storage bucket

### DIFF
--- a/src/api/cache-storage.ts
+++ b/src/api/cache-storage.ts
@@ -55,9 +55,7 @@ export async function checkCache(
 
   try {
     const cache = await caches.open(cacheConfig.cacheName);
-    // ignoreVary: defensive against entries stored by older versions of this
-    // wrapper that carried over a `Vary` header from a redirected origin.
-    const cachedResponse = await cache.match(url, { ignoreVary: true });
+    const cachedResponse = await cache.match(url);
 
     if (!cachedResponse) {
       return { state: 'miss', response: null };
--- a/src/api/cache-storage.ts
+++ b/src/api/cache-storage.ts
@@ -1,0 +1,171 @@
+import { log } from '@/utils/logger';
+
+export type CacheConfiguration = {
+  enabled: boolean;
+  ttlInSeconds: number;
+  cacheName: string;
+  excludeUrls?: RegExp[];
+  // When set, a cached entry whose age is between ttlInSeconds and
+  // (ttlInSeconds + staleWhileRevalidateSeconds) is returned immediately while a
+  // background fetch refreshes the cache. Past that window, treat as a miss.
+  staleWhileRevalidateSeconds?: number;
+};
+
+export type CacheState = 'fresh' | 'stale' | 'expired' | 'miss';
+
+// Defaults: 1 day fresh, 6 days stale-while-revalidate. Tuned for slow-changing
+// reference data (atlas regions, meshes, point clouds, composition summary).
+// Override per call site when a different cadence makes sense.
+export function swrCacheConfig(
+  cacheName: string,
+  ttlInSeconds: number = 24 * 60 * 60,
+  staleWhileRevalidateSeconds: number = 6 * 24 * 60 * 60
+): CacheConfiguration {
+  return {
+    enabled: true,
+    cacheName,
+    ttlInSeconds,
+    staleWhileRevalidateSeconds,
+  };
+}
+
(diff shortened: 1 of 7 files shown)


pgetta:
Relates to [Optimize brain mesh viewer](https://github.com/openbraininstitute/core-web-app/issues/1650).

src/api/cache-storage.ts
```diff
@@ -0,0 +1,169 @@
...
+  // When set, a cached entry whose age is between ttlInSeconds and
+  // (ttlInSeconds + staleWhileRevalidateSeconds) is returned immediately while a
+  // background fetch refreshes the cache. Past that window, treat as a miss.
+  staleWhileRevalidateSeconds?: number;
+};
+
+export type CacheState = 'fresh' | 'stale' | 'expired' | 'miss';
+
+// Defaults: 1 day fresh, 6 days stale-while-revalidate. Tuned for slow-changing
+// reference data (atlas regions, meshes, point clouds, composition summary).
```
bilalesi: it's hard to assume this values, i understand that this resources are rarely changed, but i would say the re-validation request should comes from the server side (see here https://teams.microsoft.com/l/message/19:ee65d1fe2ad445aa8f127255886df473@thread.v2/1775545916677?context=%7B%22contextType%22%3A%22chat%22%7D) ,
as the server is source of truth, it must command the frontend what to do.

I did add caching to brain regions before (first migration to entitycore) but then i concluded that it's hard to assume things specially now, that the multi-species is fresh in production and can be changed anytime, 

I would say we keep things stable and we move this direction when multi-species is confirmed  (if server side headers are not a priority)

pgetta: Yeah, good point, I've reverted the change in `src/api/entitycore/queries/general/brain-atlas.ts`

So now only brain meshe, cell composition and the point cloud assets are cached.

bilalesi: let's go !

bilalesi (approved):

pgetta:
Thank you Bilal

PR #1664, opened 2026-05-19
fix(scan-config): preserve dictionary entry order on rename

bilalesi:
rebuild the atoms map in place when renaming `BlockDictionary` keys so items stay in their original position instead of moving to the end of the list

### DIFF
--- a/src/features/scan-config/components/block-dictionary-entries.tsx
+++ b/src/features/scan-config/components/block-dictionary-entries.tsx
@@ -8,6 +8,7 @@ import {
   WarningFilled,
 } from '@ant-design/icons';
 import { Input } from 'antd';
+import { kebabCase } from 'es-toolkit/compat';
 import { atom, useAtom } from 'jotai';
 import { AnimatePresence, motion } from 'motion/react';
 import { Fragment, memo, useMemo } from 'react';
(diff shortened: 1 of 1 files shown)


pgetta (approved):

src/features/scan-config/components/block-dictionary-entries.tsx
```diff
@@ -158,8 +160,11 @@ export default function BlockDictionaryEntries({
     allEntries.delete(selectedEntry);
     allEntries.add(newKey);
 
-    selectedTabAtoms[newKey] = selectedTabAtoms[selectedEntry];
-    delete selectedTabAtoms[selectedEntry];
+    const renamedTabAtoms = Object.fromEntries(
+      Object.entries(selectedTabAtoms).map(([key, value]) =>
+        key === selectedEntry ? [newKey, value] : [key, value]
+      )
+    );
```
g-bar, line 167: 👍

g-bar:
This is probably going to conflict with : https://github.com/openbraininstitute/core-web-app/pull/1619

But I guess we can merge and I'll fix the conflict on my branch.

g-bar (approved):

PR #1665, opened 2026-05-19
Fix brain region selection being overridden by global workspace state in cell morphology contribution form

dkeller9:
When contributing a cell morphology, any brain region selected inside the contribution form was being overridden by the globally selected brain region from the workspace.
The root cause was that `BrainRegionSelector` called `useWorkspaceHierarchyRegistry()` directly, coupling the form field to live global state. Any change to the global brain region selection would propagate into the dropdown and overwrite the user's local selection.
The fix removes the global registry dependency from `BrainRegionSelector` entirely. The component now accepts `defaultBrainRegion`, `value`, and `onChange` as props, making it purely form-controlled via Ant Design's `Form.Item` value/onChange injection. The initial brain region is read once from the form instance via `useRef` in `Setup`, which was itself initialized from the session's starting brain region. This means the selector correctly pre-populates on open but remains independent of any subsequent global state changes.

The original bug ticket filed by Natali is:
https://github.com/openbraininstitute/prod-explore-functionality/issues/494


src/ui/segments/contribute/cell-morphology/steps/setup.tsx
```diff
@@ -12,9 +12,11 @@ import {
   renderLabel,
 } from '@/ui/segments/contribute/shared/helpers';
 import { cn } from '@/utils/css-class';
+import { useRef } from 'react';
```
pgetta: Import order: `react` is an external package and belongs in the top group (above the blank line, alongside `antd`/`dayjs`). The current placement after the `@/*` block will fail Biome's `useSortedImports` rule. Move to line 5 area:

```ts
import { InfoCircleFilled } from '@ant-design/icons';
import { DatePicker, Form, Input, InputNumber, Space } from 'antd';
import dayjs from 'dayjs';
import { useRef } from 'react';
```

src/ui/segments/contribute/cell-morphology/steps/setup.tsx
```diff
@@ -12,9 +12,11 @@ import {
...
 
 export function Setup() {
   const form = Form.useFormInstance();
+  const initialBrainRegionId = useRef(form.getFieldValue(['setup', 'brain_region_id']));
```
pgetta: **Suggestion A (preferred): initialize via the form, not a per-caller `useRef`.**

This line pushes the "snapshot the initial value" responsibility onto every caller of `BrainRegionSelector`. The form is the natural owner of initial field values, and Ant Design has a first-class API for it. Where this form is mounted, do one of:

```tsx
<Form initialValues={{ setup: { brain_region_id: selectedBrainRegion?.id } }} ...>
```

or, if `selectedBrainRegion` isn't available at form construction:

```tsx
useEffect(() => {
  form.setFieldValue(['setup', 'brain_region_id'], selectedBrainRegion?.id);
}, []); // mount only — intentional empty deps
```

With that, `BrainRegionSelector` can stay zero-prop (no `defaultBrainRegion`, no type cast), the other 5 contribute forms get fixed for free, and there's no risk a future caller forgets the snapshot dance.

src/ui/segments/contribute/cell-morphology/steps/setup.tsx
```diff
@@ -62,8 +64,8 @@ export function Setup() {
           },
         ]}
       >
-        <BrainRegionSelector />
-      </Form.Item>
+      <BrainRegionSelector defaultBrainRegion={{ id: initialBrainRegionId.current } as IBrainRegionHierarchy} />
```
pgetta: Three things on this line:

1. **Type cast lie.** `{ id: initialBrainRegionId.current } as IBrainRegionHierarchy` works only because `BrainRegionDropdownWithFormItem` happens to read only `.id` today. If anyone reads another field from `defaultBrainRegion`, this crashes at runtime with no type error. If we keep the prop, narrow it to `defaultBrainRegionId?: string` or `Pick<IBrainRegionHierarchy, 'id'>`.
2. **Indentation:** 6 spaces, but siblings are at 8. Biome will rewrite.
3. If Suggestion A (above) is taken, this whole line goes away and `<BrainRegionSelector />` stays unchanged from `main`.

src/ui/segments/contribute/cell-morphology/steps/setup.tsx
```diff
@@ -62,8 +64,8 @@ export function Setup() {
...
+ </Form.Item>
```
pgetta: Indentation: only 1 leading space here vs. 6 on the sibling `</Form.Item>` tags. Biome format will fix, but please run `pnpm format` before pushing.

src/ui/segments/contribute/shared/components/brain-region-selector.tsx
```diff
@@ -1,17 +1,22 @@
...
 
-export function BrainRegionSelector() {
-  const { selectedBrainRegion } = useWorkspaceHierarchyRegistry();
+interface IBrainRegionSelectorProps {
+  defaultBrainRegion?: IBrainRegionHierarchy;
+  value?: string;
+  onChange?: (value: string) => void;
+}
 
+export function BrainRegionSelector({ defaultBrainRegion, value, onChange }: IBrainRegionSelectorProps) {
```
pgetta: After Suggestion A (initialize via the form), this component can revert to its original zero-prop shape — `Form.Item` injects `value`/`onChange` automatically, and `defaultBrainRegion` becomes unnecessary because the form field is already populated. The bug only existed because the component was reading global state directly; once that's gone, no prop is needed.

If you'd rather keep a prop-based API, please narrow `defaultBrainRegion` to `defaultBrainRegionId?: string` so callers don't have to lie to the type system as in `setup.tsx:67`.

src/ui/segments/contribute/shared/components/brain-region-selector.tsx
```diff
@@ -1,17 +1,22 @@
...
   return (
     <BrainRegionDropdownWithFormItem
       clsx={{ trigger: 'rounded-full w-full h-12', content: 'z-[99999]' }}
       showIcon={false}
       charsPerLine={200}
-      defaultBrainRegion={selectedBrainRegion as IBrainRegionHierarchy}
+      defaultBrainRegion={defaultBrainRegion}
```
pgetta: **Suggestion B: also fix the root cause one layer deeper in `src/features/brain-region-dropdown/form-dropdown.tsx`.**

The reason live global state propagated into this dropdown is that `BrainRegionDropdownWithFormItem` re-derives `defaultBrainRegion` on every render. Capturing it once on mount makes "default" behave as the name implies and prevents this class of bug from recurring through any future caller:

```tsx
export function BrainRegionDropdownWithFormItem({ defaultBrainRegion, value, onChange, ... }) {
  const [initialDefault] = useState(defaultBrainRegion); // snapshot once
  const { result: brainRegionHierarchy } = usePrimaryExtendedHierarchySpeciesQuery();
  ...
  return (
    <BrainRegionDropdown
      defaultBrainRegion={
        value
          ? brainRegionHierarchy?.options.find(({ value: v }) => value === v)?.data
          : brainRegionHierarchy?.options.find(({ value: v }) => initialDefault?.id === v)?.data
      }
      ...
    />
  );
}
```

One line, zero API churn, defense-in-depth on top of Suggestion A.

pgetta (commented):
Thanks for tracking this one down — the root-cause analysis is spot-on. The fix works locally for cell-morphology, but I think we can solve it at a better layer and cover the other 5 contribute forms (`em-cell-mesh`, `experimental-neuron-density`, `experimental-bouton-density`, `synapses-per-connection`, `electrical-cell-recording`) in the same change. Two suggestions inline; here's the shape:

### Suggestion A — initialize the form field, not the component
The form is the natural owner of "what value this field starts at". If we set `setup.brain_region_id` once (via `initialValues` on the parent `<Form>`, or `form.setFieldValue` in a mount-only effect), then `BrainRegionSelector` can stay a zero-prop dumb controlled input — `Form.Item` already injects `value`/`onChange`. No `useRef` snapshot per caller, no `defaultBrainRegion` prop, no `as IBrainRegionHierarchy` cast. The other 5 contribute forms get the same treatment for free.

### Suggestion B — make `BrainRegionDropdownWithFormItem`'s `defaultBrainRegion` actually mean "default"
The deeper bug is in `src/features/brain-region-dropdown/form-dropdown.tsx`: `defaultBrainRegion` is re-derived on every render. If it were captured once on mount (`const [initialDefault] = useState(defaultBrainRegion);`), "default" would behave the way every caller already assumes, and this class of bug couldn't recur via a different caller. Worth doing **in addition** to A as defense-in-depth — it's a one-line change at the root of the problem.

### Other small things
- Import order at `setup.tsx` line 15 — `react` belongs in the external-packages block at the top, not after `@/*`. Biome will flag this.
- Indentation at `setup.tsx` lines 67–68 is off (6 spaces / 1 space vs. siblings at 8). `pnpm format` will rewrite, but worth doing pre-push.
- No tests — a small render test that mounts the selector inside a `Form`, asserts initial value renders, then mutates global state and asserts the field is unchanged would lock this in. Not blocking.

Nothing here is a hard blocker — happy to discuss if you'd rather keep the per-caller `useRef` pattern.

src/ui/segments/contribute/cell-morphology/steps/setup.tsx
dkeller9: I have reverted `setup.tsx `to its original.

src/ui/segments/contribute/cell-morphology/steps/setup.tsx
dkeller9: Implemented by freezing initialValues with an empty dependency useMemo in `context.tsx`, which applies to all 6 contribute forms.

src/ui/segments/contribute/cell-morphology/steps/setup.tsx
dkeller9: `setup.tsx` is not modified in the final solution.

src/ui/segments/contribute/shared/components/brain-region-selector.tsx
dkeller9: Implemented by snapshotting defaultBrainRegion with useState in BrainRegionDropdownWithFormItem, preventing re-derivation on subsequent renders.

src/ui/segments/contribute/shared/components/brain-region-selector.tsx
dkeller9: `BrainRegionSelector` reverted to zero-prop, no global registry dependency

src/ui/segments/contribute/cell-morphology/index.tsx
```diff
@@ -76,12 +77,13 @@ interface ICellMorphologyProps {
...
   const { projectId, virtualLabId } = useWorkspace();
   const { selectedBrainRegion } = useWorkspaceHierarchyRegistry();
+  const brainRegionId = useRef(selectedBrainRegion?.id ?? null);
 
   return (
     <ContributionForm
       config={cellMorphologyConfig}
       sessionId={sessionId}
-      brainRegionId={selectedBrainRegion?.id!}
+      brainRegionId={brainRegionId.current}
```
pgetta: Now that `context.tsx` freezes `initialValues` via empty-deps `useMemo`, this per-caller `useRef` snapshot is doing the same work one layer down — exactly the pattern the previous review pushed back on. The other 5 contribute forms already work without it; suggest dropping it here too for consistency.

```suggestion

  return (
    <ContributionForm
      config={cellMorphologyConfig}
      sessionId={sessionId}
      brainRegionId={selectedBrainRegion?.id ?? null}
```

pgetta (commented):
Architectural fix looks solid — freezing `initialValues` at the pipeline-context layer cleanly covers all 6 contribute forms, and the `useState` snapshot in `BrainRegionDropdownWithFormItem` is good defense-in-depth. Two small follow-ups before merge.

src/ui/segments/contribute/cell-morphology/index.tsx
```diff
@@ -7,6 +7,7 @@ import {
   createCellMorphologyConfig,
 } from '@/ui/segments/contribute/cell-morphology/config';
 import { useCellMorphologyPipeline } from '@/ui/segments/contribute/cell-morphology/pipeline';
+import { useRef } from 'react';
```
pgetta: If you drop the `useRef` below (see other comment), this import can go too.

src/features/brain-region-dropdown/form-dropdown.tsx
```diff
@@ -268,4 +269,4 @@ export function BrainRegionDropdownWithFormItem({
       showIcon={showIcon}
     />
   );
-}
+}
```
pgetta: Missing trailing newline at EOF (`\ No newline at end of file` in the diff). Please run:

```bash
pnpm format src/features/brain-region-dropdown/form-dropdown.tsx
```

Scoped to this file on purpose — parts of the codebase haven't been migrated to Biome formatting yet, so we don't want to run `pnpm format` over everything.

src/ui/segments/contribute/cell-morphology/index.tsx
dkeller9: The useRef snapshot in index.tsx has been dropped since context.tsx already freezes initialValues at mount and consequentally the per-caller snapshot is redundant. Reverted to passing selectedBrainRegion?.id ?? null directly.

src/ui/segments/contribute/cell-morphology/index.tsx
dkeller9: Removed along with the useRef usage.

src/features/brain-region-dropdown/form-dropdown.tsx
dkeller9: Newline added.

pgetta (approved):
Thank you for the fixes, Daniel. Nice work!

dkeller9:
Thanks for reviewing!

PR #1667, opened 2026-05-19
feat-110: multi selection UI (part-1)

bilalesi:
refactors scan-config **configure** routes into a shared workflow layer: declarative definitions, provider context, and a page factory. workflow pages now is a single `createScanConfigWorkflowPage(definition)` export.

#### Main changes:
- **`useScanConfiguration`** — entity + ObiOne schema + grid endpoint + optional circuit mapping; feeds `ScanConfigContainer` / `ScanConfigTemplate`.
- **`useScanConfigOriginCampaign`**:  loads campaign via `?initialCampaignId=` and workflow-specific `resolve`.
- **`ScanConfigWorkflow`**:  `Provider` → entity load (`route-id` | `static-type`) → campaign → `Gate` / `Frame` / `Editor`.
- **Definitions** under `workflow/definitions/` for simulate circuit, ion channel, extract circuit, process EM cell mesh, build EM synapse mapping.

- entity sources: **`route-id`** (suspense fetch) or **`static-type`** only.

**NOTE:** this PR is the first step to introduce the multi-selection that can be done either in the first step "/new" page, or within the scan configuration, there will be 2 PRs later to be able to achieve the wanted behavior
1. move the selection to session storage
2. implement multi-selection for `em-synapse-mapping`

### DIFF
--- a/src/features/scan-config/workflow/use-workflow-entity.ts
+++ b/src/features/scan-config/workflow/use-workflow-entity.ts
@@ -1,0 +1,64 @@
+'use client';
+
+import { useSuspenseQuery } from '@tanstack/react-query';
+import { useMemo } from 'react';
+
+import type {
+  ScanConfigEntitySourceMode,
+  TResolvedScanConfigEntity,
+  TScanConfigEntitySource,
+} from '@/features/scan-config/workflow/types';
+import type { WorkspaceContext } from '@/types/common';
+
+function readRouteEntityId(
+  routeParams: Record<string, string | undefined>,
+  param = 'id'
+): string | undefined {
+  return routeParams[param];
+}
+
+export function useStaticTypeWorkflowEntity(
+  entitySource: Extract<
+    TScanConfigEntitySource,
+    { mode: typeof ScanConfigEntitySourceMode.StaticType }
+  >
+): TResolvedScanConfigEntity {
+  return useMemo(
+    () => ({
+      entity: null,
+      entityType: entitySource.entityType,
+    }),
+    [entitySource.entityType]
+  );
+}
+
+export function useRouteIdWorkflowEntity({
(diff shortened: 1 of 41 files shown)


src/features/scan-config/workflow/use-workflow-entity.ts
```diff
@@ -0,0 +1,64 @@
...
+  routeParams: Record<string, string | undefined>;
+}): TResolvedScanConfigEntity {
+  const param = entitySource.param ?? 'id';
+  const entityId = readRouteEntityId(routeParams, param);
+
+  if (!entityId) {
+    throw new Error(`Missing route param "${param}" for scan config workflow entity`);
+  }
+
+  const { data: entity } = useSuspenseQuery({
```
tolokoban, line 56: A minor point, but this violates a bit React rules for hooks: there is a throw just before this `useSuspenseQuery`.

tolokoban (commented):
Looks good @bilalesi !
I just have a little comment, to be sure it was intentional.

src/features/scan-config/workflow/use-workflow-entity.ts
bilalesi: good point,  but in this case is different: 
`ID`: url param
on a ` configure` page, id is almost always either:
- there from the first paint, or
- missing and we keep throwing (error screen / 404),
not: missing on render 1, then present on render 2 without unmounting
If it never flips like that, react always sees the same path -> so not bug

tolokoban (dismissed):

bilalesi:
@g-bar @pgetta it's also nice to have your feedback here 🙏🏼

pgetta (approved):
Very solid refactor, LGTM

PR #1669, opened 2026-05-19
Upgrade biome and apply other related corrections

pgetta:
This also makes sure that the local Biome version from `node_modules` is used (and not the `latest` one every time pulled over the network) during the pre-commit hook execution. As well as the config schema version is matching.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.4.13/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.4.15/schema.json",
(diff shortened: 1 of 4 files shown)


tolokoban (approved):
Good!
I'll you know if I still get the false positive.

pgetta:
Thanks for the approval.

Please make sure your VSCode is showing the correct version in the panel:
image

tolokoban:
Most of the times, it is showing me this:
image

And I have to restart it. But afterwards, i get the right version:

image


@pgetta  the false positives are still here:

image

> This private class member is defined but never used.

PR #1670, opened 2026-05-20
Display detailed morphology validation error messages

dkeller9:
## Summary

When a morphology file fails server-side validation, the upload UI now displays 
the specific error message from the backend (e.g., "Morphology validation failed: 
SomaNonConform", "Soma diameter validation failed: Unrealistic soma diameter 
detected") instead of a generic "could not be validated" message.

## Motivation

The backend validation endpoints (`/test-neuron-file` and 
`/register-morphology-with-calculated-metrics`) return structured error responses 
with specific failure reasons when morphology files fail validation. Previously, 
the frontend discarded this information entirely:

1. The API client threw before the response body was parsed when using 
   `asRawResponse: true`
2. Even if the error had been parsed, the error handler only read `error.message` 
   (a flat string) and ignored the structured `detail` field

Users saw only "The neuron file ... could not be validated by our tool" with no 
indication of what was actually wrong — whether it was a file format issue, 
invalid soma, structural problems, or something else.

### `api-client.ts`
- Modified the `!response.ok` error-throw block inside `_request` to check 
  `&& !config.asRawResponse` before throwing
- When `asRawResponse: true` is set, the contract is that the caller receives 
  the raw `Response` and handles errors itself
- The previous code honored this for success responses but threw unconditionally 
  on failure, preventing the caller from reading the error body

**Why this matters**: This change allows any future `asRawResponse` caller to 
handle error responses themselves, not just our morphology validation use case.

### `api/one/cell-morphology.ts`
- `resolveNeuronFile` now handles non-ok responses explicitly:
  - Calls `response.json()` to parse the error body
  - Unwraps FastAPI's outer `{ detail: ... }` envelope
  - Extracts the inner `{ code, detail }` object
  - Throws a plain `Error` with the specific validation message as 
    `error.message` and the full structured body attached as 
    `error.neuronFileError`
- Exported `NeuronFileError` type (containing `code` and `detail` fields)
- Exported `isNeuronFileError` type-guard function for the UI layer
- The guard checks for the `neuronFileError` property by existence rather than 
  class identity, making it robust to module boundary issues

### `ui/.../asset-upload.tsx`
- Updated the `if (error)` handler in `onFilesAdded` to use `isNeuronFileError` 
  to detect structured validation errors
- Reads `error.neuronFileError.detail` as the specific error message (e.g., 
  "Morphology validation failed: SomaNonConform")
- Falls back to `error.message` for any other error type (network failures, 
  auth errors, file-size rejections)
- The existing `Alert` component renders the two-line error message cleanly:
  - Line 1: "The neuron file ... could not be validated by our tool."
  - Line 2: The specific validation failure reason from the backend

## UI output example

**Before:**
```
File upload error(s)

The neuron file 2004-02-16-C-L23dendax.CNG.swc could not be validated by our tool.
An unknown error occurred
```

**After:**
```
File upload error(s)

The neuron file 2004-02-16-C-L23dendax.CNG.swc could not be validated by our tool.
Morphology validation failed: <morphio._morphio.SomaNonConform object at 0x120bdb830>
```

Or:
```
File upload error(s)

The neuron file invalid_soma.swc could not be validated by our tool.
Soma diameter validation failed: Unrealistic soma diameter detected.
```

## Notes

- The `api-client.ts` change is general-purpose and benefits any future code 
  using `asRawResponse: true`
- Non-validation errors (network failures, auth errors, file-size rejections) 
  are unaffected — `isNeuronFileError` returns false and the handler falls back 
  to `error.message` as before
- The error messages come directly from the backend validation logic, so 
  improving backend error messages automatically improves the frontend UX with 
  no additional frontend changes needed
- The script `batch-register-morphologies.ipynb` has been removed since it has been superceded by the csv-handling gui


src/ui/segments/contribute/cell-morphology/steps/protocol.tsx
```diff
@@ -31,18 +31,25 @@ export function Protocol() {
   const isDigitalReconstruction =
     selectedProtocolGenerationType === CellMorphologyGenerationType.DigitalReconstruction.key;
 
+  const selectedProtocolId = Form.useWatch('cell_morphology_protocol_id', form);
```
tolokoban: This variable is defined, but never used.

tolokoban (commented):
Nice PR Daniel!
Only a minor lint fix.

src/ui/segments/contribute/cell-morphology/steps/protocol.tsx
dkeller9: Thanks, the unused variable has been removed.

src/api/api-client.ts
```diff
@@ -325,7 +325,7 @@ class ApiClient {
         responseData = (await response.arrayBuffer()) as unknown as T;
       }
 
-      if (!response.ok) {
+      if (!response.ok && !config.asRawResponse) {
```
bilalesi: this is breaking change, did you check if everything is working fine by adding this

tolokoban (dismissed):
LGTM

src/api/api-client.ts
dkeller9: Good point, just to be sure I've added a passThroughErrors configuration flag to api-client.ts, to ensure that no other parts are affected.

dkeller9: @bilalesi is this fix okay with you?

src/api/api-client.ts
```diff
@@ -325,7 +317,7 @@ class ApiClient {
         responseData = (await response.arrayBuffer()) as unknown as T;
       }
 
-      if (!response.ok) {
+      if (!response.ok && !config.passThroughErrors) {
```
bilalesi: i don't think we need this `passThroughErrors` at all, 
we already passing the responseData to the `makeApiError` 
please check here https://github.com/openbraininstitute/core-web-app/blob/c1e6dd4a2201f057b21d0ec6b45286bd49b8279f/src/api/utils.ts#L65
you can extend  `ApiErrorCause` to 
```ts
export interface ApiErrorCause {
  code?: string;
  message?: string;
  status?: number;
  details?: any;
  originalError: Error or any;  <—— pass `responseData` here
}
```
and use it later in your api call

src/api/api-client.ts
bilalesi: i commented below

src/api/api-client.ts
dkeller9: @bilalesi good suggestion, originalError has now been added to ApiErrorCause. Is this what you had in mind?

PR #1671, opened 2026-05-20
feat(scan-config): integrate stream-log to microcircuit and whole-bra…

bilalesi:
…in region circuits


src/features/scan-config/use-cases/simulations/results.tsx
```diff
@@ -268,11 +289,17 @@ export default function SimulationsTab({
         duration: 10,
       });
     }
+
+    setSelectedSimulationIds([]);
```
tolokoban, line 293: This line appears twice.
Already at line 244.

src/features/scan-config/use-cases/simulations/results.tsx
```diff
@@ -221,7 +241,8 @@ export default function SimulationsTab({
         }
       }
     }
-
+    setSelectedSimulationIds([]);
+    setSimRequestInProgress(false);
```
tolokoban, line 245: Did you miss to set it to `true` at the beginning of function `runViaLaunchSystem()`?

src/features/scan-config/components/simulation-files/index.tsx
```diff
@@ -39,12 +49,46 @@ export function SimulationFiles({
...
+      }),
+    [simulation.id, jobId]
+  );
+
+  const inputFilesWithLogs = useMemo(() => {
+    return prependLogStreamFile({
+      file: logStreamFiles.input
+        ? makeTaskConfigurationFile({
+            descriptor: logStreamFiles.input,
+            config: simulation as unknown as ITaskConfig<Record<string, unknown>>,
```
tolokoban, line 66: `simulation` is supposed to be a `ISimulation`.
Is it safe to force cast is to `ITaskConfig<Record<string, unknown>>`?

src/features/scan-config/use-cases/build/results.tsx
```diff
@@ -32,14 +33,8 @@ type Props = {
...
-export function BuildTab({
-  campaignOriginAction,
-  campaignId,
-  virtualLabId,
-  projectId,
-  isCampaignIdChanged,
-}: Props) {
-  const context = useMemo(() => ({ virtualLabId, projectId }), [projectId, virtualLabId]);
+export function BuildTab({ campaignOriginAction, campaignId, isCampaignIdChanged }: Props) {
+  const context = useWorkspace();
```
tolokoban, line 37: In Build, you are using `useWorkspace()`, but in Simulation (`src/features/scan-config/use-cases/simulations/results.tsx`) you are getting `virtualLabId` and `projectId` as props.

src/features/scan-config/use-cases/simulations/results.tsx
bilalesi: yeah there is two cases here, some entities are using task-manager, and others using legacy simulator, so this setter was meant to be here as have two different boundaries,  i am not fun mixing the two sims, it just not priority now to refactor it as it bring conflicts too, and i think soon will have all simulations running in the task manager

src/features/scan-config/components/simulation-files/index.tsx
bilalesi: it's safe in a sense that it will be migrated soon, and 
`TActivityCustomFile` accept any `IEntity`, i just satisfy the linter and basically we need only the config_id in the logs items

src/features/scan-config/use-cases/build/results.tsx
bilalesi: yeah that's meant to be migrated too to not pass the context as props, will do it in any future PR

src/features/scan-config/use-cases/simulations/results.tsx
bilalesi: i fixed it, good catch

tolokoban (approved):
Good to go!

bilalesi:
thanks @tolokoban

PR #1672, opened 2026-05-20
Fixed deleted synapses still visible in 3D

tolokoban:
This PR does 3 things:

1. Replaces the tooltip-based delete confirmation with a proper modal dialog ( SynapticInputDeletionConfirmationDialog)
2. Cleans up synapse placement state when removing a synaptic input config
3. Simplifies `neuron-visualizer.tsx` by removing unused imports

It relates to this ticket: https://github.com/openbraininstitute/prod-singlecell-simulation/issues/278

The preview has been validated by @ayimaok and @darshanmandge

### DIFF
--- a/src/ui/segments/workflows/simulate/single-neuron/single-neuron-synaptome/item/synaptic-input-deletion-confirmation-dialog/synaptic-input-deletion-confirmation-dialog.tsx
+++ b/src/ui/segments/workflows/simulate/single-neuron/single-neuron-synaptome/item/synaptic-input-deletion-confirmation-dialog/synaptic-input-deletion-confirmation-dialog.tsx
@@ -1,0 +1,50 @@
+import { Button } from '@/ui/molecules/button';
+import { cn } from '@/utils/css-class';
+
+import styles from './synaptic-input-deletion-confirmation-dialog.module.css';
+
+export interface SynapticInputDeletionConfirmationDialogProps {
+  className?: string;
+  open: boolean;
+  onOpenChange(open: boolean): void;
+  /**
+   * the user has clicked "confirm".
+   */
+  onConfirm(): void;
+}
+
+export function SynapticInputDeletionConfirmationDialog({
+  className,
+  open,
+  onOpenChange,
+  onConfirm,
+}: SynapticInputDeletionConfirmationDialogProps) {
+  const handleClose = () => onOpenChange(false);
+
+  if (!open) return null;
+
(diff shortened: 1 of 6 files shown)


src/ui/segments/workflows/simulate/single-neuron/single-neuron-synaptome/item/synaptic-input-deletion-confirmation-dialog/synaptic-input-deletion-confirmation-dialog.tsx
```diff
@@ -0,0 +1,66 @@
...
+  }, [open]);
+
+  return (
+    <dialog
+      ref={refDialog}
+      className={cn(className, styles.synapticInputDeletionConfirmationDialog)}
+      closedby="any"
+      onClose={handleClose}
+    >
+      <p className="text-primary-9 text-justify text-sm">
```
bilalesi, line 47: why are we using a dialog here?
this will introduce multiple modal styles and behaviors across the app, which will make the ux feel inconsistent and also make maintenance harder over time
i think it’s better to stick to a single modal pattern in the whole app unless there’s a strong technical or ux reason not to

bilalesi: another question, is there a reason why moving from tooltip to modal ?
iirc: I think the tooltip is nearer to the user click per synaptic input rather than open a modal and the user has to move a bit per each deletion to approve

tolokoban: It was something we discussed long ago with Loris and Ayima.
The current issue is that you have two icons/buttons side by side: one for show/hide the synapses, one to remove them.
The first one is clickable, but not the second one.
Every user will try to click on the trashbin and nothing will happen.
They have to hover it, stay few milliseconds still, then try to reach the button that appears in the popup.
I don't know where we have this in the rest of the app. I haven't seen this pattern elsewhere.
And I don't think I have ever seen another website doing this neither.

bilalesi (approved):

tolokoban:
Thanks @bilalesi

PR #1678, opened 2026-05-21
Remove duplicate util/logger, use utils/logger instead

tolokoban:
There was an old `@/util/logger` module that is now duplicated and improved by `@/utils/logger`.
I removed the old one to avoid using it by mistake when auto-completing the imports.

In the process I had to fix some linter issues on the files that I have changed.


bilalesi (approved):
nice!

tolokoban:
Thanks @bilalesi !

PR #1679, opened 2026-05-21
Spike-viewer: round markers, sizing control, smoother steps

pgetta:
## Summary

- **Round markers** — fragment shader now uses a circular SDF instead of the vertical-tick `alphaX`/`alphaY` math (`renderer/webgl-points.ts`).
- **Marker-size slider** — new popover control (`components/raster-plot-controls.tsx`) wired through `useRasterRenderer.setBaseSize`. Popover uses shadow-only styling.
- **Perceptible steps** — dropped `Math.round(pointSize * dpr)`; WebGL now receives the fractional size so each slider step changes the rendered diameter by a sub-pixel amount instead of rounding to the same integer.
- **Minimum stays visible** — point-size formula is `(baseSize + 1) * factor + zoomBonus`, so the smallest slider value still renders ≥1 px on dense datasets (`factor` can drop to 0.5).
- **No horizontal grid lines** — removed the y-tick loop from `axis-overlay.drawGrid`; y-axis ticks/labels are unchanged.

## Test plan

- [x] Open a raster plot (virtual lab simulation result).
- [x] Markers render as round dots, not vertical sticks.
- [x] Drag the marker-size slider across its range; every step produces a visible change.
- [x] At slider = min, dots are still visible on a dense dataset (≥100k spikes).
- [x] Zoom into a dense region — overlapping dots read as solid colour (no pale halo).
- [x] Axis overlay shows vertical grid lines and tick marks/labels on both axes; no horizontal grid lines.
- [x] Hover tooltip still appears near a spike.

### DIFF
--- a/src/features/spike-viewer/renderer/webgl-points.ts
+++ b/src/features/spike-viewer/renderer/webgl-points.ts
@@ -1,36 +1,7 @@
+import FRAGMENT_SHADER from './webgl-points.frag';
+import VERTEX_SHADER from './webgl-points.vert';
+
 import type { SpikePopulation } from '@/features/spike-viewer/spike-trace';
-
-const VERTEX_SHADER = `#version 300 es
-layout(location = 0) in float a_x;
-layout(location = 1) in float a_y;
-uniform vec4 u_bounds; // xMin, yMin, xMax, yMax
-uniform mediump float u_pointSize;
-
-void main() {
-  float x = 2.0 * (a_x - u_bounds.x) / (u_bounds.z - u_bounds.x) - 1.0;
-  float y = 2.0 * (a_y - u_bounds.y) / (u_bounds.w - u_bounds.y) - 1.0;
-  gl_Position = vec4(x, y, 0.0, 1.0);
-  gl_PointSize = u_pointSize;
-}`;
-
-const FRAGMENT_SHADER = `#version 300 es
-precision mediump float;
-uniform vec4 u_color;
-uniform float u_pointSize;
-out vec4 fragColor;
-
-void main() {
-  // Round marker: solid core with ~1 px outer feather so subpixel
-  // jitter shows as opacity falloff at the edge instead of a ±1 px
-  // size step. Keeping the feather *outside* the core preserves a
-  // fully opaque body so dense overlapping dots read as solid colour
-  // rather than accumulating a translucent haze.
-  float feather = 1.0 / max(u_pointSize, 1.0);
-  float dist = distance(gl_PointCoord, vec2(0.5));
-  float alpha = 1.0 - smoothstep(0.5 - feather, 0.5, dist);
-  if (alpha <= 0.001) discard;
-  fragColor = vec4(u_color.rgb, u_color.a * alpha);
-}`;
 
 type PopulationBuffer = {
   name: string;
--- a/src/features/spike-viewer/renderer/webgl-points.ts
+++ b/src/features/spike-viewer/renderer/webgl-points.ts
@@ -20,17 +20,14 @@ uniform float u_pointSize;
 out vec4 fragColor;
 
 void main() {
-  // Vertical tick: solid core with ~1 px outer feather so subpixel
+  // Round marker: solid core with ~1 px outer feather so subpixel
   // jitter shows as opacity falloff at the edge instead of a ±1 px
   // size step. Keeping the feather *outside* the core preserves a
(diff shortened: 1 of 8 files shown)


pgetta:
Relates to [Optimize marker size in spike raster viewer](https://github.com/openbraininstitute/prod-circuit-simulation/issues/171).

src/features/spike-viewer/renderer/webgl-points.ts
```diff
@@ -20,17 +20,14 @@ uniform float u_pointSize;
 out vec4 fragColor;
 
 void main() {
-  // Vertical tick: solid core with ~1 px outer feather so subpixel
+  // Round marker: solid core with ~1 px outer feather so subpixel
```
tolokoban: In `src/webgl.d.ts` we defined two special modules.
This way, you can put your fragment shader in a file `shader.frag`, then import it's string content like this:

```ts
import FRAG from "./shader.frag"
```

There is the same for vertex shaders.
It makes the shader code easier to read because it's not between backticks, and the IDE often knows how to highlight GLSL.

src/features/spike-viewer/renderer/webgl-points.ts
```diff
@@ -20,17 +20,14 @@ uniform float u_pointSize;
...
   // rather than accumulating a translucent haze.
   float feather = 1.0 / max(u_pointSize, 1.0);
-  float dx = abs(gl_PointCoord.x - 0.5);
-  float dy = abs(gl_PointCoord.y - 0.5);
-  float alphaX = 1.0 - smoothstep(0.2, 0.2 + feather, dx);
-  float alphaY = 1.0 - smoothstep(0.5 - feather, 0.5, dy);
-  float alpha = alphaX * alphaY;
+  float dist = distance(gl_PointCoord, vec2(0.5));
+  float alpha = 1.0 - smoothstep(0.5 - feather, 0.5, dist);
   if (alpha <= 0.001) discard;
```
tolokoban: That's a pure detail, but I see you are using a magic number: 0.001

Alpha channel is coded with 256 bits.
So you can check it's nullity with this value:

```glsl
if (alpha < 1.0 / 255.0) ...
```

src/features/spike-viewer/renderer/webgl-points.ts
```diff
@@ -20,17 +20,14 @@ uniform float u_pointSize;
...
-  // fully opaque body so dense overlapping ticks read as solid colour
+  // fully opaque body so dense overlapping dots read as solid colour
   // rather than accumulating a translucent haze.
   float feather = 1.0 / max(u_pointSize, 1.0);
-  float dx = abs(gl_PointCoord.x - 0.5);
-  float dy = abs(gl_PointCoord.y - 0.5);
-  float alphaX = 1.0 - smoothstep(0.2, 0.2 + feather, dx);
-  float alphaY = 1.0 - smoothstep(0.5 - feather, 0.5, dy);
-  float alpha = alphaX * alphaY;
+  float dist = distance(gl_PointCoord, vec2(0.5));
```
tolokoban: If you want a little performance improvement, you can get rid of the square root computation done by function `distance`.

```glsl
vec2 v = 2.0 * (gl_PointCoord - vec2(0.5));
float dist = dot(v, v);
float alpha = 1.0 - smoothstep(0.0, 1.0, dist);
```

Then, you can play with the first argument of `smoothstep` to change the aspect of the dot:

`smoothstep(0.0, 1.0, dist)`

Image

`smoothstep(0.25, 1.0, dist)`

Image

`smoothstep(0.5, 1.0, dist)`

Image

`smoothstep(0.75, 1.0, dist)`

Image

src/features/spike-viewer/renderer/webgl-points.ts
pgetta: Good idea, I've moved the code for both shaders into dedicated files.

src/features/spike-viewer/renderer/webgl-points.ts
pgetta: Should be fixed now

src/features/spike-viewer/renderer/webgl-points.ts
pgetta: Nice, I've dropped sqrt, there is indeed some performance gain!

tolokoban (approved):
Good to go!

pgetta:
Thank you Fabien.

PR #1684, opened 2026-05-22
feat-110: workflows multi selection

bilalesi:
the PR centralizes scan-config workflow configuration in the workflow registry, with bindings, route builders, and schema resolution as a single source of truth.

it adds session backed browse selection, schema-driven single/multi-entity picking, then routes configure through unified `/workflows/{activity}/configure/{type}/{sessionId}` pages instead of per-entity urls

the workflow browse flow uses the new data-table entity selector and listing, supporting multi-select before configure for supported workflows

activities table actions (duplicate, view configuration) and mini-detail/breadcrumb links now resolve configure urls from the registry, including `initialCampaignId` for resuming campaigns

legacy scan-config helpers (`getGeneratedApiUrl`, `getScanConfigSchemaName`, entity query maps) are removed in favor of registry resolution

- all `/configure` pages now uses:  `makeActivityScanConfigConfigureCatchAllPage`
- all `/new` pages now uses: `createWorkflowNewRoutePage` 

mode details here: https://teams.microsoft.com/l/message/19:kRU5l_iuGbBd50PKiXKDDIr7trhAb0nM8_YuJEJAMnA1@thread.tacv2/1779360250552?tenantId=c355c1d8-2b0b-4398-a8ea-b2134d2db027&groupId=ee40e654-3bca-4126-a5f9-020fd958c8b3&parentMessageId=1779360250552&teamName=Web%20application%20devs&channelName=OBI%20web%20application&createdTime=1779360250552

Related to : https://github.com/openbraininstitute/prod-build-ion-channel-model/issues/117


src/features/scan-config/workflow/use-workflow-entity.ts
```diff
@@ -34,33 +37,41 @@ export function useStaticTypeWorkflowEntity(
...
-    queryFn: () => entitySource.query.queryFn({ context: workspace, id: entityId }),
-  });
+  const workflowSessionSelection = readWorkflowSessionSelection(sessionId);
 
-  return {
-    entity,
-    entityType: getSupportedEntityTypesForScanConfiguration({ entity }),
-    entityId: entity.id,
-  };
+  return useMemo(() => {
```
tolokoban, line 61: You are using the `useMemo` hook after a potential `throw` (line 56).
This breaks the React hooks rules that stipulates that the call of hooks must be the same for every render.

src/features/scan-config/workflow/create-page.tsx
tolokoban, line 107: I know this is not some code you changed in this PR, but the linter was complaining about the fact that `options` is not defined in the dependencies.

tolokoban (commented):
Great work, Bilal!
I just have two little comments (one is even out of scope).

src/features/scan-config/workflow/create-page.tsx
bilalesi: the factory below runs only once during module evaluation, when the route exports
```ts
export default makeSimulateCircuitScanConfigPage({...})
```
the returned Page component keeps the same options reference for its entire lifetime, it gets created once by the factory and never changes between renders
the only thing that changes at runtime is `definition` (coming from `searchParams.dataType`), and that’s already included in the dependency array

src/features/scan-config/workflow/use-workflow-entity.ts
bilalesi: I think I answered this before, but here are a few more techincal details
the lint rule cannot distinguish between `throw` and `return`, but they are fundamentally different

when a throw happens during render, react aborts the entire render attempt and propagates the error to the nearest error boundary, no render is committed, which means no hook state is recorded for that render

so the flow is:

* invalid sessionId -> render abort -> error boundary -> no hooks recorded
* valid sessionId -> render completes -> `useMemo` runs consistently on every render

these two states never exist as committed renders at the same time, so React’s hook bookkeeping cannot get de-sync

src/features/scan-config/workflow/create-page.tsx
tolokoban: If it does not change, then it doesn't hurt to have it in the dependency.
This way, if the behavior of `options` change in the future, we don't get an issue.

tolokoban (dismissed):

src/features/scan-config/workflow/create-page.tsx
bilalesi: yes agree, it does not hurt, but in the same time it break the hook dependencies array rule, 
which require to include only what's require for the function to re-run

bilalesi:
@tolokoban can you re-approve please, i had to rebase

tolokoban (approved):

bilalesi:
thanks @tolokoban

PR #1685, opened 2026-05-22
Auth: prevent redirect loop and HTTP 431 on repeat sign-in

pgetta:
## Summary

A redirect loop ending in HTTP 431 (Request Header Fields Too Large) on auth. Three independent issues combined to produce it; this PR fixes all three plus an open-redirect surface noticed during review.

## Changes

**`src/app/app/log-in/page.tsx`**
- Probe `useSession()` and short-circuit straight to the target when already authenticated — no redundant OAuth round-trip.
- Validate `callbackUrl` is same-origin before navigating in the authenticated branch (closes an open-redirect via `/app/log-in?callbackUrl=https://evil`).
- Stop re-wrapping `callbackUrl` when it already targets `/sync`. Re-wrapping added a `?redirectUrl=` layer + re-encoded percents on every iteration, growing `__Secure-next-auth.callback-url` past upstream header limits.

**`src/auth/index.ts` — `redirect` callback**
- Gate the preview auth-proxy detour on presence of `auth-proxy-redirect-target` cookie. Direct sign-ins on the proxy domain no longer get routed through a handler that has nothing to read.
- Reorder checks so cheap predicates short-circuit before `await cookies()`; hoist `authProxyHostname` to module scope.
- Defer `next/headers` to a dynamic import (this module is reachable from client bundles via `auth-fetch`).

**`src/app/api/auth/redirect-after-signin/route.ts`**
- Build the fallback URL from `serverConfig.AUTH_PROXY_URL` instead of `request.url`. Under Amplify Lambda the latter resolves to the function's internal binding (`localhost:3000`), which is why direct logins on `main.preview` were bouncing to localhost.

**`src/app/showcases/build-platform-login-url.ts`**
- Link directly to `/sync`. The log-in page intercepts unauthenticated visits already, so the explicit `?callbackUrl=` wrap layer just contributed to cookie growth.

**`src/auth/constants.ts`**
- Extract `PREVIEW_DOMAIN_SUFFIX` so the cookie scope (client write) and the redirect-target allowlist (server read) can't drift apart.

## Test plan

- [x] Sign in on feature branch routes via `develop.preview`. TODO once the PR is merged: the auth proxy host to be updated to `main.preview`, `develop.preview` - to be retired
- [ ] Sign in directly on auth proxy host works
- [x] Visit `/app/log-in` while already authenticated with `?callbackUrl=/app/foo` → goes to `/app/foo` without OAuth round-trip
- [x] Visit `/app/log-in?callbackUrl=https://evil.example.com` while authenticated → falls back to `/sync` instead of redirecting off-platform
- [x] Repeated re-auth flows don't grow `__Secure-next-auth.callback-url`
- [x] Showcase entry URL (`buildPlatformLoginUrl`) still onboards correctly for new/anonymous users

### DIFF
--- a/src/app/api/auth/redirect-after-signin/route.ts
+++ b/src/app/api/auth/redirect-after-signin/route.ts
@@ -14,13 +14,8 @@ export async function GET() {
     return response;
   }
 
-  // Fall back to the auth-proxy URL from config, not `request.url`. Under
-  // Amplify Lambda, `request.url` reflects the function's internal binding
-  // (e.g. `https://localhost:3000`) rather than the public host, so building
-  // the fallback off the request would redirect the browser to localhost.
-  // `AUTH_PROXY_URL` is guaranteed at runtime on preview (this route is only
-  // reachable through the preview detour), but the schema marks it optional,
-  // so guard the construction.
+  // Build the fallback from config, not `request.url` — under Amplify Lambda
+  // the latter resolves to the function's internal binding, not the public host.
   if (!serverConfig.AUTH_PROXY_URL) {
     return new NextResponse('Auth proxy is not configured', { status: 500 });
   }
--- a/src/app/api/auth/redirect-after-signin/route.ts
+++ b/src/app/api/auth/redirect-after-signin/route.ts
@@ -1,7 +1,7 @@
 import { cookies } from 'next/headers';
 import { NextResponse } from 'next/server';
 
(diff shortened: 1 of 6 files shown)


pgetta:
Relates to [Auth getting in a redirect loop](https://github.com/openbraininstitute/core-web-app/issues/1680).

pgetta:
FYI, @loris-olivier-obi

bilalesi (approved):
lgtm

pgetta:
Thanks!

PR #1689, opened 2026-05-26
Circuit viz optimize fetching

tolokoban:
This PR improved a bit the issues Ayima notices in Safari with very low download times and the impression than the UI is blocked.

To try to mitigate that, we did:

1. **Sequential cell loading** — `SequentialLoader` class queues morphology fetches one at a time, preventing request flooding.
2. **Time-based morphology cache** — Module-level cache with 1-hour TTL per entry (via `setTimeout` garbage collection). Avoids unbounded growth.
3. **Debounced switch** — `DebouncedSwitch` delays `onChange` by 200ms to prevent rapid toggling from triggering expensive re-renders.
4. **Progressive loading UX** — Viewer is mounted immediately with an overlay spinner; cells load progressively instead of blocking on initial load.
5. **Stable cell IDs** — Cell IDs now encode `circuitId`, `showAxon`, and index for correct invalidation on toggle.
6. **Dynamic cell colors** — HSL-based color generation instead of a single hardcoded blue, giving each cell a distinct hue.

### DIFF
--- a/src/features/scan-config/components/circuit-viz/debounced-switch/debounced-switch.tsx
+++ b/src/features/scan-config/components/circuit-viz/debounced-switch/debounced-switch.tsx
@@ -1,0 +1,28 @@
+import { Switch } from 'antd';
+import React from 'react';
+
+export interface DebouncedSwitchProps {
+  className?: string;
+  value: boolean;
+  onChange(value: boolean): void;
+  onClick(): void;
+}
+
+export default function DebouncedSwitch({
+  className,
+  value,
+  onChange,
+  onClick,
+}: DebouncedSwitchProps) {
+  const refId = React.useRef<NodeJS.Timeout>(null);
+  const [internalValue, setInternalValue] = React.useState(value);
+  const handleChange = (value: boolean) => {
+    onClick();
+    setInternalValue(value);
+    const id = refId.current;
+    if (id !== null) globalThis.clearTimeout(id);
+    refId.current = globalThis.setTimeout(() => onChange(value), 200);
+  };
+
+  return <Switch className={className} value={internalValue} onChange={handleChange} />;
+}
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@bprogress/next": "^3.2.12",
     "@monaco-editor/react": "^4.7.0",
     "@octokit/core": "^7.0.6",
-    "@openbraininstitute/morphoviewer": "^0.25.2",
+    "@openbraininstitute/morphoviewer": "^0.25.4",
     "@portabletext/react": "^6.2.0",
     "@radix-ui/react-accordion": "^1.2.12",
     "@radix-ui/react-checkbox": "^1.3.3",
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "memoize-one": "^6.0.0",
     "moment": "^2.30.1",
     "motion": "^12.38.0",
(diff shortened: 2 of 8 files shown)


src/features/scan-config/components/circuit-viz/debounced-switch/debounced-switch.module.css
```diff
@@ -0,0 +1,2 @@
+.debouncedSwitch {
+}
```
bilalesi: this class is empty

src/features/scan-config/components/circuit-viz/debounced-switch/debounced-switch.tsx
```diff
@@ -0,0 +1,37 @@
+import { Switch } from 'antd';
+import React from 'react';
+
+import { classNames } from '@/util/utils';
+
+import styles from './debounced-switch.module.css';
```
bilalesi: i think no need for this one

src/features/scan-config/components/circuit-viz/debounced-switch/debounced-switch.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const handleChange = (value: boolean) => {
+    onClick();
+    setInternalValue(value);
+    const id = refId.current;
+    if (id !== null) globalThis.clearTimeout(id);
+    refId.current = globalThis.setTimeout(() => onChange(value), 200);
+  };
+  return (
+    <Switch
+      className={classNames(className, styles.debouncedSwitch)}
```
bilalesi: what if: e.g: user toggles the switch and immediately change the page (or parent conditionally unmounts) within 200ms, the timeout still triggers `onChange(value)` on an unmounted tree, which  causes state update on unmounted component, probably you will see some warnings/error in the console  

I prefer useEffect for cleaning up the timeout (which probably need a new implementation for the entire componentn)

bilalesi: you passed "value" as prop and as default value for the internal state, does the parent also control the component ?

src/features/scan-config/components/circuit-viz/debounced-switch/debounced-switch.module.css
tolokoban: Good catch!
I removed it.

src/features/scan-config/components/circuit-viz/debounced-switch/debounced-switch.tsx
tolokoban: Indeed!
I removed it.

src/features/scan-config/components/circuit-viz/debounced-switch/debounced-switch.tsx
tolokoban: This one is a tricky one...

I wrote it in many different ways and always end up with a but in the asynchronous flow of loading cells and displaying them.
But I tried to put `2000 ms` instead of `200 ms` to be able to test the quick tab switch before the timeout goes, and I got no error in the console.

This PR is blocking the freeze, unfortunately.
So, for now, I will keep this switch implemented as it is to unblock the situation, and I hope I will find time later to see what's going on really here.

bilalesi (approved):

tolokoban:
Thanks a lot @bilalesi !

PR #1691, opened 2026-05-27
Circuit download: per-component archives, config files, JSON metadata

pgetta:
## Summary

Extends the Circuit Download panel so users can pull individual circuit artifacts instead of always grabbing the full circuit tarball.

## What's new

- **Configuration files** — `circuit_config.json`, node sets, ID mapping as single-file downloads via the existing presigned-URL path. JSON/text MIME types are routed through a fetch-blob shim so the browser saves them instead of rendering inline.
- **Electrical models & mechanisms** — on-the-fly `tar.gz` archives streamed by a new `asset-folder` ticket flow that shares the ticket store, auth, and gzip wiring with the existing entity-batch pipeline.
- **JSON metadata** — stand-alone client-side dump of the `ICircuit` payload.

## Backend changes

- `DownloadTicket` becomes a discriminated union (`entity-batch` | `asset-folder`); existing entity-batch routes unchanged.
- New routes: `POST /api/entity-download/asset-folder/ticket`, `GET /api/entity-download/asset-folder/ticket/[ticketId]`.
- New generator `getAssetFolderFiles` lists assets under one folder prefix and streams them through the shared `tar-stream + createGzip` pipeline.
- `GET /api/entity-download/presigned-url` now validates its query params with Zod (entity type enum, UUIDs, non-empty asset path) instead of trusting raw search params.

## UI changes

- New `ComponentsConfig` section in the download panel, plus a `FolderConfigItem` row visual aligned with the existing network/morphology rows.
- React Query dedupes the new `resolveCircuitConfigAndDirectory` call with the one already issued by `NetworkAndMorphologyConfig` (same key), so no extra network round-trips.

## Notable fallbacks

- ID mapping: falls back to root `id_mapping.json` when `components.provenance.id_mapping` isn't set but the file exists in the directory listing.
- Mechanisms: falls back to `/mod` when `components.mechanisms_dir` is missing (per prod-build-circuit#32).

## Tooling

- Adds a `:NODE:` import group to `biome.json` so `node:`-prefixed built-ins sort above external packages.

## Test plan

- [x] Open a circuit's Download panel: Configuration files (3 rows), Electrical models (≥1), Mechanisms (1), and Metadata button render
- [x] Single-file rows download via presigned URL (verify JSON/text save, not render)
- [x] Folder rows produce a valid `.tar.gz`; `tar -tzf <file> | wc -l` matches the directory listing under the prefix
- [x] Metadata button downloads `<circuit-name>-metadata.json` with the full circuit payload
- [x] Regression: existing Nodes / Edges / Morphologies / Connectivity rows and entity-batch tarball downloads still work

### DIFF
--- a/src/features/entity-download/ticket-store.ts
+++ b/src/features/entity-download/ticket-store.ts
@@ -1,10 +1,11 @@
 import { randomUUID } from 'crypto';
 
-import type { TEntityTypeDict } from '@/api/entitycore/types';
-import type { PartialBy } from '@/types/common';
 import { logError, logInfo } from '@/utils/logger';
 
-type DownloadTicket = {
+import type { TEntityTypeDict } from '@/api/entitycore/types';
+
+export type EntityBatchDownloadTicket = {
+  kind: 'entity-batch';
   entityType: TEntityTypeDict;
   virtualLabId?: string | null;
   projectId?: string | null;
@@ -12,7 +13,23 @@ type DownloadTicket = {
   createdAt: number;
 };
 
-type DownloadTicketRequest = PartialBy<DownloadTicket, 'createdAt'>;
+export type AssetFolderDownloadTicket = {
+  kind: 'asset-folder';
+  entityType: TEntityTypeDict;
+  entityId: string;
+  assetId: string;
+  prefix: string;
+  filename: string;
+  virtualLabId?: string | null;
+  projectId?: string | null;
+  createdAt: number;
+};
+
+export type DownloadTicket = EntityBatchDownloadTicket | AssetFolderDownloadTicket;
+
+type DownloadTicketRequest =
+  | Omit<EntityBatchDownloadTicket, 'createdAt'>
+  | Omit<AssetFolderDownloadTicket, 'createdAt'>;
 
 // Ticket expiration time in milliseconds (60 seconds)
 const TICKET_EXPIRATION_MS = 60 * 1000;
@@ -53,7 +70,7 @@ class TicketStore {
 
     // Generate a unique ticket ID
     const ticketId = randomUUID();
-    const newTicket = { ...ticket, createdAt: Date.now() };
+    const newTicket = { ...ticket, createdAt: Date.now() } as DownloadTicket;
     this.tickets.set(ticketId, newTicket);
 
     return ticketId;
--- a/src/features/entity-download/ticket-store.ts
+++ b/src/features/entity-download/ticket-store.ts
@@ -1,4 +1,4 @@
-import { randomUUID } from 'crypto';
+import { randomUUID } from 'node:crypto';
 
 import { logError, logInfo } from '@/utils/logger';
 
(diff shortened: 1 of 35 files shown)


pgetta:
Relates to [Download options of all circuit components in Data](https://github.com/openbraininstitute/prod-explore-functionality/issues/495).

src/features/entity-download/ticket-store.ts
```diff
@@ -53,7 +70,7 @@ class TicketStore {
 
     // Generate a unique ticket ID
     const ticketId = randomUUID();
-    const newTicket = { ...ticket, createdAt: Date.now() };
+    const newTicket = { ...ticket, createdAt: Date.now() } as DownloadTicket;
```
tolokoban: `ticket` is a `DownloadTicketRequest` which is a `DownloadTicket` without a `createAt` attribute. So, we don't need the `as DownloadTicket`.

src/ui/segments/explore/circuit/elements/download-panel/index.tsx
tolokoban, line 28: This is outside of the scope of this PR,
but the linter complains about this "non-null assertion".
Same on the next line.

src/ui/segments/explore/circuit/elements/download-panel/folder-config-item.tsx
```diff
@@ -0,0 +1,160 @@
...
+        )}
+      >
+        <Tooltip title="Uncompressed total. The downloaded .tar.gz archive will be smaller.">
+          <div className="cursor-help">
+            {entry.totalSize ? `~${formatBytes(entry.totalSize)}` : renderEmptyOrValue(null)}
+          </div>
+        </Tooltip>
+        <div>{renderEmptyOrValue(mimeType)}</div>
+        <Button
+          onClick={onClick}
```
tolokoban, line 85: The `onClick()` function is asynchronous.
I don't know how long can take the `createAssetFolderDownloadTicket` call,
but, in the meantime, the button is still enabled and the use has no notice that their click has been taken into consideration. So they could click again.

Is that an issue, or can we just ignore this point?

tolokoban (commented):
Great job on this PR, @pavlo!
I just left few questions, but none of them are preventing the approval.
Just check if they are all false alarms.

src/features/entity-download/ticket-store.ts
pgetta: Yeah, this type assertion is not needed indeed. I've removed it.

src/ui/segments/explore/circuit/elements/download-panel/index.tsx
pgetta: Yeah, makes sense. I'll be creating separate PRs to resolve all the remaining linting issues bit by bit if you don't mind.

src/ui/segments/explore/circuit/elements/download-panel/folder-config-item.tsx
pgetta: That's a good question.

The POST request is not doing any I/O on the server side, it's just storing an object in the memory, so at the end the delay equals to the round-trip network latency which is pretty small: tens of up to couple of hundred ms. The worst that can happen - a double download.

Probably worth addressing later, but I would leave it as is for the time being if that's fine with you.

tolokoban (approved):
LGTM!!

PR #1694, opened 2026-05-28
Fix single neuron simulator regression

g-bar:


src/features/scan-config/components/ui-elements/ion-channel-variable-modification/shared/editor.tsx
```diff
@@ -149,7 +149,6 @@ function SectionConfigRow({
       >
         <input
           data-scan-config-block-element-item={`${uiElement}_editor_item_input`}
-          type="number"
```
g-bar, line 152: This is needed because if type is 'number' and user inputs an invalid number the onChange below doesn't get triggered, so no error shows and it looks buggy.

Like this the error is shown.

tolokoban (approved):

src/features/scan-config/components/hooks/use-scan-configuration.ts
```diff
@@ -150,7 +150,7 @@ export function useScanConfiguration({
     if (schemaMappingKey) {
       return schemaMappingKey;
     }
-
+    // TODO: WHY DO WE NEED THIS?: THERE SHOULD BE A SINGLE SOURCE OF TRUTH (THE WORKFLOW CONFIG)
```
bilalesi, line 153: good point, is fixed in the multi-selection PR: here https://github.com/openbraininstitute/core-web-app/blob/e87a03d8b8c521887f65d6bdf1bbb4437f12d5b1/src/features/scan-config/components/hooks/use-scan-configuration.ts#L123

PR #1699, opened 2026-06-01
Better chat input

BoBer78:


BoBer78:
Small comments for reviewers. 

The chat input bar now now can span multiple lines, and the attachment button has been moved to the left : 
image
When the input becomes too long, a second line appears : 
image
Finally if the user has a HUGE prompt, the textArea becomes scrollable : 
image



Here is with image / file uploads : 
image
And then in the chat : 
image

ayimaok:
c'est vraiment cool @BoBer78

src/features/ai-assistant/message-item/pending-user-message.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const pdfFiles = files.filter((f) => !f.type.startsWith('image/'));
+
+  return (
+    <div className={messageStyles.user}>
+      <div className={messageStyles.userContent}>
+        {imageFiles.length > 0 && (
+          <div className={messageStyles.userImageRow}>
+            {imageFiles.map((file) =>
+              file.uploaded ? (
+                <img
```
bilalesi, line 24: can't we use the nextjs/image component, or ANTD image (the latter have preview image renderer so the user can zoom, rotate images)

bilalesi (commented):
tiny comment, lemme know if it's not needed; i will approve it

src/features/ai-assistant/message-item/pending-user-message.tsx
BoBer78: So this component is only loader, it often only appears on screen for .5 seconds. 
--> Not sure if it is necessary to change. 

Once the image is loaded, I used the custom `ExpandableImage` component. 
Since we use it everywhere in the chat, I wanted to be consistant. 
You tell me if we should replace it.

bilalesi (approved):

PR #1708, opened 2026-06-04
Prod circuit simulation/132/big circuits as points clouds

tolokoban:
image


src/api/entitycore/types/entities/circuit.ts
```diff
@@ -98,7 +98,13 @@ export interface ICircuit
     Timestamps,
     EntityCoreOwnership,
     EntityCoreType,
-    EntityCoreBaseAsset {}
+    EntityCoreBaseAsset {
+  subject?: {
+    species?: {
+      name?: string;
```
bilalesi: there is the type already: import type { ISubject } from '@/api/entitycore/types/shared/global';

src/features/scan-config/components/model-preview/large-circuit-preview/hooks.ts
```diff
@@ -0,0 +1,83 @@
...
+  React.useEffect(() => {
+    const context = { virtualLabId, projectId, id, type };
+    setNodes(undefined);
+    loadCircuitNodes(context)
+      .then(setNodes)
+      .catch((ex) => {
+        setNodes(ex instanceof Error ? ex : new Error(String(ex)));
+      });
+  }, [virtualLabId, projectId, id, type]);
+
```
bilalesi, line 22: why not `useQuery` ?

src/features/scan-config/components/model-preview/large-circuit-preview/hooks.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (species.includes('homo')) return SOMA_RADII.human;
+    if (species.includes('rattus')) return SOMA_RADII.rat;
+    if (species.includes('musculus')) return SOMA_RADII.mouse;
+    if (species.includes('aquarana')) return SOMA_RADII.bullfrog;
+    if (species.includes('cricetulus')) return SOMA_RADII.hamster;
+    if (species.includes('catus')) return SOMA_RADII.cat;
+    if (species.includes('human-mouse')) return SOMA_RADII.hybridHumanMouse;
+    if (species.includes('loligo')) return SOMA_RADII.squid;
+    if (species.includes('xenopus')) return SOMA_RADII.clawedFrog;
+    if (species.includes('drosophila')) return SOMA_RADII.fly;
```
bilalesi: can be an mapping array or a Map, instead of IFs

src/features/scan-config/components/model-preview/large-circuit-preview/hooks.ts
```diff
@@ -0,0 +1,83 @@
...
+  projectId: string;
+  id: string;
+  type: TEntityTypeDict;
+}
+
+async function loadCircuitNodes(context: Context) {
+  const { virtualLabId, projectId, id, type } = context;
+  const loader = new ServiceSonataLoader({ virtualLabId, projectId, modelId: id, modelType: type });
+  try {
+    const populationBeforeFiltering = await loader.asyncPopulations;
```
bilalesi, line 76: if you move to useQuery, i would suggest to instantiate the class in the upper scope and just pass the loader to the caller function

src/features/scan-config/components/model-preview/large-circuit-preview/hooks.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (species.includes('loligo')) return SOMA_RADII.squid;
+    if (species.includes('xenopus')) return SOMA_RADII.clawedFrog;
+    if (species.includes('drosophila')) return SOMA_RADII.fly;
+    return SOMA_RADII.unknown;
+  }, [circuit]);
+}
+
+interface Context {
+  virtualLabId: string;
+  projectId: string;
```
bilalesi, line 67: those are the values of  `WorkspaceContext` so `interface IContext extends WorkspaceContext { ...others }`,

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
```diff
@@ -0,0 +1,97 @@
...
+  morphologyId: string;
+  position: [number, number, number];
+}
+
+export function LargeCircuitPreview({ className, circuit }: LargeCircuitPreviewProps) {
+  const somaRadius = useSomaRadius(circuit);
+  const nodes = useCircuitNodes(circuit);
+  const handleDownload = () => {
+    const blob = new Blob([JSON.stringify(nodes)], { type: 'application/json' });
+    saveAs(blob, `${circuit.id}.json`);
```
bilalesi, line 33: what if we add the name as prefix too ?

src/utils/logger.ts
```diff
@@ -1,24 +1,13 @@
...
-const logColors = {
-  log: '\x1b[32m',
-  error: '\x1b[31m',
-  warn: '\x1b[33m',
-  debug: '\x1b[34m',
-  trace: '\x1b[35m',
-  info: '\x1b[36m',
-};
-const resetColor = '\x1b[0m';
-
```
bilalesi, line 13: is there a reason why we removed those colors ?

tolokoban: They are not interpreted by the browser's console.

src/features/scan-config/components/model-preview/large-circuit-preview/hooks.ts
tolokoban: Pavlo has a special cache for SONATA files. I'm waiting for him to make it reusable, because right now it's use in the tables but it's not yet independent to be re-used.

So React queries will just add another useless level of caching.

src/features/scan-config/components/model-preview/large-circuit-preview/hooks.ts
tolokoban: Pavlo has a special cache for SONATA files. I'm waiting for him to make it reusable, because right now it's use in the tables but it's not yet independent to be re-used.

So React queries will just add another useless level of caching.

src/utils/logger.ts
bilalesi: actually it does, e.g. this is from the progress bar when loading a circuit 
Screenshot 2026-06-19 at 10 24 33

src/features/scan-config/components/model-preview/large-circuit-preview/hooks.ts
tolokoban: Fixed it.
Not sure it's a great improvement, though...

src/features/scan-config/components/model-preview/large-circuit-preview/hooks.ts
tolokoban: Not sure about this one.
Here, I am declaring what is needed for my function.
If I start extending `WorkspaceContext` and at some point someone add an attribute to it, that would made my function need a new attribute that is actually not needed.

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
tolokoban: This would end up in very long names sometimes. And it would need to be sanitazed to remove all `/` and `.`.
Moreover, James is not sure he wants to keep this download feature.

src/api/entitycore/types/entities/circuit.ts
tolokoban: Fixed it.

src/utils/logger.ts
tolokoban: I think it only work with some browsers.
Here is what I get with Firefox:

image

If we really need to use colors in the console, the way to go is to use CSS styling, like this: https://developer.mozilla.org/en-US/docs/Web/API/console#styling_console_output

src/features/scan-config/components/model-preview/large-circuit-preview/hooks.ts
pgetta: Will address that most likely on Monday.

tolokoban (commented):
A comment to submit my other comments.

src/features/scan-config/components/model-preview/large-circuit-preview/hooks.ts
bilalesi: here you are not extending the `WorkspaceContext`; you are extending your context 
```ts
interface Context (this you already declared it) extends WorkspaceContext  (this is in shared types) { 
      ...others // these are the remaining from your Context
}
```

src/features/scan-config/components/model-preview/large-circuit-preview/hooks.ts
bilalesi: if you think it's not great then please bring it back, i thought use a map is better, but you knows better

g-bar (approved):

src/utils/logger.ts
bilalesi: mmm, got it, i useually use chromium browsers and it works, but it's fine to delete them

bilalesi (approved):

tolokoban:
Thank @bilalesi and @g-bar !

PR #1709, opened 2026-06-05
With white

jdcourcol:
1- Detail view with Blue background
2- Image / visualization always on top (interactive / smaller viewer selected by default)
3- Name + description at the very top

grafik


jdcourcol:
ah, now it is asking me to create a virtual lab. I think we need to fetch some changes ?

tolokoban (dismissed):

bilalesi (approved):

PR #1724, opened 2026-06-09
Prod singlecell simulation/285/single neuron simulate button gives error

tolokoban:
This error is very hard to reproduce (https://github.com/openbraininstitute/prod-singlecell-simulation/issues/285).
Because it seems to be related to the cache.

In any case, the issue is in the `getMEModel()` function when used with `useSuspensedQuery()`.
Sometimes, it can return `null`. But after a refresh of the browser, the error does not appear anymore.

The fix I found, for now, is to check for this `null` and throw an error. Then, I have added a retry in `useSuspensedQuery()` and I was not able to see the error again.

```ts
queryFn: async () => {
  const result = await getMEModel({ id: modelId, context: { virtualLabId, projectId } });
  if (!result) throw new Error('ME-Model not found!');

  return result;
},
retry: 2,
```

## Warning!

The correct URL for the preview is this one:
https://prod-singlecell-simulation-285-single-neuron-simulate-button-gi.preview.openbraininstitute.org/


src/app/app/virtual-lab/[virtualLabId]/[projectId]/workflows/simulate/configure/memodel/[id]/page.tsx
```diff
@@ -36,13 +35,9 @@ export default function Page({
...
 >) {
-  const { virtualLabId, projectId, id: modelId } = use(pathParams);
   const sessionId = useLegacyWorkflowSessionFromSearchParams(searchParams);
-
-  const { data: entity } = useSuspenseQuery({
-    queryKey: keyBuilder.meModel({ virtualLabId, projectId, entityId: modelId }),
-    queryFn: () => getMEModel({ id: modelId, context: { virtualLabId, projectId } }),
-  });
+  const entity = useEntity(pathParams);
+  if (!entity) return <Spinner />;
```
Comment (bilalesi) [inline src/app/app/virtual-lab/[virtualLabId]/[projectId]/workflows/simulate/configure/memodel/[id]/page.tsx:40]: this will keep the spinner for ever if the entity does not exists at all

src/app/app/virtual-lab/[virtualLabId]/[projectId]/workflows/simulate/configure/memodel/[id]/hooks.ts
```diff
@@ -0,0 +1,41 @@
...
+import { query } from '@/components/documentation/query/features-item-hooks.groq';
+import { keyBuilder } from '@/ui/use-query-keys/data';
+
+import type { IMEModel } from '@/api/entitycore/types';
+
+export function useEntity(
+  params: Promise<
+    {
+      virtualLabId: string;
+      projectId: string;
```
Comment (bilalesi) [inline src/app/app/virtual-lab/[virtualLabId]/[projectId]/workflows/simulate/configure/memodel/[id]/hooks.ts:14]: i think we can bypass the issue without the new hook,
i think the reason is we don't have a suspense boundary for the useSuspenseQuery
in the layout we are prefetching with hydrating to client and in the page we are suspensing without suspense boundary in the layout

Comment (tolokoban) [reply src/app/app/virtual-lab/[virtualLabId]/[projectId]/workflows/simulate/configure/memodel/[id]/hooks.ts:14]: I think you may be right, but for now I would prefer to keep it like this and create another ticket to improve it.

This bug was appearing rarely and therefore was hard to reproduce.
With the current code, at least, nor Darshan, Ilkan or Ayima couldn't reproduce it.
And Darshan is pushing hard to have this in prod as soon as possible...

bilalesi (approved):

tolokoban:
Thanks @bilalesi !

I have created this ticket to address the technical debt this fix has added:
https://github.com/openbraininstitute/prod-singlecell-simulation/issues/291

PR #1731, opened 2026-06-11
290/maximum call stack

tolokoban:
Google Chrome has a lower tolerance for deep recursivity.
With complex morphologies, we hit the maximum call stack size.

To mitigate this, we have refactored several resursive functions to make them use the heap now.


src/components/neuron-viewer/debug-panel/debug-panel.tsx
```diff
@@ -61,7 +61,7 @@ export function DebugPanel({ className, morphology, synapses }: DebugPanelProps)
 function useDebugMode(): boolean {
   const [debugMode, setDebugMode] = React.useState(false);
   React.useEffect(() => {
-    const item = globalThis.localStorage.getItem('@bbp/morphoviewer:debug');
+    const item = globalThis.localStorage.getItem('@openbraininstitute/morphoviewer:debug');
```
pgetta, line 64: Localstorage is already scoped to the origin so the `@openbraininstitute` part is redundant. I would just drop the prefix entirely.

pgetta (dismissed):

src/components/neuron-viewer/debug-panel/debug-panel.tsx
pgetta: But that's not a blocker.

tolokoban: Yes, you're right about the scoping.

Here, it's more the name of the library as defined in the `package.json` that we used as scoping. And I can use the same in MorphoViewer. So, this way, I'm sure its not colliding with something else in core-web-app.

pgetta (approved):

tolokoban:
Thanks again @pgetta  !

PR #1736, opened 2026-06-11
feat(circuit-nodes-table): download progress bar

pgetta:
## Summary

The circuit nodes table downloads a `nodes.h5` file (up to ~1 GB) into a web worker before rendering. Today it shows only an indeterminate spinner, so during a ~minute-long download users can't tell whether it's working or stuck. This replaces the spinner with a real horizontal progress bar showing percentage and downloaded/total MB.

## Changes

- **`worker/fetch-and-cache.ts`** — stream the live network response into the Emscripten FS while `tee()`-ing a second branch into CacheStorage, so progress reflects the actual download instead of a fast local cache read. Emits are throttled (per-percent, or ~2 MB when `Content-Length` is absent). Caching is best-effort (a failed write just re-downloads next session). **Progress is reported only on a genuine network fetch** — cache hits read fast enough that a bar would just flicker.
- **`worker/nodes.worker.ts`** — `open()` forwards an optional `onProgress` callback.
- **`hooks/use-nodes-worker.ts`** — passes a Comlink-proxied progress callback; exposes `{ received, total }` state guarded by the existing generation/cancel checks; resets on load/ready/error/teardown.
- **`circuit-nodes-table.tsx`** — `DownloadProgress` renders an antd `Progress` line bar (`var(--color-primary-6)`) with live MB counts, falling back to the spinner before the first byte or when no `Content-Length` is available.

## Verification

- [x] First load (network throttled): bar fills 0→100% with live MB counts
- [x] Reload same population: cache hit → quick spinner, no flicker
- [ ] Switch populations mid-download: progress resets cleanly
- [x] `obi-circuit-h5-v1` cache still populated after a fresh download

## Notes

- Live percentage depends on the asset response including `Content-Length` (presigned S3 GETs normally do; it's CORS-safelisted). Without it, the UI shows a MB-downloaded spinner fallback.

### DIFF
--- a/src/features/circuit-nodes/circuit-nodes-table.tsx
+++ b/src/features/circuit-nodes/circuit-nodes-table.tsx
@@ -174,7 +174,7 @@ function renderBody({
 function CenteredSpin({ label }: { label: string }) {
   return (
     <div className={styles.centered}>
-      <Spin tip={label} size="large">
+      <Spin tip={label} size="large" className="[&_.ant-spin-dot-item]:bg-primary-6!">
         <div className={styles.spinBox} />
       </Spin>
     </div>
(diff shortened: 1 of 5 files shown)


pgetta:
Relates to [Circuit nodes table: show download progress instead of a spinner](https://github.com/openbraininstitute/prod-explore-functionality/issues/508).

tolokoban (approved):

pgetta:
Thanks!

PR #1746, opened 2026-06-15
fix: recover logs stream for supported simulations as task

bilalesi:


src/entity-configuration/domain/simulation/status-utils.ts
```diff
@@ -17,6 +17,14 @@ export function hasSimulationConfigAsset(simulation: SimulationLike) {
   );
 }
 
+export function getLatestSimulationExecution<T extends { creation_date: string }>({
+  executions,
+}: {
+  executions: T[];
+}): T | undefined {
+  return [...executions].sort((a, b) => a.creation_date.localeCompare(b.creation_date)).at(-1);
```
tolokoban, line 25: Using `sort` lead to a complexity of `o(n.log(n))` plus the slicing of the `executions` array, so `o(2n.log(b))`.
With a `reduce` you could have done it in `o(n)`.

tolokoban (approved):
LGTM

PR #1756, opened 2026-06-17
feat(em-cell-mesh): add Version & Dataset columns and Version filter

pgetta:
## Summary
Implements the EM mesh (EMCellMesh) Explore → Data table changes from openbraininstitute/prod-explore-functionality#504:

- Columns are now **Name · Brain region · Species · Version · Dataset · Registration date**, replacing the previous **Mesh type** and **Contributors** columns. (Those field definitions are kept — shared by other views — only this view's column list changed.)
- **Version** (from `release_version`): sortable column + a multi-select checklist filter sourced from server facets (`release_version__in`).
- **Dataset**: sortable column. The dataset name isn't in the list response (the nested dataset is serialized as `{ id, type }`), so it's fetched client-side by id via a new `DatasetNameCell`, deduped across rows by react-query — a page of meshes sharing one dataset issues a single request. Mirrors the existing `ActivityStatusCell` lazy-cell pattern.

## Backend dependency
The **Version checklist filter** depends on an entitycore change that is already deployed to staging :
- add `release_version__in: list[int]` to `EMCellMeshFilter`
- expose a `release_version` facet (scalar-column facet: getattr-guarded id + String-cast label + no-op `filter_joins` entry; add `"release_version"` to `facet_keys`)

Closes (tracking ticket, cross-repo): https://github.com/openbraininstitute/prod-explore-functionality/issues/504


src/ui/segments/data-table/elements/listing-filter-panel/hooks/create-filter-item-component.tsx
```diff
@@ -77,6 +78,12 @@ export function createFilterItemComponent(
           value: label,
         }));
 
+        // Honor a field's opt-in facet ordering (e.g. numeric for release versions).
+        // Without it, options render in the backend's order (alphanumeric for scalars).
+        if (getFieldDefinition(filter.field)?.sortFacetOptions === 'numeric') {
+          facetItems.sort((a, b) => Number(a.value) - Number(b.value));
+        }
+
```
bilalesi, line 86: i don't think a good idea to mix FE sorting in the same table of backend sorting, this will lead to confusion for the user, we can discard the sort by this field until it fixed in the entitycore

src/features/entities/em-cell-mesh/dataset-name-cell.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const context = { virtualLabId, projectId };
+
+  const { data, isLoading } = useQuery({
+    queryKey: keyBuilder.emDenseReconstructionDataset({ id: datasetId ?? '', context }),
+    queryFn: () => getEmDenseReconstructionDataset({ id: datasetId ?? '', context }),
+    enabled: Boolean(datasetId),
+  });
+
+  if (!datasetId) return <>{EmptyValue}</>;
+  return <>{isLoading ? '…' : (data?.name ?? EmptyValue)}</>;
```
bilalesi: may be a skeleton rather then 3-dots

src/ui/segments/data-table/elements/listing-filter-panel/hooks/create-filter-item-component.tsx
pgetta: This is for facet options, not for the table values.

If you don't mind - I'll address that separately (will need to extend entitycore to apply different ordering based on the data type and removing redundant mechanism in the frontend).

Without current FE ordering the user might see in the filter dropdown something like:
* 1
* 11
* 2
* 27
* 3

Which is a bit more confusing than:
* 1
* 2
* 3
* 11
* 27

bilalesi (dismissed):

src/features/entities/em-cell-mesh/dataset-name-cell.tsx
pgetta: Good point, I've added Skeleton, looks indeed better (and more consistent with existing codebase).

bilalesi (dismissed):

pgetta:
Thanks!

bilalesi (approved):

PR #1757, opened 2026-06-18
feat: e2e tests

bilalesi:
this PR adds the e2e test setup and configuration, the tests will run automatically for every PR in a dedicated workflow

the e2e suite uses a dedicated test user that creates its own virtual lab and project during execution, and cleans them up during teardown once the test run is complete

for now, this PR includes a first set of tests covering:

* Data → Listing page navigation and behavior
* Brain Region hierarchy view

additional test coverage will be added in follow-up PRs.


package.json
```diff
@@ -146,6 +149,7 @@
     "@vitest/coverage-v8": "^4.1.8",
     "autoprefixer": "^10.5.0",
     "check-barrel-import-violation": "^0.3.0",
+    "dotenv": "^17.4.1",
```
pgetta, line 152: Is this used anywhere?

pgetta (approved):
Nice work Bilal!

I've left one small comment.

package.json
bilalesi: mm, it's leftover, i was using it to load the envs but then i separate the test envs into `env.test` file

bilalesi: i will remove it in the next PR with few other tests

PR #1760, opened 2026-06-18
fix sentense case in credits history

bilalesi:


tolokoban (approved):
An easy one to review ;-)

PR #1767, opened 2026-06-22
[CHAT] implement HIL

BoBer78:


BoBer78:
Small comment for reviewers. We added 2 tools that support / need approval. These are the "run-python" and "execute-shell" tools. 

Here is a simple query you can test : 
```
What packages are available on the OBI sandbox  ?
Use python to figure it out
```
Be aware that you need credits to start the sandbox. There are some known issue, for example the first tool call often fails. See : https://github.com/openbraininstitute/prod-ai/issues

Here is what it should look like : 
image

Feel free to critique the UI.

jankrepl:
Looks great:)

jankrepl (dismissed):

jankrepl (approved):

PR #1768, opened 2026-06-22
migrate from clsx+tailwind-merge to cnfast

bilalesi:
HLYhuOJa0AA8NNQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "chroma-js": "^3.2.0",
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
+    "cnfast": "^0.0.8",
     "comlink": "^4.4.2",
     "date-fns": "^4.1.0",
     "dayjs": "^1.11.20",
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -11,5 +11,6 @@ minimumReleaseAgeStrict: true
 minimumReleaseAgeExclude:
   - '@openbraininstitute/*'
   - '@tolokoban/tgd'
+  - cnfast@0.0.8
 blockExoticSubdeps: true
 verifyDepsBeforeRun: error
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -67,6 +67,7 @@
     "chroma-js": "^3.2.0",
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
+    "cnfast": "^0.0.8",
```
pgetta, line 70: Do we need also to delete tailwind-merge?

pnpm-workspace.yaml
```diff
@@ -11,5 +11,6 @@ minimumReleaseAgeStrict: true
 minimumReleaseAgeExclude:
   - '@openbraininstitute/*'
   - '@tolokoban/tgd'
+  - cnfast@0.0.8
```
pgetta, line 14: Not sure it's worth to add this exception right away. Especially given the fact that the very first release of this package was 5 (!) days ago. Why do we need to switch to such cutting edge module instead of a proven one, even thought it's a bit slower (in absolute terms) ?

pgetta (commented):
Hi Bilal, while the speed comparison looks good I'm somehow hesitant to make this switch given the maturity of the module.

What's the necessity?

pnpm-workspace.yaml
bilalesi: you are right, i will remove it from the list

package.json
bilalesi: yes, i want to keep it a bit more and delete it in different PR along with other un-used packages

bilalesi:
> Hi Bilal, while the speed comparison looks good I'm somehow hesitant to make this switch given the maturity of the module.
> 
> What's the necessity?

there is no necessity, this is just an optimization, the outcome might not be obvious but it's really fascinating about the performance table, may be you can check this https://github.com/aidenybai/cnfast/blob/main/docs/architecture.md for another judgement

bilalesi:
btw: it's true it's not fully mature yet  but most of the famous UI libs already migrate to it, so i don't think that's bad

pgetta (approved):

PR #1775, opened 2026-06-24
fix virtual lab creation conflicts at registration

bilalesi:


pgetta (approved):

pgetta:
Ah, we ran out of AWS Amplify quota. I'll check if we can clean up some branches.

bilalesi:
that's okey for this PR, is mostly can not be tested quickly before the merge to prod, I tried already locally

PR #1778, opened 2026-06-25
feat(circuit): derivation type column + filter on circuits overview

pgetta:
## Summary

Adds a **"Derivation type"** column to the Circuits overview showing how each circuit was derived — `Extracted` (`circuit_extraction`), `Rewired` (`circuit_rewiring`), `Customized` (`circuit_customization`), or empty when not derived — plus a **static dropdown filter** by derivation type. The column renders identically in both the **flat list and the hierarchical (tree) view**. Sub-types are out of scope per the issue.

The implementation mirrors the existing **Build category** field (`fields-defs/model.tsx`):
- **Column data:** the circuit list request opts in via `?expand=generated_from_derivations`; each circuit carries its incoming derivations, and the column renders the derivation type(s) — comma-separated in the rare multi-derivation case. The hierarchical view's enrichment fetch (`useFullRawHierarchy`) opts into the same `expand`, so the column populates in the tree too.
- **Filter:** static dropdown using the `generated_derivation__derivation_type__in` query param (a plain enum filter, not a facet).
- **Scope:** only the three circuit derivation types above are labelled (via `CircuitDerivationShortLabel`); other/future types are omitted from the column. The column (`getCircuitDerivationColumnLabels`) and the filter (`CircuitDerivationFilterOptions`) both derive from that one map, so adding a type is a one-line change.

## Files changed

- `api/entitycore/types/entities/derivation.ts` — short labels (`CircuitDerivationShortLabel`), `getCircuitDerivationColumnLabels`, `CircuitDerivationFilterOptions`; **new shared mixins** `EntityDerivations` / `EntityExpandParam` / `EntityDerivationFilter` (+ `INestedEntityRead`, `IGeneratedFromDerivation`, `IUsedByDerivation`)
- `api/entitycore/types/entities/derivation.test.ts` — **new** unit tests for `getCircuitDerivationColumnLabels` (single / multiple / dedup / non-circuit / empty cases)
- `api/entitycore/types/entities/circuit.ts` — `ICircuit` composes `EntityDerivations`, `ICircuitFilter` composes `EntityDerivationFilter` (dropped the Circuit-specific `generated_derivations` field, `ExpandCircuitParam`, `ICircuitGeneratedDerivation`, inline filter keys)
- `api/entitycore/queries/model/circuit.ts` — `expand` param on `getCircuits` typed `EntityExpandParam`
- `entity-configuration/domain/model/circuit.ts` — list query sends `expand: 'generated_from_derivations'`
- `entity-configuration/definitions/fields-defs/{enums.ts,model.tsx}` — new field + definition
- `entity-configuration/definitions/view-defs/model/circuit.ts` — column added to the overview
- `ui/segments/explore/circuit/use-hierarchy.tsx` — hierarchy enrichment fetch (`useFullRawHierarchy`) opts into `expand=generated_from_derivations` (+ matching query key) so the column renders in the tree view

## Related

- Feature request: openbraininstitute/prod-explore-functionality#517
- Backend: openbraininstitute/entitycore#647
- Sibling (shares the backend work): openbraininstitute/prod-explore-functionality#518

### DIFF
--- a/src/ui/segments/explore/circuit/use-hierarchy.tsx
+++ b/src/ui/segments/explore/circuit/use-hierarchy.tsx
@@ -81,12 +81,17 @@ export function useFullRawHierarchy({
               withFacets: false,
               virtualLabId,
               projectId,
+              // Keep the cache key in sync with the expanded fetch below.
+              expand: 'generated_from_derivations',
               ...circuitScaleFilter,
             }),
             queryFn: () =>
               getCircuits({
                 withFacets: false,
                 context: virtualLabId && projectId ? { virtualLabId, projectId } : undefined,
+                // Enrich hierarchy rows with their incoming derivations so the "Derivation type"
+                // column renders in the tree view too, matching the flat list (issue #517).
+                expand: 'generated_from_derivations',
                 filters: {
                   page: 1,
                   page_size: DEFAULT_PAGE_SIZE,
--- a/src/api/entitycore/queries/model/circuit.ts
+++ b/src/api/entitycore/queries/model/circuit.ts
@@ -1,7 +1,11 @@
 import { entityCoreApi, getEntityCoreContext } from '@/api/entitycore/utils';
 import { compactRecord } from '@/utils/dictionary';
 
-import type { ICircuit, ICircuitFilter } from '@/api/entitycore/types/entities/circuit';
+import type {
+  ExpandCircuitParam,
+  ICircuit,
+  ICircuitFilter,
+} from '@/api/entitycore/types/entities/circuit';
 import type { TDerivationType } from '@/api/entitycore/types/entities/derivation';
 import type { HierarchyTreeResponse } from '@/api/entitycore/types/shared/hierarchy';
 import type { EntityCoreResponse } from '@/api/entitycore/types/shared/response';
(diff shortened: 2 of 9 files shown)


src/ui/segments/explore/circuit/use-hierarchy.tsx
```diff
@@ -81,12 +81,17 @@ export function useFullRawHierarchy({
               withFacets: false,
               virtualLabId,
               projectId,
+              // Keep the cache key in sync with the expanded fetch below.
+              expand: 'generated_from_derivations',
```
bilalesi, line 85: may be in another Ticket/PR but before this line we are calling `getCircuitHierarchyByDerivation`, which may be reduced to use the new expand query param

bilalesi (approved):

src/ui/segments/explore/circuit/use-hierarchy.tsx
pgetta: You mean to drop that call to the hierarchy endpoint and reconstruct the tree on the client, right? If so, I've noted that down - will look into the details.

pgetta:
Thanks for the review, @bilalesi

src/ui/segments/explore/circuit/use-hierarchy.tsx
pgetta: Bilal, I'll resolve this comment in order to merge the PR, but will be still waiting for your confirmation on the context.

PR #1785, opened 2026-06-29
fix(e2e): reuse shared virtual lab and provision one project per run

bilalesi:
The shared test user may own only one virtual lab. e2e-provision was deleting all of the user's labs and creating a fresh one each run, so concurrent PRs clobbered each other's workspace and the E2E suite failed.

Treat the virtual lab as a persistent resource (find-or-create) and give each CI run its own isolated project inside it. Cleanup and teardown now delete only the run's project, never the lab. A best-effort prune removes e2e projects abandoned by crashed/cancelled runs so the lab doesn't grow without bound.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,10 +59,10 @@ jobs:
           path: test-artifacts/unit-coverage-summary.json
           if-no-files-found: warn
 
-  # provisions exactly one virtual lab/project for the whole E2E workflow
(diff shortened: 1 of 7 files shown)


tolokoban (dismissed):

pgetta (dismissed):

tolokoban (approved):

bilalesi:
thanks @tolokoban

PR #1789, opened 2026-06-30
fix(atlas-viewer): stop spurious "Unable to load mesh" popups

pgetta:
## Problem

Fixes openbraininstitute/prod-explore-functionality#490.

A red **`Unable to load mesh`** notification popped up in the 3D brain atlas viewer in two benign situations:

1. **Randomly when navigating between the top-level Data / Workflows / Notebooks tabs** (staging/prod).
2. **When selecting a region that has no 3D mesh in the atlas** — e.g. non-volumetric grouping regions like **"grooves"** on the mouse atlas.

In both cases nothing is actually broken, yet the viewer surfaced a global error toast.

## Root cause

The viewer drives a `Painter` WebGL instance; `setRegions` → `addMesh` `await`s a mesh fetch + `TgdDataGlb.parse()`, then builds painters against the captured `context`. A failure anywhere in that path was dispatched via `eventError` → a **global** antd notification.

- **Tab navigation:** unmount/remount calls `painter.start(null)`, which deletes the WebGL context and bumps `contextVersion`. If the user switches tabs mid-await, the resolved load runs against the **deleted** context and throws `"[TgdContext] This context has been deleted: …"`. `setPointCloud()` already guarded against this with a `contextVersion` snapshot; `setRegions`/`addMesh` did not.
- **No-mesh region:** the region hierarchy is a *superset* of the atlas mesh set. Regions with no `brain-atlas-region` mesh entity are greyed out but still selectable, so selecting one fetches a mesh that cannot exist. The lookup threw a generic `Error` (`Unable to find region "…" in current Atlas …`), surfaced as the popup.

## Fix

Distinguish expected/benign conditions from genuine failures, and only pop up for the latter:

- **Tab navigation:** `setRegions` snapshots `contextVersion` at entry and bails out silently after each await (post-fetch, post-`addMesh`, and in the `catch`); `addMesh` re-checks after `parse()` before touching the context. Suppression is keyed on lifecycle, never on error type.
- **No-mesh region:** a typed `RegionMeshNotAvailableError` (kept in its own `services/errors.ts` so the tests' module-mock of `services.ts` leaves the class real and `instanceof` keeps working) is thrown from both no-mesh sites; `setRegions` detects it via `isMeshNotAvailableError` and logs-and-skips that region instead of dispatching.

Genuine failures on the live context (mesh download errors, corrupt GLB) still show the notification exactly as before.

## Tests

`painter.test.ts` (Vitest) covers both paths: a load that resolves after a teardown, and a region with no atlas mesh, each log without a popup; a genuine failure on the live context still pops. Verified end-to-end against Species: mouse → Region: grooves.

### DIFF
--- a/src/features/brain-atlas-viewer/brain-atlas-viewer-gltf/painter.test.ts
+++ b/src/features/brain-atlas-viewer/brain-atlas-viewer-gltf/painter.test.ts
@@ -102,6 +102,10 @@ vi.mock('@/utils/logger', () => ({
 
 // Imported after the mocks are registered (vitest hoists vi.mock above this).
 import { Painter } from './painter';
+// `errors` is intentionally NOT mocked, so the real class is shared with painter.ts
+// and `instanceof` works; `getCachedBrainRegionMeshArrayBuffer` comes from the mock above.
+import { RegionMeshNotAvailableError } from './services/errors';
+import { getCachedBrainRegionMeshArrayBuffer } from './services/services';
 
 import type { VisibleRegion } from './types';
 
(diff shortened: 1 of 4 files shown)


tolokoban (approved):

pgetta:
Thank you Fabien!

PR #1797, opened 2026-07-02
fix: guard against null href in Get Started card links

loris-olivier-obi:
## Summary
Harden link handling in the project **Get Started** card so a `null`/empty value is never assigned to a `next/link` `href`.

- `resolveHref` now accepts `string | null | undefined` links and a nullable `baseRoute`, trims the value, and returns `null` when there is no valid target (instead of building a broken href).
- Added a `toHref` helper to normalize resource link strings.
- `baseRoute` is only built when both `virtualLabId` and `projectId` exist; otherwise it is `null`.
- When a resolved href is missing, the CTA and each resource item render as a non-clickable `<span>` (same styling) instead of a `<Link>` with a null href.
- Extracted shared `ctaLinkClassName` / `resourceLinkClassName` constants.

## Notes
This branch was cleaned to contain **only** this change — Storybook scaffold files that had been committed here by mistake now live on `storybook-initialisation`.

[Warp conversation](https://app.warp.dev/conversation/dcf11d0a-b1c9-4e75-abac-db08a95ef4ac)

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/src/ui/segments/project/get-started/sections/get-started-card.tsx
+++ b/src/ui/segments/project/get-started/sections/get-started-card.tsx
@@ -42,14 +42,30 @@ function CardMedia({ card }: { card: TProjectHomeGetStartedCard }) {
   );
 }
 
-function resolveHref(link: string, baseRoute: string): string {
-  if (link.startsWith('/app/') || link.startsWith('http')) return link;
-  return `${baseRoute}${link}`;
+function resolveHref(link: string | null | undefined, baseRoute: string | null): string | null {
+  const trimmedLink = link?.trim();
+  if (!trimmedLink) return null;
+  if (trimmedLink.startsWith('/app/') || trimmedLink.startsWith('http')) return trimmedLink;
+  if (!baseRoute) return null;
+  return `${baseRoute}${trimmedLink}`;
 }
+
+function toHref(link: string | null | undefined): string | null {
+  const trimmedLink = link?.trim();
+  return trimmedLink || null;
+}
+
+const ctaLinkClassName =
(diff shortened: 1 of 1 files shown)


pgetta:
It seems there are unrelated changes, is that on purpose?

src/ui/segments/project/get-started/sections/get-started-card.tsx
```diff
@@ -42,14 +42,30 @@ function CardMedia({ card }: { card: TProjectHomeGetStartedCard }) {
   );
 }
 
-function resolveHref(link: string, baseRoute: string): string {
-  if (link.startsWith('/app/') || link.startsWith('http')) return link;
-  return `${baseRoute}${link}`;
+function resolveHref(link: string | null | undefined, baseRoute: string | null): string | null {
+  const trimmedLink = link?.trim();
+  if (!trimmedLink) return null;
+  if (trimmedLink.startsWith('/app/') || trimmedLink.startsWith('http')) return trimmedLink;
```
pgetta: I'm curios why do we have such a variety of values: `null`, `undefined`, strings padded with some whitespaces?

pgetta (changes requested):
Overall looks good but we need to revert those unrelated changes.

loris-olivier-obi:
@pgetta I removed the changes not related to our current PR.

BoBer78:
> Overall looks good but we need to revert those unrelated changes.

So there where some things that should not have been touched in the original PR. It broke some of the chat actions. 
Should we do it in another PR ?

src/ui/segments/project/get-started/sections/get-started-card.tsx
loris-olivier-obi: @pgetta I simplified the null or undefined check

pgetta:
> > Overall looks good but we need to revert those unrelated changes.
> 
> So there where some things that should not have been touched in the original PR. It broke some of the chat actions. Should we do it in another PR ?

That's the best practice.

First - PR title and description say nothing about the chat changes so that would be very confusing if one will be looking into the commit history. And the second general reason is that if later on we see that one of the commits had broken something instead of fixing it - we can just revert it and that becomes problematic if we have multiple unrelated changes in that commit.

pgetta (approved):
Thank you Loris!

PR #1799, opened 2026-07-03
fix(grading): rename exercise_id to assignment_id for LTI 1.3 launches

pgetta:
## Context

`grading-service` migrated from **LTI 1.1** to **LTI 1.3 Advantage + AGS** ([grading-service#4](https://github.com/openbraininstitute/grading-service/pull/4)). A Moodle activity now is an **assignment** (one notebook holding one or more independently-graded exercises), so the signed web-launch handed off to the Core Web app now carries **`assignment_id`** instead of `exercise_id`.

The Core Web app never speaks LTI itself — it verifies the HMAC-signed handoff URL and spawns the notebook pod. Grade write-back stays in-pod (`obi-notebook`), so this is an identifier rename, not new grading logic.

## Changes — `exercise_id` → `assignment_id` end to end

**Launch handoff**
- HMAC signing string is now `token|assignment_id|virtual_lab_id|exp` (byte-for-byte matching grading-service), the parsed query param, and the notebook-service `grading` block key.

**Notebook-template lookup + authoring**
- `getNotebooks` filter, `INotebook` / `NotebookFilter` types (`assignment_id` / `assignment_id__in`), the create payload, and the authoring form field (now **"Assignment ID"**).

**Copy**
- User-facing error/copy strings; the project-picker button now reads **"Launch notebook"** (it opens a notebook for the assignment).

**Tests**
- New `src/features/grading/launch.test.ts` pins the HMAC verification contract: a correctly-signed `assignment_id` launch passes; tampered / altered / wrong-secret / expired / non-numeric / missing-field are rejected (the last also covers a legacy `exercise_id`-only URL now being rejected).

### DIFF
--- a/src/api/entitycore/queries/experimental/analysis-notebook-template.ts
+++ b/src/api/entitycore/queries/experimental/analysis-notebook-template.ts
@@ -18,7 +18,7 @@ const AnalysisNotebookTemplateSchema = z.object({
       message: 'Experimental analysis notebook template description is required',
     }),
   scale: z.string().optional(),
-  exercise_id: z.string().optional(),
+  assignment_id: z.string().optional(),
 });
 
 export type TAnalysisNotebookTemplateCreate = z.infer<typeof AnalysisNotebookTemplateSchema>;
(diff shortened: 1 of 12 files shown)


tolokoban (approved):

pgetta:
Thank you Fabien!

pgetta:
Relates to https://github.com/openbraininstitute/PMO/issues/687

PR #1807, opened 2026-07-06
feat(191): color by node property

bilalesi:


package.json
```diff
@@ -44,7 +44,7 @@
     "@lezer/javascript": "^1.5.4",
     "@monaco-editor/react": "^4.7.0",
     "@octokit/core": "^7.0.6",
-    "@openbraininstitute/morphoviewer": "file:vendor/openbraininstitute-morphoviewer-0.30.0.tgz",
+    "@openbraininstitute/morphoviewer": "0.30.0",
```
tolokoban, line 47: Just a tip to avoid generating a `*.tgz` every time you want to test the library.

Write this in the `package.json` file:

```json
"@openbraininstitute/morphoviewer": "file:../morphoviewer",
```

And git clone the library in the same folder where core-web-app is.

With `pnpm` you will still have to do `pnpm install` everytime the code of the library changes. Strtangely, this is not the case with `npm` because it deals differently with the links.

But at least, you can spare the time of making a package for the library.

scripts/sync-morphoviewer.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
```
tolokoban: See my comment in `package.json`.
This script could have been avoided.

package.json
bilalesi: i tried that, and i got some errors due turbopack can not use file paths, may be i was missing something, but next time definitely i will try it 

thanks for the tip

bilalesi: btw: i have in `scripts/sync-morphoviewer.sh` to that automatically

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
```diff
@@ -1,8 +1,10 @@
 /** biome-ignore-all lint/suspicious/noArrayIndexKey: The list is not suppose to change */
+
 import { saveAs } from 'file-saver';
 import React from 'react';
 
-import { LoadingNeuronSpinner } from '@/components/neuron-viewer';
+import { VERTICAL_SCALEBAR } from '@/features/scan-config/components/shared/3d-viewer';
+import { VisualizationLoadingIndicator } from '@/features/scan-config/components/shared/visualization-loading-indicator';
```
tolokoban, line 7: Should we also change the spinner for the other 3D viewers to be consistent?

tolokoban (commented):
Thats' great!
Just a little question about the new spinner.

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
bilalesi: yes i think so, but it will be in a second step, as not only the spinner should change but also settings menu that contains fullscreen, background color, screenshot, which may cause the PR to be too big

scripts/sync-morphoviewer.sh
bilalesi: done

tolokoban (approved):
Very good!

PR #1813, opened 2026-07-09
feat(scan-config): refresh project balance on task launch and terminal status

pgetta:
## Problem

When a scan-config task is launched, the accounting service reserves credits; when the task ends in `error`/`cancelled`, the reservation returns to the project balance. The UI mostly didn't reflect this:

- The generic task-runner path (build / extraction / skeletonization tabs and task-system simulations) never refreshed the balance — not on launch, not on completion.
- Only the legacy simulations adapter refreshed it, on `DONE`/`ERROR` only, and with no transition detection — it re-fired the refetch for already-finished simulations on every page visit.

## Changes

- New `src/features/task-runner/hooks/use-balance-refresh.ts`:
  - `invalidateProjectBalance` — invalidates both balance keys (`keyBuilder.accounting` for the balance card, `keyBuilder.wallet` for the top-nav wallet). Uses `cancelRefetch: false` so several task rows finishing on the same 10s poll tick share one in-flight refetch instead of cancel-restarting it.
  - `useBalanceRefreshOnTaskCompletion` — fires exactly once when a task transitions in-session from an active status to a terminal one (`done`/`error`/`cancelled`); deliberately silent when a status first loads already terminal, so revisiting finished campaigns no longer triggers refetches.
- Launch-time refresh (reservation leaves the balance): once per batch in `useTaskLaunchMutation` (covers build/extraction/skeletonization) and in both simulation launch paths (`runViaLaunchSystem` and the legacy `runSimulationBatch` `onInit`).
- Completion-time refresh (reservation returns on error/cancel): the hook is mounted in the two per-row status observers — `TaskConfigSelectionCardWithStatus` (all three generic tabs) and `SimulationListItem` (both simulation flows, replacing the old inline block).
- Hoisted the terminal-status set into `task-runner/status.ts` as `isTerminalActivityStatus` and deduped the inline `[CANCELLED, DONE, ERROR].includes(...)` copies in the build/extraction/skeletonization tabs.

## Testing

- New Vitest suite for the helper and every transition case (fires on active→terminal, silent on already-terminal mount, active→active, terminal→terminal, unchanged re-renders).
- Full unit suite green; Biome clean; no new `tsc` errors in changed files.
- Manual check worth doing on a real backend: launch a task and watch the wallet refetch — the API's `ProjectBalance.reservation` field is fetched but not rendered, so the visible number only moves if the backend deducts reservations from `balance`.

### DIFF
--- a/src/features/scan-config/components/shared/task-config-selection-list.tsx
+++ b/src/features/scan-config/components/shared/task-config-selection-list.tsx
@@ -5,6 +5,7 @@ import { useEffect } from 'react';
 import { TaskConfigSelectionCard } from '@/features/scan-config/components/shared/task-config-selection-card';
 import { TASK_STATUS_QUERY_KEY_HEAD } from '@/features/task-runner';
 import { useTaskConfigExecution } from '@/features/task-runner/hooks/queries';
+import { useBalanceRefreshOnTaskCompletion } from '@/features/task-runner/hooks/use-balance-refresh';
 import { cn } from '@/utils/css-class';
 
 import type {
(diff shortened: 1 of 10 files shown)


pgetta:
Relates to [Credits are not returned to user when the job fails](https://github.com/openbraininstitute/prod-virtual-lab/issues/361).

tolokoban (approved):

pgetta:
Thanks Fabien!

PR #1814, opened 2026-07-10
glb mesh upload support

dkeller9:
## Summary

Allow the EM cell mesh upload step to accept a single `.obj` **or** `.glb` file, instead of only `.obj`.

## Changes

- **`schema.ts`**: Replaced the per-extension `assets.obj` field with a single `assets.mesh` field, validated against a shared `EM_CELL_MESH_FILE_TYPES` list (`obj`, `glb`). `getEMCellMeshMimeType` now resolves the mime type for whichever extension was uploaded.
- **`steps/asset-upload.tsx`**: Updated the dropzone `accept` list to include `model/gltf-binary` / `.glb`, bound the form to `assets.mesh`, and renamed the validation handler to `validateMeshFile`.
- **`pipeline.ts`**: Asset creation now reads `values.assets.mesh` and derives `mimeType` via `getEMCellMeshMimeType(file)` instead of hardcoding `application/obj`.
- **`api/one/em-cell-mesh.ts`**: Renamed `resolveOBJFile` → `resolveMeshFile` and `OBJResolution` → `MeshResolution` to reflect that it validates either mesh format. Still hits the same `/declared/test-mesh-file` obi-one endpoint.
- **`i18n/en/upload.ts`**: Added `ResolveMeshFileFailed` (replacing a reference to `ResolveOBJFileFailed`, which was never actually defined — the old error path would have thrown on `.replace()` of `undefined`).

Only one file can be selected at a time (`maxFiles: 1`, unchanged).

## Dependencies

Requires the `/declared/test-mesh-file` endpoint on obi-one to accept `.glb` uploads — tracked separately in the backend mesh validation/registration work: https://github.com/openbraininstitute/obi-one/pull/894

### DIFF
--- a/src/i18n/en/upload.ts
+++ b/src/i18n/en/upload.ts
@@ -3,6 +3,8 @@ export const messages = {
     'The nwb file $$ could not be validated by our tool. Please ensure it is a valid .nwb file.',
   ResolveNeuronFileFailed:
     'The neuron file $$ could not be validated by our tool. Please ensure it is a valid .swc, .asc, or .h5 file loadable by neurom.',
+  ResolveMeshFileFailed:
+    'The mesh file $$ could not be validated by our tool. Please ensure it is a valid .obj or .glb file.',
   FileSizeExceeded: 'File "$$" exceeds the maximum size of $$$.',
   FileNotAccepted: 'File "$$" is not an accepted file type.',
   MaxFilesExceeded: 'You can only upload a maximum of $$ files.',
--- a/src/i18n/en/upload.ts
+++ b/src/i18n/en/upload.ts
@@ -1,10 +1,10 @@
 export const messages = {
   ResolveNWBFileFailed:
-    'The nwb file $$ could not be validated by our tool. Please ensure it is a valid .nwb file.',
+    'The nwb file $$ could not be validated. Please ensure it is a valid .nwb file.',
   ResolveNeuronFileFailed:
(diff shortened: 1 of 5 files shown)


src/i18n/en/upload.ts
```diff
@@ -3,8 +3,10 @@ export const messages = {
     'The nwb file $$ could not be validated by our tool. Please ensure it is a valid .nwb file.',
   ResolveNeuronFileFailed:
     'The neuron file $$ could not be validated by our tool. Please ensure it is a valid .swc, .asc, or .h5 file loadable by neurom.',
+  ResolveMeshFileFailed:
+    'The mesh file $$ could not be validated by our tool. Please ensure it is a valid .obj or .glb file.',
```
pgetta: Just a nitpick: "_by our tool_" doesn't carry any useful information for the user and probably it's best to just drop it.

pgetta (dismissed):
Looks good, I've left a small comment about the wording: please see if it makes sense to address.

src/i18n/en/upload.ts
dkeller9: Thanks, the wording has now been changed.

dkeller9:
> Looks good, I've left a small comment about the wording: please see if it makes sense to address.

@pgetta your suggestion is now incorporated, please re-approve if okay.

pgetta (approved):

PR #1819, opened 2026-07-14
Course vlabs feeback

g-bar:


g-bar:
I added the query invalidation post course activation to this PR.

src/components/activate-enrolments.tsx
```diff
@@ -1,29 +1,62 @@
...
+            headers: { 'content-type': 'application/json' },
+          }
+        );
+
+        const data: ActivateEnrolmentsResponse = await response.json();
+
+        if (data.results?.some((r) => r.activated)) {
+          queryClient.invalidateQueries({ queryKey: keyBuilder.membership() });
+          queryClient.invalidateQueries({
+            queryKey: [`${prefix}/list-filtered-virtual-labs`],
```
pgetta, line 46: Probably this should not be hardcoded (as well as the next key)

Similar key is defined in `src/ui/use-query-keys/workspace.tsx`, but I see there are other places where this value has been hardcoded, but we can address that later.

pgetta (approved):

PR #1820, opened 2026-07-14
feat(189): extracellular preview and download

bilalesi:


bilalesi:
approved by the PO

src/ui/segments/mini-detail-view/previews/extracellular-recording-array-preview.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const imageAsset = getAsset({
+    assets: record.assets,
+    label: AssetLabel.electrode_array_image,
+  }).getOneOrNull();
+
+  if (!imageAsset) return null;
+
+  return (
+    <div className="flex w-full items-center justify-center rounded-md bg-white">
+      {renderPreview(
```
pgetta, line 27: Nice - reusing `renderPreview` with a `customRender` to piggyback on the thumbnail machinery avoids a whole second fetch/loading flow for the electrode image. 👍

src/features/scan-config/use-cases/build/in-out-files.tsx
```diff
@@ -48,97 +145,95 @@ export function InOutFiles({
...
+  const shouldPollVisualization =
+    isCircuitBuild && campaignOrigin !== ScanConfigCampaignOriginActionDict.View;
 
-  const builtCircuitId = execution?.generated?.[0]?.id;
-  const { data: builtCircuit, isLoading } = useQuery({
-    queryKey: keyBuilder.oneCircuit({
-      virtualLabId: context.virtualLabId,
-      projectId: context.projectId,
-      entityId: builtCircuitId ?? '',
+  const { data: builtEntity, isLoading } = useQuery({
```
pgetta, line 165: This two-step resolution (fetch the type via `getEntity`, then the concrete entity) is exactly what `useModelQuery` already does, and it even reuses the same `keyBuilder.entity` keys so the caches line up. The only thing stopping direct reuse is its `.otherwise` throws for the recording array.

I wonder, would it be worth teaching `useModelQuery` to fall back to `retrieveEntity` for unsupported types, so we keep a single resolution path instead of two that have to stay in sync? Not a call for action for this PR - more a note for a follow-up.

src/features/scan-config/use-cases/build/in-out-files.tsx
```diff
@@ -38,6 +55,86 @@ type Props = {
...
+
+function isCircuitType(type: string | null | undefined): boolean {
+  return type === EntityTypeDict.Circuit;
+}
+
+function shouldListAssetsAsOutputs(type: string | null | undefined): boolean {
+  return !!type && ENTITY_TYPES_WITH_ASSET_OUTPUTS.has(type);
+}
+
+function findAssetByLabel(assets: readonly IAsset[], label: AssetLabel): IAsset | null {
```
pgetta, line 71: `getOneOrNull()` throws `AssetMatchAmbiguousError` when more than one asset matches the label - it doesn't return null. The previous code used `.some()` / `.find()`, which tolerated duplicates.

My concern is mostly `shouldPollForCircuitVisualization`, since it runs inside `refetchInterval`: if a circuit ever exposes two `circuit_visualization` assets, the callback would throw on every tick instead of just reporting "has vis asset". Same reasoning applies to the `sonata_circuit` lookup and to `electrode_array_image` in the preview.

Is a single match guaranteed for these labels? If so, all good - maybe worth a comment saying so. If not, `getAllOrNull()?.[0]` would keep the old forgiving behavior. Please correct me if the selector can't throw here and I misread it.

src/features/scan-config/use-cases/build/in-out-files.tsx
```diff
@@ -38,6 +55,86 @@ type Props = {
...
+
+function isSingleNeuronCircuit(entity: TActivityCustomFile['entity']): boolean {
+  if (!isCircuitType(entity.type) || !('scale' in entity)) return false;
+  return (entity as ICircuit).scale === CircuitScaleDictionary.Single;
+}
+
+function getOutputEntityLabel(file: TActivityCustomFile): string | null {
+  if (file.renderer !== ActivityCustomFileRenderer.MiniDetailView) return null;
+
+  // EM synapse-mapping outputs a single-scale circuit, labeled Synaptome (beta)
```
pgetta, line 113: Just want to note a small visual change: the badge used to render `Synaptome <span class="lowercase">(beta)</span>` ("SYNAPTOME (beta)"), and now it comes from `getEntityTypeTagLabel(SingleNeuronCircuit)`, which is already uppercased, so it reads "SYNAPTOME (BETA)". Probably fine, just flagging in case the lowercase "(beta)" was intentional.

src/features/scan-config/components/file-viewer/image-viewer.tsx
```diff
@@ -0,0 +1,75 @@
...
+        entityId: entity.id,
+        virtualLabId: context.virtualLabId,
+        projectId: context.projectId,
+        configAssetId: asset.id,
+        assetPath,
+      });
+      return url;
+    },
+    enabled: !!entity.id && !!asset.id,
+    staleTime: Infinity,
```
pgetta, line 39: Two small things around the presigned URL. `staleTime: Infinity` caches it for the lifetime of the query - these usually expire server-side after a while, so a long-lived viewer could end up holding an expired URL. And the `<Image>` below has no `onError`, while `showLoading` stays true until `onLoad` fires, so a failed load (expired/broken URL) leaves the skeleton spinning forever.

Would be nice to route `onError` into the same `Empty` "Failed to load image" state you already have for the fetch failure. Is the expiry long enough that this never bites in practice?

src/features/entity-download/file-handlers/simulatable-extracellular-recording-array.ts
```diff
@@ -0,0 +1,57 @@
...
+import { Metadata } from '@/features/entity-download/metadata';
+import {
+  createAssetFileEntry,
+  createTemplateFileEntry,
+  getMetadataCsvEntryBase,
+} from '@/features/entity-download/utils';
+
+import type { WorkspaceContext } from '@/types/common';
+
+export async function* getSimulatableExtracellularRecordingArrayFiles(
```
pgetta, line 13: Btw, this generator is pure enough to be pleasant to cover with a small Vitest test (assets yielded + metadata entries), and it's the kind of thing that regresses silently. Fine to leave for later given the scope - just a note for the future.

pgetta (commented):
Nice work, Bilal, the generalization from circuit-only to any built entity reads well. I've left a few small comments and questions - nothing blocking.

pgetta (approved):

pgetta:
Hey Bilal, that's me testing LLM code rieviews, it's impressive: getting better and better.

I still need to adjust the sensitivity.  Most of these comments are nitpicks to me. Feel free to merge them when you’re happy.

src/features/scan-config/use-cases/build/in-out-files.tsx
bilalesi: it's different because there specific handling of the refetching

src/features/scan-config/use-cases/build/in-out-files.tsx
bilalesi: yes it's guaranteed

src/features/entity-download/file-handlers/simulatable-extracellular-recording-array.ts
bilalesi: will do in another PR, for the whole download feature

src/features/scan-config/components/file-viewer/image-viewer.tsx
pgetta: Actually, the `staleTime: Infinity` seems to be not an issue at all, so please disregard.

bilalesi:
thanks @pgetta I will add tests for the download feature for all entities in next PR

PR #1832, opened 2026-07-24
feat(1854): migration `antd` data tables to `aggrid` 

bilalesi:
related #1854 

it's big PR, as this defined the full api to manage grids, and filters
the legacy will be removed in a another PR to have "less" changes 
the scientific team already tested and now the new grids are similar to the legacy ones, in terms of columns and capabilities


pgetta (commented):
Nice work, Bilal, that's a huge migration and the ring layering reads really well - the core ring knowing nothing about entitycore or AG Grid makes this much easier to follow than the old tables.

I've left a number of comments, mostly questions about behaviour from the antd tables that I couldn't find again in the new grid.

src/features/data-grid/host/browse-entity-grid.tsx
```diff
@@ -0,0 +1,501 @@
...
+        <div
+          id="mini-detail-view-container"
+          className={cn(
+            'h-full max-h-[calc(100vh-10.8rem)] w-full min-w-0',
+            '[grid-area:mini-view]',
+            { hidden: !mdv },
+            classNames?.miniView
+          )}
+        >
+          <MiniDetailView section={section} {...miniViewProps} dataType={dataType} />
```
pgetta, line 446: It seems `isPrivate` and `virtualLabData` are not passed to `MiniDetailView` anymore, while `BrowseEntityScope` used to fetch and forward them. In `NotebookActions` `canDelete` is `isPrivate && !!virtualLabData && ...` (mini-detail-actions.tsx:134-137) and the "Sync notebook to students" button checks `virtualLabData?.course`, so with `virtualLabData` always undefined both are permanently false and every notebook's mini-detail loses its delete and sync actions. `isPrivate` also silently defaults to `true` instead of `scope === WorkspaceScope.Project`.

Or maybe I'm missing something and these get wired somewhere else?

src/features/data-grid/host/browse-entity-grid.tsx
```diff
@@ -0,0 +1,501 @@
...
+      if (!loading) setGridTotal(total);
+    },
+    [setGridTotal]
+  );
+  useEffect(() => () => setGridTotal(undefined), [setGridTotal]);
+
+  const operators = useMemo(() => createDefaultOperatorRegistry(), []);
+  const cellRenderers = useMemo(() => buildCellRenderers(definition), [definition]);
+
+  const handleRowClick = useCallback(
```
pgetta, line 185: From what I can see `mainTableProps` is only read for `selectionType`/`onRowsSelected`/`selectedRows`, and the row click is hardcoded to the mini-detail event. A few build pickers pass `mainTableProps.onCellClick` to navigate to the overview page (m-model.tsx:56, e-model.tsx:86, me-model.tsx:72) - there a row click now opens the mini-detail panel instead, and the callback is never invoked.

Was that an intended behaviour change for the pickers?

src/features/data-grid/host/browse-entity-grid.tsx
```diff
@@ -0,0 +1,501 @@
...
+
+  // Picker selection is opt-in: only when `mainTableProps` supplies both a type and
+  // an `onRowsSelected` callback. Otherwise selection stays schema-driven (bulk actions).
+  const selectionType = mainTableProps?.selectionType;
+  const onRowsSelected = mainTableProps?.onRowsSelected;
+  const controlledSelectedRows = mainTableProps?.selectedRows;
+  const pickerSelection = useMemo<
+    IDataGridSelection<EntityCoreIdentifiableNamed> | undefined
+  >(() => {
+    if (!selectionType || !onRowsSelected) return undefined;
```
pgetta, line 198: The main data browse page passes `selectionType: 'checkbox'` together with `allowDownload allowDelete` but no `onRowsSelected`, so it falls back to `isSelectionEnabled(schema, ctx)`. For dataTypes whose schema declares no `selection` (simulation-campaign and circuit-simulations have none) the checkbox column, the "N selected" footer and both bulk actions disappear, while the antd table rendered them from `selectionType` alone.

Probably it's best to treat `selectionType` on its own as enough to enable selection.

src/features/data-grid/host/grid-search.tsx
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/**
+ * The listing quick-search: a round button that expands into a pill input with a
+ * clear affordance. Debounced text is pushed to the grid store's quick filter via
+ * {@link onSearch}.
+ */
+export function GridSearch({ onSearch, openOnMount = false, className }: IGridSearchProps) {
+  const [open, setOpen] = useState(openOnMount);
+  const [text, setText] = useState('');
```
pgetta, line 23: `quickFilter` is in `SESSION_KEYS` and the controller hydrates it from session storage, but the input starts from `useState('')` and there's no `value`/`defaultValue` prop to reflect it. Coming back to a listing in the same tab, the request still carries `search=pyramidal` while the pill renders open and empty with no clear button, so the user sees a filtered result set with no indication why.

src/ui/segments/workflows/elements/workflow-activity.tsx
```diff
@@ -538,160 +372,99 @@ export function WorkflowActivity() {
...
-                      className="disabled:bg-background disabled:text-label select-none disabled:cursor-not-allowed text-primary-9!"
-                      disabled={!canDuplicate}
-                    >
-                      Duplicate
-                    </Button>
-                  </div>
-                )}
+            <div className="relative min-h-0 flex-1">
+              <div className="h-full overflow-hidden">
+                <SimpleGrid<EntityCoreObjectTypes>
```
pgetta, line 381: I wonder, what's the plan for the expanded campaign rows? The removed `expandableOptions`/`useExpandableTable` rendered the per-execution table (scan parameters, names, statuses) for the campaign types, and `SimpleGrid` here gets no `expansion` config, so from a simulation campaign row there's no way to reach the individual simulations anymore. The status badge popover shows the scan parameter cards but not the per-execution rows, and only on hover.

If it's coming in a follow-up that's fine, I just want to make sure it wasn't dropped accidentally.

src/ui/segments/workflows/elements/workflow-activity.tsx
```diff
@@ -135,187 +99,64 @@ export function WorkflowActivity() {
...
-            );
-          });
+  // server data source: the grid owns paging; each fetch also surfaces total/loading
+  // into `gridMeta` for the empty and loading states
+  const entity = getEntityByExtendedType({ type: resolvedEntityType });
+  const listQuery = entity?.api.query?.list;
+  const dataSource = useMemo<IGridDataSource<EntityCoreObjectTypes>>(
+    () => ({
+      fetch: async (q): Promise<IGridPage<EntityCoreObjectTypes>> => {
+        setGridMeta((m) => ({ ...m, status: 'loading' }));
```
pgetta, line 137: `setGridMeta` is called as a side effect inside `dataSource.fetch`, and the empty card is rendered from `gridMeta.status === 'loaded' && gridMeta.total === 0`. Let's imagine the following: the user selects a type with 0 activities and sees the card, switches to another type and comes back within the 60s staleTime. `updateEntityType` resets `gridMeta` to `idle`, React Query serves the cached empty page without calling `queryFn`, so `setGridMeta` never runs and instead of "You don't have any activities yet" the user gets an empty grid with the generic no-rows overlay.

Would be more robust to derive this from `useQuery`'s own status/data rather than from a setter inside the fetcher.

src/features/entities/e-model/detail-view/exemplar-traces.tsx
```diff
@@ -57,144 +58,53 @@ type Props = {
...
+          filters: {
+            derivation_type: DerivationTypeDictionary.Unspecified,
+            page: query.page,
+            page_size: query.pageSize,
+          },
+        });
+        const total = derivations.pagination.total_items;
+        const ids = derivations.data.map((p) => p.id);
+        if (total <= 0 || ids.length === 0) return { rows: [], total };
+        const recordings = await getElectricalCellRecordings({
```
pgetta, line 83: `getElectricalCellRecordings` is called without `page_size`, so entitycore's default of 10 applies. The new grid has a page size selector (10/20/30) which the old `Pagination simple` didn't, so picking "30 / page" on an e-model with 30 traces gives 30 ids and a total of 30, but only the first 10 recordings come back - the grid shows 10 rows and claims 30 results.

`page_size: ids.length` should cover it.

src/features/entities/e-model/detail-view/exemplar-traces.tsx
```diff
@@ -57,144 +58,53 @@ type Props = {
...
-              pageNumber: _page,
-              pageSize: _pageSize,
-            });
+      <Header variant={variant}>Exemplar Traces</Header>
+      <div className={cn(detailViewInsetPanelClass(variant))}>
+        <SimpleGrid<IElectricalCellRecording>
+          columns={columns}
+          getRowId={(row) => row.id}
+          pageSize={DEFAULT_PAGE_XSMALL_SIZE}
+          serverSide={{
```
pgetta, line 102: The `ts-pattern` match with the `ErrorData` banner and the loading skeleton is gone, and as far as I can see `ISimpleServerSide` has no error hook - `InMemoryGrid` forwards `serverResult.isLoading` to the AG overlay but never reads `serverResult.error`. So if `getEntityDerivations` fails, the user gets an empty grid, indistinguishable from an e-model that genuinely has no traces.

Maybe worth exposing a `renderError` on the server side config, since other migrated views will run into the same thing.

src/features/data-grid/react/filters/filter-editor.tsx
```diff
@@ -0,0 +1,717 @@
...
+        className={INPUT_CLASS}
+        placeholder="Search…"
+        value={query}
+        onChange={(e) => setQuery(e.target.value)}
+      />
+      <div className="flex justify-between text-xs font-medium">
+        <button
+          type="button"
+          className="text-primary-6 hover:text-primary-7"
+          onClick={() => onChange(visible.map((o) => o.value))}
```
pgetta, line 603: "Select all" replaces the whole selection with whatever is currently visible. So ticking "L5_TPC:A", then typing "L23" in the option search and hitting "Select all" silently drops the L5 pick, with nothing on screen showing it went away. `toggle()` right below accumulates into the existing set, so the two controls disagree - probably "Select all" should union `visible` into `selectedSet`.

src/features/data-grid/bindings/entitycore/schemas/circuit.tsx
```diff
@@ -0,0 +1,368 @@
...
+/** ISO → localized date, matching the legacy `renderDate` (empty when absent). */
+function formatDate(iso?: string | null): string {
+  if (!iso) return '';
+  const d = new Date(iso);
+  return Number.isNaN(d.getTime())
+    ? ''
+    : d.toLocaleDateString(undefined, { year: 'numeric', month: 'short', day: 'numeric' });
+}
+
+export const circuitGridDefinition: IEntityGridDefinition<Row> = {
```
pgetta, line 364: `contributionsColumn` sets `cellRenderer: CONTRIBUTORS_RENDERER`, but this definition doesn't implement `registerCellRenderers`, and `buildCellRenderers` only registers the lifecycle status and description ones by default. So `ctx.cellRenderers.get('contributors')` is undefined and `AgCellHost` falls through to plain text - enabling the Contributors column here shows a comma joined string instead of the "first name +N" pill with its tooltip, unlike every other listing.

Same applies for `schemas/extracellular-recording-array.ts`.

src/features/data-grid/bindings/entitycore/columns/catalog.ts
```diff
@@ -0,0 +1,819 @@
...
+export function ionChannelColumn<Row extends IHasIonChannel>(
+  o?: TColumnOverride<Row>
+): IColumnModel<Row> {
+  return mergeColumnDef<Row>(
+    {
+      id: 'ionChannel',
+      header: 'Ion channel',
+      sortable: true,
+      sortField: 'ion_channel__name',
+      getValue: (r) => r.ion_channel?.name ?? EMPTY_PLACEHOLDER,
```
pgetta, line 442: A small thing, but `ionChannelColumn` returns `EMPTY_PLACEHOLDER` from `getValue` while every other factory returns `''`. Since the value is non-empty, `withEmptyPlaceholder`'s `cellClassRules` never applies `text-gray-300` so the dash renders at full strength, the quick filter matches rows on the dash itself, and sorting treats it as a value rather than a blank.

Probably best to return `''` and let `withEmptyPlaceholder` do the display part. Same for `cellLineColumn` at line 483.

src/features/data-grid/config.ts
```diff
@@ -0,0 +1,18 @@
...
+ *           from the schema's defaults. Previously-saved layouts are then ignored,
+ *           NOT deleted, so flipping this back on restores what users already had.
+ *
+ * This is deliberately separate from the SESSION slice (filters, sort, page, search),
+ * which always persists for the tab's lifetime and is not affected by this flag —
+ * a saved layout is a lasting preference, a filter is a transient act of browsing.
+ */
+export const PERSIST_COLUMN_LAYOUT = true;
+
+export const DEFAULT_PAGE_SIZE_OPTIONS = [10, 20, 30] as const;
```
pgetta, line 18: ProjectActivities and WorkflowActivity pass `DEFAULT_PAGE_MEDIUM_SIZE` (15) and ExemplarTraces passes `DEFAULT_PAGE_XSMALL_SIZE` (5), and none of them sets `pageSizeOptions`, so `GridPagination` falls back to this list. The trigger reads "15 / page" with no item selected, and once you pick anything 15 is unreachable afterwards.

Either pass matching `pageSizeOptions` per preset, or fold these sizes into the shared list.

src/features/data-grid/presets/in-memory-grid.tsx
```diff
@@ -0,0 +1,752 @@
...
+    const byId = new Map(columns.map((c) => [c.id, c] as const));
+    return reconcileColumnOrder(
+      columns.map((c) => c.id),
+      dropPinnedColumns(columns, state.columnOrder)
+    )
+      .map((id) => byId.get(id))
+      .filter((c): c is ISimpleColumn<Row> => c !== undefined);
+  }, [columns, state.columnOrder]);
+
+  const isExpandable = expansion?.isExpandable ?? (() => true);
```
pgetta, line 412: `isExpandable` gets a new identity on every render, and it feeds `renderExpander` which feeds `colDefs`, so for any caller that doesn't pass `expansion.isExpandable` AG Grid receives a brand new `columnDefs` array each render and re-processes all columns. `colDefs` also depends on `expandedSet`, so every expand/collapse rebuilds all of them, which is most noticeable on the nested `CircuitRecursiveGrid`.

Hoisting the default to a module constant and reading the expanded set through a ref/context in the expander cell would avoid both.

src/features/data-grid/presets/in-memory-grid.tsx
```diff
@@ -0,0 +1,752 @@
...
+  // Client mode computes facets from the full row set so set-filter options stay
+  // stable as the grid narrows; server mode reads them off the fetched page.
+  const clientFacets = useMemo(
+    () => (filterable && !isServerMode ? computeInMemoryFacets(rows, columns) : undefined),
+    [filterable, isServerMode, rows, columns]
+  );
+  const facets: TFacets | undefined = isServerMode ? serverResult.data?.facets : clientFacets;
+
+  // Client mode filters/sorts/paginates in memory; server mode uses the fetched page.
+  const clientPage = useMemo(
```
pgetta, line 360: `buildGridQuery` returns a fresh object each render and `query` is in the deps, so the in-memory filter/sort/slice never hits its cache and re-runs on every unrelated re-render. Depending on the state slices (`state.filters`, `state.sort`, `state.page`, `state.pageSize`, `state.quickFilter`) instead would keep it memoised.

src/features/data-grid/renderers/aggrid/empty-cell.tsx
```diff
@@ -0,0 +1,75 @@
...
+ */
+export const EMPTY_PLACEHOLDER = '—';
+
+export const EMPTY_PLACEHOLDER_CLASS = 'text-gray-300';
+
+/**
+ * Columns left blank rather than given the placeholder. Ideally a per-column
+ * `blankWhenEmpty` opt-out on `IColumnModel` rather than an id list here.
+ */
+export const DELIBERATELY_BLANK_COLUMN_IDS: ReadonlySet<string> = new Set<string>([
```
pgetta, line 20: The JSDoc above already says a per-column `blankWhenEmpty` on `IColumnModel` would be preferable, and I'd agree. The file duplicates `EMPTY_PLACEHOLDER` precisely to stay independent from entitycore, and this list brings that dependency back for one circuit column. As it stands, every future column that wants a blank cell has to be added here, and renaming a column id in a schema silently loses the behaviour.

Not a call for action for this PR, but would be nice to flip it around at some point.

pgetta (approved):
Awesome work!

As agreed, I'm approving this change. Possible fixes to be done in a separate PR.

PR #1836, opened 2026-07-27
Remove title prop from CodeBlock in file viewers

danifr:
Minor thing, but the other day while hovering over the file viewer I realized it was showing a tooltip with the file extension (e.g. "json"). I don't think this is the right behavior since it adds clutter and makes the file content less visible. 

This PR tries to fix that by removing the tooltip.

tooltip_bug

### DIFF
--- a/src/features/scan-config/components/file-viewer/code-viewer.tsx
+++ b/src/features/scan-config/components/file-viewer/code-viewer.tsx
@@ -81,7 +81,6 @@ export function CodeFileViewer({
 
   const fileName = assetPath?.split('/').at(-1) ?? asset.path.split('/').at(-1);
   const language = fileName?.split('.').at(-1) as BundledLanguage;
-  const filename = assetPath?.split('.').at(-1) ?? asset.path.split('/').pop() ?? 'file';
 
   const LARGE_FILE_CHAR_LIMIT = 10000;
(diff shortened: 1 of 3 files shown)


src/features/scan-config/components/file-viewer/code-viewer.tsx
```diff
@@ -81,7 +81,6 @@ export function CodeFileViewer({
 
   const fileName = assetPath?.split('/').at(-1) ?? asset.path.split('/').at(-1);
   const language = fileName?.split('.').at(-1) as BundledLanguage;
-  const filename = assetPath?.split('.').at(-1) ?? asset.path.split('/').pop() ?? 'file';
```
bilalesi, line 84: probably keep the title and pass it to `CodeBlockLanguageLabel` component, so it only appear when hover ontop of the file extension pill

danifr: alright, thanks, added in this commit https://github.com/openbraininstitute/core-web-app/pull/1836/changes/4692880b03051958386e393e3ff717a39547fce2

bilalesi (approved):

src/features/scan-config/components/file-viewer/code-viewer.tsx
danifr: Seems to be working:

image

PR #1838, opened 2026-07-27
feat(231): route remaining small NEURON simulation scales through the launch system

pgetta:
## Summary

Relates to openbraininstitute/prod-circuit-simulation#231.

The four small non-legacy NEURON simulate workflows — **Single neuron (beta)**, **Synaptome (beta)**, **Paired neurons (beta)**, **Small microcircuit (beta)** — now launch through obi-one's `POST /declared/task/launch` instead of the small-scale simulator's `/circuit/simulation/run-batch`. Larger scales already did.

Enabled by obi-one #785 ("Neurodamus-fueled beta simulations"), which adds the two dedicated single-neuron task types and repoints the `single|pair|small` branch of `select_simulation_task` at `circuit_simulation_neurodamus_machine`. Frontend-only change.

## Task type per workflow

| Workflow | Model entity | `task_type` sent | obi-one resolves to |
|---|---|---|---|
| Single neuron (beta) | `Memodel` | `single_neuron_simulation_execution` | — |
| Synaptome (beta) | `Circuit`, scale `single` | `single_neuron_synaptome_simulation_execution` | — |
| Paired neurons (beta) | `Circuit`, scale `pair` | `circuit_simulation` | `…_neurodamus_machine` |
| Small microcircuit (beta) | `Circuit`, scale `small` | `circuit_simulation` | `…_neurodamus_machine` |

The two beta workflows can't use the `circuit_simulation` group: its resolver fetches `simulation.entity_id` **as a Circuit**, so it can't serve a me-model campaign. The dedicated types also run on 1 core / 2 GB with their own billing subtypes (`SINGLE_CELL_SIM` / `SYNAPTOME_SIM`) instead of the group's 4 cores / 8 GB.

## Offline-token consent

The launch system requests an offline token from auth-manager only for **cluster** jobs; machine jobs get a session-bound refresh token (`launch-system app/routers/job.py::create_one`). Of the simulation tasks, only `circuit_simulation_neurodamus_cluster` declares cluster resources — reached by NEURON circuits above scale `small`.

The UI asked for consent on *every* launch-system launch, so without this the four workflows above would prompt for a 10-day `offline_access` grant their jobs never use. Consent is now gated on `requiresOfflineTokenConsent`; Brian2 and LearningEngine sims stop over-prompting as a side effect.

⚠️ This mirrors backend policy in the frontend. If obi-one moves a scale onto the cluster executor, launches here start failing at job creation until the table is updated. The durable fix is obi-one returning the resolved executor from `/declared/task/estimate`, which already resolves the task group server-side but returns only cost — worth a follow-up ticket either way.

## Implementation

The launch decision and the task-type choice were previously split across a scale allow-list and three copies of a `target_simulator === 'Brian2'` ternary, which let the cost estimate and the launched task diverge. Both now come from one `resolveSimulationLaunchTarget()`, built on a `LAUNCH_TARGET_BY_SCALE` table; `null` means "not launchable via obi-one" and also drives job-id recovery and the task config/log viewers.

Me-model campaigns now get the cost-confirmation modal they previously bypassed (obi-one gained an estimator for them).

Also fixes a small pre-existing bug in the touched block: declining the consent modal left `simRequestInProgress` set, so the Launch button stayed stuck in progress.


pgetta:
Relates to https://github.com/openbraininstitute/prod-circuit-simulation/issues/231.

src/features/scan-config/template.tsx
```diff
@@ -223,7 +223,6 @@ function ScanConfigTemplateContent({
           projectId={projectId}
           campaignOriginAction={campaignOriginAction}
           isCampaignIdChanged={isCampaignIdChanged}
-          taskTypeBindings={taskTypeBindings}
```
bilalesi: this sounds weird, this get passed down to all the results tabs components e.g: 
```js
<BuildTab
            isCampaignIdChanged={isCampaignIdChanged}
            campaignOriginAction={campaignOriginAction}
            campaignId={campaignId}
            taskTypeBindings={taskTypeBindings}
/>
```

bilalesi: ignore it please, it's just github display

bilalesi (approved):

PR #1842, opened 2026-07-29
feat(202): intracellular efeature extraction

bilalesi:


bilalesi:
@openbraininstitute/webapp can someone have a look please 
(PO approved it)

src/features/scan-config/components/ui-elements/select-efeatures-by-protocol/index.tsx
```diff
@@ -0,0 +1,286 @@
...
+   */
+  const offeredDefs = useMemo(() => {
+    // whatever is already selected stays on screen no matter what the endpoint says: the
+    // selection is what gets submitted, so it must remain visible (and checked) while the
+    // properties query is in flight, after it fails, and once the editor goes read-only on
+    // generate — a card that disappears reads as a selection that was silently dropped
+    const shown = new Set(selection.protocols.map((protocol) => protocol.type));
+    for (const name of discoveredProtocols ?? []) shown.add(name);
+
+    return protocolDefs.filter((def) => shown.has(def.typeName));
```
pgetta, line 139: `discoveredProtocols` arrives from the endpoint as eCode names (`IDrest`, `IV` - see the [fixture](https://github.com/openbraininstitute/core-web-app/blob/0249b360632bf4e734e606888ffa895f0c711d62/src/__tests__/scan-config/electrical-cell-recording-properties.test.ts#L73-L79)), while `def.typeName` is the class name (`IDRestProtocol`, `IVProtocol`), so this filter never matches anything the endpoint reports. With a fresh config the widget falls through to "No protocols found in the selected recordings" even when the recordings do contain them.

[`ecodeNameFromTypeName`](https://github.com/openbraininstitute/core-web-app/blob/0249b360632bf4e734e606888ffa895f0c711d62/src/features/scan-config/components/ui-elements/select-efeatures-by-protocol/helpers.ts#L189-L198) looks like it was written for exactly this, and its doc says the comparison also has to be case-insensitive - right now it only feeds the label.

src/features/scan-config/components/ui-elements/select-efeatures-by-protocol/index.tsx
```diff
@@ -0,0 +1,286 @@
...
+  );
+  const errorPrefix = errorPathPrefix ?? ScanConfigUIElementDict.SelectEFeaturesByProtocol;
+
+  const selectedByType = useMemo(
+    () => new Map(selection.protocols.map((protocol) => [protocol.type, protocol])),
+    [selection.protocols]
+  );
+
+  const amplitudesFor = (def: TProtocolDef) =>
+    recordingProperties?.AmplitudesByProtocol?.[def.typeName];
```
pgetta, line 161: Same here: `AmplitudesByProtocol` is keyed by eCode name (`{ IDrest: [0.1, 0.2], IV: [-0.02] }`), so a lookup by `def.typeName` always gives `undefined`. Every protocol then goes through `makeFilledProtocolValue(def, undefined)` with `extraction_amplitudes: []`, and the settings panel permanently shows "No amplitudes discovered for this protocol in the selected recordings".

src/features/scan-config/template.tsx
```diff
@@ -128,7 +128,7 @@ function ScanConfigTemplateContent({
     workflowSessionSelection,
     resolveFromIdType: resolveSessionFromIdType,
   });
-
+  console.log('@@@schema', schema);
```
pgetta: I guess some leftovers from local debugging, or do you want to keep this? The dereferenced schema is fairly big, so it would land in the console on every render of the editor.

src/features/scan-config/components/model-preview/helpers.ts
```diff
@@ -67,37 +67,76 @@ export function resolveHostNeuronOpacity(options: {
...
+const WorkflowOwnedPreviews = new Set<TRightPreviewMode>([RightPreviewModeDict.EFeatures]);
+
 /**
  * Resolve which exclusive preview the right column should show.
  *
- * Entity mini-detail wins over model previews so browse selection stays visible.
+ * Entity mini-detail wins over model previews so browse selection stays visible, except over the
+ * {@link WorkflowOwnedPreviews} when it was opened by a selection rather than a click.
  */
 export function resolveRightPreviewMode(options: {
```
pgetta, line 96: The ordering reads fine on its own, but I'm a bit concerned about what it costs in the e-features workflow. Opening a protocol or feature settings form flips the right column to `Settings`, which unmounts `EFeaturesPreviewPanel` and with it `TracesTab`, so `nwbWorkerRegistry.release()` runs and the session is disposed after [`DISPOSE_GRACE_MS`](https://github.com/openbraininstitute/core-web-app/blob/0249b360632bf4e734e606888ffa895f0c711d62/src/features/ephys-viewer/hooks/nwb-worker-manager.ts#L34-L38) - 1.5s. So anything longer than that spent in a settings form pays a full NWB re-download on close, and by that same note the file is a few hundred megabytes.

Possible options:

* Rendering the settings pane beside the preview rather than in place of it (split or overlay), so the traces stay mounted - I think that is the best one, given the traces are exactly what these settings are meant to be read against.
* Letting `Settings` win only over the modes that are not workflow-owned, the way `WorkflowOwnedPreviews` already does for the entity preview.
* Raising the grace period for this case.

Or maybe I'm missing something and the session is kept alive elsewhere.

src/features/scan-config/components/ui-elements/select-efeatures-by-protocol/helpers.ts
```diff
@@ -0,0 +1,497 @@
...
+): string | null {
+  const base = (schema as unknown as Record<string, unknown>).efel_figures_base_url;
+  if (typeof base !== 'string' || !base) return null;
+
+  const image =
+    typeof variant.efel_feature_image === 'string' && variant.efel_feature_image
+      ? variant.efel_feature_image
+      : `${efelNameFromDef(variant)}.png`;
+
+  if (image.startsWith('null')) return null;
```
pgetta: The header of [`efel-figures.ts`](https://github.com/openbraininstitute/core-web-app/blob/0249b360632bf4e734e606888ffa895f0c711d62/src/features/scan-config/components/ui-elements/select-efeatures-by-protocol/efel-figures.ts#L7-L15) says the opposite of what this does: the file names don't line up with eFEL keys (`AP_Amplitude.png` against `AP_amplitude`) and deriving a URL per feature 404s for all but one of the 26 classes, which is given as the reason the directory is listed instead. This derives one anyway, and `EFeatureFigure` hides the image on error, so the illustration in a feature's settings form ends up blank nearly every time. Is one of the two meant to go away?

The `startsWith('null')` guard also looks a little risky - it would reject a genuine `nullcline.png`, and the only source of `null` is the template right above it, so checking `efelNameFromDef` before building the string would be safer.

src/features/scan-config/components/ui-elements/index.tsx
```diff
@@ -127,18 +129,79 @@ export function UIElementRender({
...
+          config={config}
           disabled={disabled}
-          value={value as number | null | number[]}
-          onChange={(value) => {
-            setState({ ...state, [k]: value });
-          }}
+          errorPathPrefix={errorPathPrefix}
+          // the shared renderer is defined in this module; passing it down keeps the widget
+          // from importing its own barrel
+          renderField={({
```
pgetta, line 187: This renderer matches on `paramSchema.ui_element` and falls through to `null` otherwise, but the nested fields carry no `ui_element` of their own - which is what [the type doc](https://github.com/openbraininstitute/core-web-app/blob/0249b360632bf4e734e606888ffa895f0c711d62/src/features/scan-config/types.ts#L267-L272) states too. In the fixture that covers every feature override field (`VoltageBaseFeature.stim_start`) and all four of `SAHPProtocol`'s timings.

So opening sAHP's settings, or any feature's, gives `STIM_START` / `STIM_MID` labels with empty space where the input should be, and the Remove button as the only working control. Probably `renderField` needs a fallback driven by the field's `type` rather than its `ui_element`.

src/features/scan-config/outputs/use-generated-outputs.ts
```diff
@@ -0,0 +1,124 @@
...
+        const data = query.state.data ?? undefined;
+        const strategy = getOutputStrategyById(data?.strategyId);
+        if (!pollingEnabled || !strategy?.refetchInterval) return false;
+
+        return strategy.refetchInterval({ data, dataUpdateCount: query.state.dataUpdateCount });
+      },
+    })),
+  });
+
+  const files = useMemo(
```
pgetta: `useQueries` returns a new array on every render, so `[queries]` never memoizes - the strategies get re-resolved and the files re-sorted each time, and `files` gets a new identity for every consumer downstream. `listingTypes` below has the same problem as an effect dep, with the `invalidated.current` string doing the actual work of not re-invalidating.

Probably worth keying the memo on the resolved data (the ids plus `strategyId`) and deriving `listingTypes` inside it.

src/ui/molecules/motion-tabs/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+  className,
+}: {
+  value: string;
+  children: ReactNode;
+  className?: string;
+}) {
+  const { value: current } = useMotionTabs();
+  const reduce = useReducedMotion();
+
+  if (current !== value) {
```
pgetta, line 214: I wonder, would it make sense to not render the inactive panels at all rather than hide them? As it stands every panel mounts and runs its effects - in the e-features preview the GitHub figures fetch fires on first paint while Traces is the default tab. For a general-purpose primitive an eager mount is a bit surprising, since deferring is usually the point of tabs.

Though it is also what keeps the NWB session alive while the user reads the Features tab, so if that is deliberate it would be good to say so in a comment.

Btw, the panels carry no `role="tabpanel"` / `aria-labelledby`, so the `role="tablist"` and `role="tab"` on the triggers have nothing to point at ([MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tabpanel_role)).

src/features/scan-config/components/ui-elements/select-efeatures-by-protocol/efel-figures.ts
```diff
@@ -0,0 +1,114 @@
...
+ * enumerated.
+ */
+export function useEFelFigures(schema: ConfigSchema) {
+  const baseUrl = efelFiguresBaseUrl(schema);
+  const listingUrl = baseUrl ? githubContentsApiUrl(baseUrl) : null;
+
+  const { data, isLoading } = useQuery({
+    queryKey: ['efel-figures', listingUrl],
+    queryFn: async (): Promise<TEFelFigure[]> => {
+      const response = await fetch(listingUrl as string, {
```
pgetta, line 100: Not a call for action for this PR, just for awareness. The unauthenticated contents API is 60 requests/hour per source IP, so everyone behind one office egress shares that budget and every fresh page load starts with a cold cache. Once it runs out the tab only says the illustrations couldn't be listed, and with `retry: false` there is no way for the user to tell why.

Later on it would be nice to proxy this through our own API and cache it there, or ship the manifest with the app.

src/ui/segments/workflows/config/activities/index.ts
```diff
@@ -31,7 +30,6 @@ export const ActivityRegistry: Record<TActivityValue, TActivityEntry> = {
     name: 'Extraction',
     order: 3,
     disabled: false,
-    requiredFeatures: [extractionActivityFlag.key],
```
pgetta, line 34: I can see from the tests that the extraction activity is meant to be on for everyone now, so dropping the gate is intended.

But then [`anyFeatureSatisfied` / `requiredAnyFeatures`](https://github.com/openbraininstitute/core-web-app/blob/0249b360632bf4e734e606888ffa895f0c711d62/src/ui/segments/workflows/config/helpers.ts#L56-L68), added in this PR, aren't set on any registry entry - is that groundwork for a later flag, or was this entry supposed to use it?

src/features/ephys-viewer/ephys-viewer.tsx
```diff
@@ -92,37 +174,47 @@ export default function EphysViewer({
...
-              onRepetitionClick={showRepetitionDetails}
-              onProtocolChange={setProtocol}
+              onRepetitionClick={(protocolClosure, repetitionClosure) => () =>
+                dispatch({
+                  type: 'showRepetition',
+                  protocol: protocolClosure,
+                  repetition: repetitionClosure,
+                })
+              }
+              onProtocolChange={(next) => dispatch({ type: 'setProtocol', protocol: next })}
```
pgetta: With `protocol` supplied by the host this dispatch has no visible effect, since `const protocol = protocolOverride ?? state.protocol` discards it on the next render. Not reachable today because the e-feature editor also turns the toggle off, but a caller passing `protocol` while leaving `showViewModeToggle` on gets a select that snaps straight back. Maybe disable it when the protocol is host-driven?

src/features/ephys-viewer/ephys-viewer.tsx
```diff
@@ -17,44 +17,120 @@ import { TraceProvider } from '@/features/ephys-viewer/trace-context';
...
+export function viewerReducer(state: TViewerState, action: ViewerAction): TViewerState {
+  switch (action.type) {
+    case 'setView':
+      return { ...state, view: action.view };
+    case 'setCellId':
+      return { ...state, cellId: action.cellId };
+    case 'setProtocol':
+      return { ...state, protocol: action.protocol };
+    case 'showRepetition':
+      // a single transition, so the details view can never open on a protocol/repetition mismatch
```
pgetta, line 63: Oh nice, making this one transition rather than two is a neat way to make the mismatch unrepresentable.

pgetta (commented):
Nice work, Bilal, and thanks for the comments throughout - they made a change this size much easier to follow.

I've left a number of comments, a few of them worth a look before merging.

src/features/scan-config/components/ui-elements/select-efeatures-by-protocol/index.tsx
bilalesi: no code change, but the test pretended the server says `IDrest`, and that lie is what caused this comment, I fixed the test to use real names.

src/features/scan-config/components/ui-elements/select-efeatures-by-protocol/index.tsx
bilalesi: no code change; same test fixture fixed.

src/features/scan-config/template.tsx
bilalesi: deleted

src/features/scan-config/components/model-preview/helpers.ts
bilalesi: took your first suggestion

src/features/scan-config/components/ui-elements/select-efeatures-by-protocol/helpers.ts
bilalesi: deleted the guessing `efelFigureUrl`
the Features tab still lists the 6 real pictures, header now states the measured numbers instead of the old "26 classes"

src/features/scan-config/components/ui-elements/index.tsx
bilalesi: fixed the fixture to match reality

src/features/scan-config/outputs/use-generated-outputs.ts
bilalesi: results are now cached against a key built from the data itself (ids, strategy, assets)

src/ui/molecules/motion-tabs/index.tsx
bilalesi: panels now have role="tabpanel" and point back at their button

src/ui/segments/workflows/config/activities/index.ts
bilalesi: it should be useful for future feature flags

bilalesi:
thanks Pavlo, 
comments are addressed

pgetta (approved):

PR #1845, opened 2026-07-30
feat(circuit-viz): display afferent synapses in the single-cell 3D view

pgetta:
Renders afferent synapses in the Simulate/Synaptome 3D view.

Ticket: openbraininstitute/prod-singlecell-simulation#274 (cross-repo, so no auto-close keyword). Picks up @tolokoban's WIP — first commit, authorship preserved.

## What changed

- Synapses are threaded through the `SmallCircuitSource` contract into `CircuitViz`. `viewer-layout`, where the WIP wired them, has been legacy since #1824. Only the SONATA source populates them.
- All synapses render, not only soma ones. The WIP skipped the rest while still allocating buffer slots for them, so every dendritic and axonal synapse would have drawn at the world origin.
- Soma synapses are projected onto the drawn mesh. SONATA computes `afferent_surface_*` against a *spherical* soma while the SWC describes a stack of cones, so the raw coordinates sink inside the mesh and disappear. New `sdf.ts` and `morphology-surface.ts` rebuild the surface morphoviewer actually paints and slide the point onto it.
- `rescueOffSurface()` pulls back synapses SONATA places on that sphere but labels as the first neurite section — 3 of 4245 here. Confined to the soma's envelope: testing every synapse against a whole morphology is ~300ms per cell on the load path, against ~6ms this way.
- Soma detection is `afferent_section_id === 0`, per MWolfR on the ticket. `afferent_section_type` is the documented signal, but several released circuits never emit a 1.
- Adds `gl-matrix`, for the cell orientation quaternion.

## One thing here is deliberately wrong

`drawnRadiusFactor()` projects soma synapses onto a soma slightly smaller than the real one, because that is what gets drawn. `TgdPainterSegments` builds each capsule's frame from an un-normalised cross product, so a segment is drawn `sqrt(1 - (fakeY·Z)²)` thinner than its stated radius — near 0.71 for a typical soma, which leaves a correctly-placed synapse standing up to 2.4µm proud of it on screen. Neurites take the same 29% but hide it inside 0.06µm. It also explains why the July screenshots failed on some cells and not others.

Shrinking errs in the safe direction: the shrunken capsule is strictly inside what is drawn, so a synapse can come out slightly buried but never floating. **Delete this once tgd ships `normalize(cross(fakeY, Z))`.** I'll report that upstream.

## Testing

29 Vitest tests in `morphology-surface.test.ts`, covering the SDF union and its bounding-sphere reject, the degenerate root, the shrink factor including its discontinuity, and the rescue's scoping. Checked in the browser on `nbS1-O1__202247__cADpyr__L5_TPC_A`: 4245 synapses, 38 on the soma, worst residual 0.404µm, 4 rescued, nothing left hanging in empty space.

## Not covered

- Neurite synapses keep their raw SONATA coordinates. A bounded sample is measured against the drawn surface and logged, so the next "these look wrong" arrives with numbers.
- Legibility is untouched: only 217 of 4245 synapses sit within 30µm of the soma, and point sprites carry no depth cue.

## To do

- Moving synapse projection logic into the morphoviewer
- Decide what to do with TGD - dependency of the morphoviewer


src/features/scan-config/components/model-preview/viewer-layout/circuit-loader/morphology-surface.ts
```diff
@@ -0,0 +1,218 @@
+import {
+  add,
+  length,
+  scale,
+  sdfCapsuleWithNormal,
```
bilalesi, line 5: i am not sure, but i feel this can go to the morphoviewer lib ?

bilalesi (dismissed):

src/features/scan-config/components/model-preview/viewer-layout/circuit-loader/morphology-surface.ts
pgetta: Good point, will address in a separate PR.

PR #1846, opened 2026-07-30
feat(entitycore): multipart asset uploads for files of 20MB and above

pgetta:
## What

Adds S3 multipart upload support to the entitycore API client. `createAsset` now routes files at or above 20MB (`MULTIPART_UPLOAD_THRESHOLD`) through the multipart flow; smaller files keep the existing single FormData POST. Call sites are unchanged.

The backend caps the FormData path at 150MB (`API_ASSET_POST_MAX_SIZE`), so larger files could previously not be uploaded at all, and mid-size uploads had no parallelism, retries or progress reporting.

With the 150MB cap gone, the shared `AssetUpload` default limit is raised from 500MB to 10GB; flows with explicit `maxSize` overrides (em-cell-mesh 500MB, ephys 75MB, morphology 5MB) are unaffected.

## How

The multipart flow (`src/api/entitycore/queries/assets/multipart.ts`) follows the entitycore contract:

1. **Hash** — streaming SHA-256 of the file via `hash-wasm` (required by the initiate endpoint; `crypto.subtle` cannot stream, and files can be multi-GB). Constant memory via 4MB chunks; `hash-wasm` is dynamically imported so it stays out of the shared asset-queries chunk.
2. **Initiate** — `POST .../assets/multipart-upload/initiate` creates the asset in `uploading` status and returns presigned part URLs (`upload_meta`).
3. **Part PUTs** — plain `fetch` to the presigned URLs (no Authorization header), 5 parts in parallel via `p-limit`, per-part retry (3 attempts, exponential backoff + jitter) on 5xx/429/network errors, fail-fast on 403. No ETag bookkeeping — the server lists parts in S3 itself on completion.
4. **Complete** — `POST .../assets/{id}/multipart-upload/complete` verifies and assembles; asset flips to `created`.

Abort (`AbortSignal`) or terminal failure cancels in-flight parts and best-effort-deletes the dangling `uploading` asset via the new `deleteAsset` wrapper. Optional `onProgress` reports `hashing`/`uploading` phases with byte counts.

Also included: `entityAssetsPath()` helper replacing the inline `/{entity-route}/{id}/assets` copies, `isAssetContentType` type guard, `AssetStatus.UPLOADING` + upload-meta types, and part failures throwing structured `ApiError`.

## Deployment prerequisite

Browser part PUTs go directly to the storage endpoint, so the entitycore bucket needs a CORS rule allowing `PUT` from the app origins — without it the preflight `OPTIONS` gets a 403 from S3. The bucket likely already has a `GET` rule (browser downloads follow presigned redirects today); `PUT` needs adding to it. Local MinIO allows all origins by default.

## Notes

- The initiate request sets `content-type: application/json` explicitly — the `ApiClient` sends no default content-type (its constructor fallback is dead code), so a stringified body otherwise goes out as `text/plain`, which entitycore rejects.
- The multipart initiate endpoint accepts no `meta` and requires `label`; `createAsset` guards both with clear errors (no current call site hits either).
- Directory multipart endpoints exist on the backend but have no frontend usage — out of scope.
- Pre-existing `getAsset` builds its URL without `kebabCase`, unlike every sibling — left untouched, worth a separate look.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
     "framer-motion": "^12.38.0",
     "fuse.js": "^7.3.0",
     "h5wasm": "^0.7.9",
+    "hash-wasm": "^4.12.0",
     "html2canvas": "^1.4.1",
     "input-otp": "^1.4.2",
     "jotai": "~2.19.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -16,7 +16,7 @@ importers:
         version: 6.2.2(react-dom@19.2.6(react@19.2.6))(react@19.2.6)
       '@ant-design/nextjs-registry':
         specifier: ^1.3.0
-        version: 1.3.0(@ant-design/cssinjs@1.24.0(react-dom@19.2.6(react@19.2.6))(react@19.2.6))(antd@5.29.3(date-fns@4.1.0)(moment@2.30.1)(react-dom@19.2.6(react@19.2.6))(react@19.2.6))(next@16.2.6(@babel/core@7.29.0)(@opentelemetry/api@1.9.1)(@playwright/test@1.60.0)(babel-plugin-react-compiler@1.0.0)(react-dom@19.2.6(react@19.2.6))(react@19.2.6)(sass@1.99.0))(react-dom@19.2.6(react@19.2.6))(react@19.2.6)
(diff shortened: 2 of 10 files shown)


g-bar (dismissed):
LGTM

g-bar (approved):

PR #1863, opened 2026-08-09
feat(770): pick morphology locations on single-scale circuits

bilalesi:
clicking a branch in the 3D viewer adds an explicit morphology location to the form, and editing the form moves the marker in 3D.

single circuits now load their morphology from OBI-One rather than parsing the SONATA asset in the browser: only the server can number sections the way the simulation does, and a location is meaningless against any other numbering. That move costs the two things the browser loader also supplied, so both are restored from the server: afferent synapses come from the new /circuit/viz synapses endpoint, and a lone cell is coloured by section type again.

colour by section type only when one cell is on screen. With several, telling the cells apart is the job, so each keeps a flat colour.

the loader-strategy layer goes with it. It resolved a scale to one of two sources; every scale now takes the same one, leaving a single-value enum behind a resolver that ignored its argument.

viewer settings gain a marker radius and a section-label toggle. The three viewer components are renamed for what they show rather than how they were built: MorphoViewerSingleNeuron, MorphoViewerCircuitMultipleNeurons and MorphoViewerCircuitMultipleNeuronsSomaOnly.

requires @openbraininstitute/morphoviewer `0.32.2`


src/features/scan-config/components/model-preview/use-morphology-location-selection.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  const selected = useMemo(() => {
+    if (!locationsKey) return [];
+    const locations = JSON.parse(locationsKey) as IStoredLocation[];
+    // `sectionName` in the viewer is OBI-One's raw morphio id as a string, and
+    // `sonata_section_id` is that plus one. Undo the shift to address the drawn section.
+    return cells.flatMap((cell) =>
+      locations.map((location) => ({
+        cellId: cell.id,
+        sectionName: location.section_id === 0 ? 'soma' : String(location.section_id - 1),
```
pgetta: I'm curious about the `- 1` here. The schema doc you added says `sonata_section_id` is "SONATA global section id: 0 for the soma, then nrn_order neurites", while `id` is "OBI-One's raw morphio id" - those are two different orderings, and if they were always one apart the backend field would not have been needed in the first place.

Where nrn_order and morphio order do diverge (NEURON groups sections by soma / axon / basal / apical), the marker lands on a different branch than the one that will be simulated, and where the derived name does not exist at all `CellSectionIndex.getPointAtOffset` returns null and the marker is dropped silently - so the row sits in the list on the left but nothing shows up in 3D.

The `/circuit/viz` sections carry both ids, so a `sonata_section_id -> id` map built where the tree is built would address the exact section without assuming anything about the ordering. Or maybe the relation holds by construction and I'm missing it.

src/features/scan-config/components/model-preview/use-morphology-location-selection.ts
```diff
@@ -0,0 +1,232 @@
...
+        sonataSectionId: location.section_id,
+      }))
+    );
+  }, [locationsKey, cells]);
+
+  const onPick = useCallback(
+    (pick: MorphoViewerMorphologyLocationPick) => {
+      if (!onConfigChange || !selectedEntry || pick.sonataSectionId === undefined) return;
+
+      // Removing an existing marker is always allowed; only adding is restricted.
```
pgetta: `sonata_section_id` is optional in the schema, for the good reason you explain there, so against an older OBI-One deployment `pick.sonataSectionId` is undefined and every click on the neuron returns here silently. From the user's side the viewer simply does not react, while the cursor and the hover popover keep promising that a click will add something.

We already show a `message.info` for a non targetable section right below, so probably worth a similar hint here.

src/features/scan-config/components/model-preview/use-morphology-location-selection.ts
```diff
@@ -0,0 +1,232 @@
...
+        // viewer's own report rather than on the click's rounded offset, which would not
+        // reliably land back on the stored value.
+        const next = removed
+          ? existing.filter(
+              (location) =>
+                !(
+                  location.section_id === removed.sonataSectionId &&
+                  location.offset === removed.offset
+                )
+            )
```
pgetta: Two things about this branch:

* The match is on value rather than on identity, so two rows holding the same `section_id` and `offset` are both removed by a single click. It is reachable: the OFFSET input is freely editable and nothing dedupes the list, so setting the second row to the first row's offset gives exactly that. Carrying an index or a stable id on the marker would keep the removal to the one that was clicked.
* When the block holds a single location, `next` is empty and the updater returns `previous` (line 177), so the click does nothing and explains nothing - right after the popover told the user "Click again to remove". A short `message.info` about the backend minimum would help; in the list editor the constraint is at least visible, since the delete button is hidden.

src/features/scan-config/components/model-preview/use-morphology-location-selection.ts
```diff
@@ -0,0 +1,232 @@
...
+          },
+        } as Config;
+      });
+    },
+    [onConfigChange, selectedEntry, message]
+  );
+
+  // Coalesced to one update per frame: the viewer republishes on every repaint, and an orbit
+  // is a repaint per frame. Without this, a drag would queue a React render per frame.
+  const pendingLabels = useRef<MorphoViewerMorphologyLocationLabel[] | null>(null);
```
pgetta: Nice, I like the per frame coalescing here - otherwise an orbit would be one React render per repaint.

src/features/scan-config/components/circuit-viz/build-morpho-tree.ts
```diff
@@ -0,0 +1,92 @@
...
+            prevNode = sNode;
+          }
+        }
+      }
+    } else if (sec.parent_id !== null && sec.parent_id !== 'soma') {
+      prevNode = terminalNodes.get(sec.parent_id) || null;
+    }
+
+    const isSoma = sec.id === 'soma';
+    const startIdx = sec.parent_id === null || sec.parent_id === 'soma' ? 0 : 1;
```
pgetta, line 47: This is moved code, so not introduced here, but it now feeds the offset that gets written into the config, which makes it worth a second look.

For a section attached to the soma `startIdx` is 0, so `points[0]` is pushed as a child of the nearest soma node, and morphoviewer's `collectSegments` then emits a soma -> dendrite segment carrying the dendrite's `sectionId`. `CellSectionIndex.add` keys it by that id with `segmentIndex` 0, so `computeSectionOffset` divides by a total distance that includes the stub and counts it whole for every click further along.

With a soma radius around 8µm and a 50µm first order dendrite, clicking the middle of the branch stores ~0.57 instead of 0.5, and that is the number the simulation gets. The viewer already recognises this stub when drawing it (`stubOffSoma` in the tree factory), so excluding it from the section index, or from the length sum in `computeSectionOffset`, would make the stored offset mean what it says. Please correct me if the stub is supposed to count as part of the section.

pgetta (commented):
Nice work, Bilal, this is a big one.

I've left a bunch of comments, quite a few of them just questions.

src/features/scan-config/components/circuit-viz/circuit-viz.tsx
```diff
@@ -79,52 +74,68 @@ interface CircuitVizProps {
...
+  const source = useObiOneVizSource(sourceOptions(props));
+  const synapses = useCircuitSynapses(props.circuit);
+  const withSynapses = useMemo(() => ({ ...source, synapses }), [source, synapses]);
+  return <CircuitVizView {...props} source={withSynapses} clearSequentialOnAxonToggle />;
+}
+
+function CircuitVizPlain(props: CircuitVizProps) {
+  const source = useObiOneVizSource(sourceOptions(props));
   return <CircuitVizView {...props} source={source} clearSequentialOnAxonToggle />;
 }
```
pgetta: Both wrappers pass `clearSequentialOnAxonToggle` now, so the `= false` default below and the `if (!clearSequentialOnAxonToggle) return` guard can never be reached - the prop is always on and could go away.

And the two components differ only by `useCircuitSynapses`, which is a `useQuery`: `enabled: circuit.scale === CircuitScaleDictionary.Single` would keep the hook order legal without the split. Not a blocker, just two components and a prop we would not have to maintain.

src/features/scan-config/components/circuit-viz/circuit-viz.tsx
```diff
@@ -146,8 +157,17 @@ function CircuitVizView({
   source,
   clearSequentialOnAxonToggle = false,
   errorActions,
```
pgetta: `errorActions` has no caller left once `CircuitVizSonata` is gone, so this prop and the block rendering it are dead.

The bigger part is that the error path got narrower with it: `source.error` only carries the nodes query error, while a failed morphology fetch throws inside `loadCell`, and morphoviewer catches it, logs it and calls `onCellLoaded(null)`. Progress still reaches 1 and `loading` clears, so a transient 5xx on `/circuit/viz/{id}/morphologies/...` leaves the user with an empty canvas, no message and no retry - and the SONATA download button that used to soften that case is gone too. Would be nice to surface a load failure here.

src/features/scan-config/components/circuit-viz/sources/use-circuit-synapses.ts
```diff
@@ -0,0 +1,228 @@
...
+ *
+ * @param circuit - The circuit whose afferent synapses to load.
+ * @returns One coloured group per edge population; empty until loaded.
+ */
+export function useCircuitSynapses(circuit: ICircuit): SmallCircuitSynapseGroup[] {
+  const { virtualLabId, projectId } = useWorkspace();
+  const queryClient = useQueryClient();
+  const circuitId = circuit.id;
+
+  const { data: projected } = useQuery({
```
pgetta, line 53: The query error is dropped here, and with our `retry: false` default plus `STATIC_RESOURCE_QUERY_OPTIONS` (`staleTime` / `gcTime: Infinity`, no refetch on mount / focus / reconnect) a single failure sticks for the whole session.

So if `/circuit/viz/{id}/synapses` 404s on a deployment that does not serve it yet - the same situation the optional `sonata_section_id` guards against - or 5xxs once, `synapses` stays empty and a single neuron circuit looks exactly like one that genuinely has no afferent synapses, with nothing in the console either. The hook already imports `logError`, so probably worth calling it on failure at least.

src/features/scan-config/components/model-preview/circuit-preview.tsx
```diff
@@ -201,6 +200,9 @@ export function CircuitPreview({
     useCircuitColorBy(enableVisualization ? circuit : undefined, {
       supportsAxons,
       supportsElectrodes: enableElectrodes && electrodesAvailable,
+      // The marker slider is only meaningful while the form is editing morphology locations,
+      // which is also the only time markers are on screen.
+      supportsMorphologyLocations: selectedRootElement === MORPHOLOGY_LOCATIONS_CONFIG_KEY,
```
pgetta: `supportsMorphologyLocations` checks the root element only, while `useMorphologyLocationSelection` additionally requires the entry's block to be `ExplicitMorphologyLocations`. For the sampling block types that means the settings menu gains "Location labels" and "Location marker size", both live and both persisted into the viewer config, while the hook returns no selection and no labels - so nothing on screen reacts to them. Probably the block type should be part of this condition too.

src/features/scan-config/components/model-preview/circuit-preview.tsx
```diff
@@ -319,6 +321,14 @@ export function CircuitPreview({
...
             electrodeRadius={config.electrodeRadius}
             features={vizFeatures}
+            morphologyLocations={{
+              config: scanConfig,
+              onConfigChange: setConfig,
+              selectedRootElement,
+              selectedEntry,
+              markerRadius: config.morphologyLocationRadius,
+              showLabels: config.showMorphologyLocationLabels,
+            }}
```
pgetta, line 349: I wonder about routing this through the electrodes group: `config`, `onConfigChange`, `selectedRootElement` and `selectedEntry` are destructured out of `electrodes` (line 128), while `ElectrodeOverlayOptions` is documented as "Everything the electrode-overlay layer needs, in one place".

In practice a host that renders `CircuitPreview` without `electrodes` - `features/entities/circuit/detail-view.tsx:48` for instance - gets `config` / `onConfigChange` undefined and picking silently unavailable, with no prop that says so. The form binding (live config plus the selected block) looks like a concern of its own that both features need, so maybe it deserves its own group.

src/features/scan-config/components/ui-elements/morphology-location-selection.tsx
```diff
@@ -0,0 +1,125 @@
...
+          className="flex items-end gap-2 border border-gray-200 p-2"
+        >
+          <div className="flex flex-1 flex-col gap-1">
+            <label className="text-neutral-3 text-xs" htmlFor={`${fieldId}-section-${index}`}>
+              SECTION ID
+            </label>
+            {/* Always disabled, never only when the form is: the id comes from the viewer. */}
+            <InputNumber
+              id={`${fieldId}-section-${index}`}
+              className="w-full [&_input]:cursor-default [&_input]:font-medium [&_input]:text-neutral-1"
```
pgetta: `text-neutral-1` is `#f5f5f5` and the input is `variant="filled"`, which antd fills with `rgba(0,0,0,0.04)` over white, so the section id would end up near white on near white (unless antd's disabled colour wins the specificity fight, in which case the class does nothing at all). The labels around it use `neutral-3` and the rest of the form is on the light side, so I suspect `neutral-10` was meant here. Could you check how it renders on your side?

src/features/scan-config/components/circuit-viz/morphology-location-help.tsx
```diff
@@ -0,0 +1,29 @@
+'use client';
+
+import { ViewerHelpRow, ViewerHelpTooltip } from './viewer-help-tooltip';
+
+/** Top-left chrome help for picking morphology locations. */
+export function MorphologyLocationHelp({ container }: { container?: HTMLElement | null }) {
```
pgetta, line 6: From what I can see nothing imports this component - `circuit-viewer-chrome` renders `ElectrodeInteractionHelp` only, so this help never appears. A pity, because it is the only place where the click to remove gesture and the hand cursor are explained, and the hover popover does not cover them. Was it meant to be rendered next to the electrode help while picking is active?

src/features/scan-config/components/model-preview/viewer-layout/viewer-layout.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import { LoadingNeuronSpinner } from '@/components/neuron-viewer';
-import { MorphoViewerSmallCircuit } from '@/morpho-viewer';
+import { MorphoViewerCircuitMultipleNeurons } from '@/morpho-viewer';
```
pgetta, line 4: With `useSonataAssetSource` and the `useDownloadHandler` import in `circuit-viz` gone, is anything still reaching this folder? From what I can see `ViewerLayout`, `CircuitLoader` and `hooks.ts` are now referenced only from inside `viewer-layout/` itself, which would leave the h5wasm SONATA parsing, `swc.ts`, `report.ts` and `circuit-config.ts` as dead weight - and `circuit-loader.ts` was gutted here (`getAfferentSynapses`, `probeNeuriteSynapses`, `hasDataset`) rather than removed.

If it is indeed unreachable, deleting the folder in this PR would be nicer than keeping it compiling. The doc comment right below still points at `resolveSmallCircuitLoaderKind`, which this PR deletes, and `model-preview/index.tsx:91` still describes the SONATA / OBI-One split as well.

src/features/scan-config/components/color-by/palette.ts
```diff
@@ -11,6 +11,22 @@ import type { CategoricalLegendEntry, ColorMapping, ContinuousLegend } from './t
...
 export const DEFAULT_NEURON_COLOR = '#2b5cd9';
 
+/**
+ * Per-section-type neuron colors, for a cell shown on its own.
+ *
+ * With one cell on screen there is nothing to tell apart, so color carries which part of the
+ * neuron a branch belongs to. Several cells use a flat color each instead. Values match the
+ * SONATA circuit loader.
+ */
+export const SECTION_TYPE_COLORS = {
```
pgetta, line 21: `circuit-loader.ts` still holds the same literal (lines 205-212), so the palette now lives in two places kept in sync by a comment. Importing it there would be nice - or it disappears on its own if that folder goes away, see my other comment.

Same idea for `DEFAULT_MORPHOLOGY_LOCATION_RADIUS` in `use-viewer-config.ts:41`: morphoviewer exports exactly that value as `DEFAULT_LOCATION_MARKER_RADIUS`, so we could re-export it instead of restating it and keeping the doc comment true by hand.

src/features/scan-config/components/color-by/viewer-controls-menu.tsx
```diff
@@ -196,6 +207,31 @@ export function ViewerControlsMenu({
...
+              <Slider
+                min={1}
+                max={30}
+                step={1}
+                value={morphologyLocationRadius}
+                onChange={onMorphologyLocationRadiusChange}
+                tooltip={{ formatter: null }}
+              />
+            </div>
+          )}
```
pgetta, line 221: Later on would be nice to de-duplicate: this is the third copy of the same labelled slider block (electrode size, neuron opacity, and now the marker size), differing only by label, bounds and value. A small `MenuSlider` next to `MenuRow` / `MenuButton` would cover all three, the same way you pulled `ViewerHelpTooltip` / `ViewerHelpRow` out in this PR. Not a call for action for this one.

src/features/scan-config/components/circuit-viz/build-morpho-tree.ts
bilalesi: fixed

src/features/scan-config/components/circuit-viz/circuit-viz.tsx
bilalesi: fixed

src/features/scan-config/components/circuit-viz/sources/use-circuit-synapses.ts
bilalesi: fixed

src/features/scan-config/components/model-preview/circuit-preview.tsx
bilalesi: fixed

src/features/scan-config/components/circuit-viz/morphology-location-help.tsx
bilalesi: fixed

src/features/scan-config/components/ui-elements/morphology-location-selection.tsx
bilalesi: fixed

src/features/scan-config/components/model-preview/viewer-layout/viewer-layout.tsx
bilalesi: fixed

src/features/scan-config/components/color-by/palette.ts
bilalesi: fixed

src/features/scan-config/components/color-by/viewer-controls-menu.tsx
bilalesi: fixed

src/features/scan-config/components/model-preview/use-morphology-location-selection.ts
bilalesi: fixed

bilalesi:
the Viewer now is reading everything client side, except for the morphology locations which required to be read from the server to have the exact `section_id` that the simulator will use
this PR require `@morphoviewer: 0.32.3` to be merged

src/features/scan-config/components/model-preview/viewer-layout/circuit-loader/circuit-loader.ts
```diff
@@ -93,26 +32,198 @@ export class CircuitLoader {
...
+  private readNodeGeometry(file: H5File, populationName: string) {
+    const group = (name: string) => `nodes/${populationName}/0/${name}`;
+    return {
+      x: this.getNumberArray(file, group('x')),
+      y: this.getNumberArray(file, group('y')),
+      z: this.getNumberArray(file, group('z')),
+      qx: this.getNumberArray(file, group('orientation_x')),
+      qy: this.getNumberArray(file, group('orientation_y')),
+      qz: this.getNumberArray(file, group('orientation_z')),
+      qw: this.getNumberArray(file, group('orientation_w')),
```
pgetta, line 160: Just to note, it turns out that Sonata format can contain either orientations or rotation angles as `rotation_angle_[xyz]axis` properties.

I've seen such circuits in the past but unsure if we have such now.

pgetta (approved):

src/features/scan-config/components/model-preview/viewer-layout/circuit-loader/circuit-loader.ts
bilalesi: good to know, i will keep an eye on this so we expand the function

PR #1868, opened 2026-08-17
fix(data-grid): resolve open review threads from #1832

bilalesi:


src/features/data-grid/host/grid-search.tsx
```diff
@@ -11,20 +11,30 @@ export interface IGridSearchProps {
...
+export function GridSearch({ onSearch, openOnMount = false, className, value }: IGridSearchProps) {
+  const [open, setOpen] = useState(openOnMount || Boolean(value));
+  const [text, setText] = useState(value ?? '');
   const inputRef = useRef<HTMLInputElement>(null);
 
   const commit = useDebouncedCallback((t: string) => onSearch(t), [onSearch], 300);
 
+  // Adopt store-driven changes (hydration on a controller swap, `resetState`). The echo
+  // of this input's own commit is a same-value set that React bails out of.
+  useEffect(() => {
```
pgetta, line 41: Bilal, I'm a bit concerned about this effect racing the 300ms debounce.

Let's imagine the following: I type "m" and pause for a moment, so `commit` flushes and dispatches `SetFreeTextSearch` with "m". That wakes `EntityDataGrid`, `DataGrid` and AG Grid, which is not a cheap render. While it is in flight I type "o", so the local `text` is already "mo". React then commits the store-driven render and runs this effect, which unconditionally does `setText(value)`, so my "o" is gone from the input and from the eventual commit. The same thing happens in the other direction on `controller.resetState()`: the effect clears the input, then the still-pending debounce (we never `commit.cancel()` here) re-dispatches my stale term and the effect puts it back.

Possible solutions:
* Keep a `lastCommittedRef` and adopt `value` only when it differs from what this input last sent - I think that is the cleanest, it turns "the echo of this input's own commit" into an explicit check instead of relying on React bailing out.
* Call `commit.cancel()` inside the effect before `setText`, which at least fixes the reset direction.
* Adopt only on a controller swap (key the effect on the controller identity) rather than on every `value` change.

Or maybe I'm missing something and the window is narrower than I think.

src/ui/segments/workflows/elements/workflow-activity.tsx
```diff
@@ -261,7 +260,8 @@ export function WorkflowActivity() {
     });
   }, [entityType, resolvedActivityType, selectedRow]);
 
-  const shouldShowEmptyState = gridMeta.status === 'loaded' && gridMeta.total === 0;
+  const shouldShowEmptyState =
```
pgetta, line 260: Just want to note a deadlock that this makes easier to hit: `gridMeta` now comes only from the grid, and the grid is exactly what we unmount when `shouldShowEmptyState` is true.

`gridMeta` is reset to `Idle` only inside `updateActivity` and `updateEntityType`. But `entityType` also comes from the URL through `useQueryStates`, so pressing browser Back lands me on a different type without going through either handler: the `key` changes, yet the grid is never rendered, so it never fetches, so `onStateChange` never fires and "You don't have any activities yet" stays on screen until a full reload.

Possible solutions:
* Keep the grid mounted and render the empty state as an overlay on top of it, so the reporter never disappears - looks the most robust to me.
* Reset `gridMeta` from an effect keyed on `[resolvedActivityType, resolvedEntityType]` instead of from the two handlers, so any source of the change is covered. Smaller change if we want to keep the current layout.
* Read total/status from the same React Query key the grid uses instead of routing it through the grid.

Please correct me if the URL cannot change without going through those handlers.

src/features/data-grid/presets/in-memory-grid.tsx
```diff
@@ -665,6 +743,10 @@ export function InMemoryGrid<Row>({
 
   if (!mounted) return <div className={cn('ag-data-grid w-full', className)} />;
 
+  if (isServerMode && serverResult.isError && renderError) {
```
pgetta: This one throws away rows we still have in hand.

`placeholderData: keepPreviousData` is set as the base option above, so on a query key change the previous page stays rendered. Concretely in `ExemplarTraces`: page 1 loads fine, I click page 2, that request fails (`retry: false` is our app default), so `isError` is true while `data` is still the page 1 placeholder. We then replace the whole grid, page 1 rows and pager included, with the error block, and there is no way back to page 1 short of a reload.

Possible solutions:
* Take this path only when there is nothing to show (`serverResult.data === undefined`) and render a smaller inline banner above the grid otherwise - I think that is the best of the three.
* Keep the pager outside the early return so navigation still works.
* Leave it and document `renderError` as first-load-failure only.

src/features/data-grid/presets/simple-grid.tsx
```diff
@@ -403,6 +422,7 @@ export function SimpleGrid<Row>(props: ISimpleGridProps<Row>) {
       autoHeight={props.autoHeight}
       rowSelection={props.rowSelection}
       operators={props.operators}
+      cellRenderers={props.cellRenderers}
```
pgetta, line 430: This is about the `enhanced` check above rather than this line itself: `cellRenderers` is not part of it, so a caller passing only `columns` with renderer keys plus `cellRenderers` lands in `SimpleGridBasic`, which ignores the prop completely. And because `buildSimpleColDefs` short-circuits on `keepsBlankWhenEmpty(c)` (true for any column with a `cellRenderer`), those cells get neither the registry component nor the empty placeholder, so a missing value renders as a fully blank cell.

Probably worth adding `props.cellRenderers` to the `enhanced` predicate, otherwise the prop silently no-ops for exactly the case its doc describes.

src/features/data-grid/presets/in-memory-grid.tsx
```diff
@@ -514,12 +580,24 @@ export function InMemoryGrid<Row>({
...
+          const getValue = c.getValue;
+          colDef.valueGetter = (p) =>
+            p.data && !isDetailRow(p.data) ? (getValue(p.data as Row) ?? null) : null;
+        } else {
+          colDef.field = (c.field ?? c.id) as ColDef<TDisplayRow<Row>>['field'];
+        }
+        // Same wiring as the shared `buildColDefs`: `AgCellHost` resolves the key off
+        // the AG context's registry, so a nested grid reusing a binding's schema
+        // columns renders the same cells as the top-level grid.
+        if (c.cellRenderer) {
```
pgetta, line 631: This sits in the `else` branch only, so the expander host column (the `isHost` case above) still falls back to `getValue`/`field` and drops its renderer key. The shared mapper handles that case by composing `rendererKey` into `AgExpandHostCell` (`renderers/aggrid/col-def-mapper.ts:107-113`), so the two are not quite the same wiring.

`EntityCoreFields.CircuitSubCircuit` declares no `cellRenderer` today, so nothing is broken right now, but as soon as a host column gains one we are back to the plain text degradation this PR is fixing. Would be nice to cover it here too, or at least call out the gap in the comment.

src/features/entities/e-model/detail-view/exemplar-traces.tsx
```diff
@@ -102,6 +104,17 @@ export function ExemplarTraces({ source, variant = ViewVariant.Light }: Props) {
           serverSide={{
             dataSource,
             queryKey: ['emodel-exemplar-traces', virtualLabId, projectId, source.id],
+            renderError: () => (
```
pgetta: From a user's perspective this reads a bit misleading: every failure ends up as "No exemplar traces found", so a 403 because I am not a member of the project, a 500, or a dropped connection all look like "this e-model has no exemplar traces". The `error` argument is ignored, so there is nothing left to tell them apart, and the red error icon on a not-found message is a slightly odd mix too.

`renderListingError` in `browse-entity-grid.tsx` already branches on `ApiError.cause?.code === 'NOT_AUTHORIZED'` for the listing, maybe we can do something similar here and keep this copy for the genuinely empty case.

src/features/data-grid/react/persistence/storage-persistence.ts
```diff
@@ -12,7 +12,7 @@ const SESSION_KEYS = [
   'sort',
   'page',
   'pageSize',
-  'quickFilter',
+  'freeTextSearch',
```
pgetta, line 16: The comment above says the keys are "versioned so a breaking state shape can bump it", and renaming a persisted field is exactly that, but `data-grid:v1:s:` stays as it is.

`load()` returns the raw parsed JSON without filtering to `SESSION_KEYS`, and the controller merges it as `hydrated = { ...hydrated, ...slice }`, so a tab holding a pre-upgrade entry loses its search term and gets a stray `quickFilter` key merged into the live `IGridState`. It goes away on the next `save()` (which does `pick`), so the blast radius is small, but bumping to `v2` here would make it a non-event.

src/features/data-grid/bindings/entitycore/cell-renderers.ts
```diff
@@ -21,6 +22,11 @@ export function buildCellRenderers(definition: TAnyEntityGridDefinition): CellRe
   const registry = new CellRendererRegistry();
   registry.register(LIFECYCLE_STATUS_RENDERER, LifecycleStatusCell);
   registry.register(DESCRIPTION_RENDERER, DescriptionCell);
+  // The shared column catalog emits the preview and contributors keys, so any schema
+  // built from it references them; registering here keeps that from depending on the
+  // author remembering an opt-in line. A definition's own registrations run last and
+  // still win.
+  registerSharedRenderers(registry);
```
pgetta, line 25: I'm curious about the reasoning here. The comment says the shared catalog emits the preview and contributors keys, but from what I can see only `CONTRIBUTORS_RENDERER` comes from the catalog (`columns/catalog.ts:350`) - every `ENTITY_PREVIEW_RENDERER` emitter is a per-schema column.

The key that does match the argument and is still missing is `EM_DATASET_RENDERER`: `emDatasetColumn` is a shared catalog helper (`columns/catalog.ts:806`) and the renderer is registered only inside `em-cell-mesh.ts`. Only that one schema composes it today, so nothing degrades yet, but by the same reasoning it belongs here.

Btw, now that this runs unconditionally the ~15 in-schema `registerSharedRenderers(registry)` calls are dead. In `synthesized-cell-morphology.tsx` and `universal-cell-morphology.ts` that line is the whole body of `registerCellRenderers`, so those two hooks and their imports could go away. Not a call for action for this PR, but leaving them makes it likely the next author copies the pattern.

src/features/data-grid/core/domain/query.ts
```diff
@@ -21,8 +21,8 @@ export interface IGridQuery {
   pageSize: number;
   sort: TSortModel;
   filters: TFilterModel;
-  /** free-text quick filter (a binding maps this to e.g. search / ilike_search) */
-  quickFilter?: string;
+  /** ilike_search (free-text search on name+description) */
```
pgetta: Probably better to keep this one binding-agnostic. `searchMode` defaults to `'plain'` and emits `params.search` rather than `ilike_search` (the serializer test asserts both modes), and the in-memory binding matches the needle across every column via `getValue`, not name plus description. The previous wording covered both, this one describes one binding in one mode inside the transport-agnostic core.

src/features/data-grid/bindings/entitycore/schemas/circuit-simulations.ts
```diff
@@ -92,6 +93,13 @@ export function buildSimulationCampaignDefinition({
     id,
     getRowId: (row) => row.id,
     defaultSort: [{ columnId: 'registrationDate', direction: SortDirection.Desc }],
+    // Checkboxes + bulk actions on Data browse only; workflow surfaces select by picker.
+    selection: {
```
pgetta: We already have this exact spec, comment included, as `memodelDataBrowseSelection` in `schemas/memodel.tsx` - both `memodel.tsx` and `me-model-circuit.tsx` consume it. With the eight campaign definitions here it becomes a third independent copy of the same rule, so the next change to the Data-browse selection policy has to be found in three places.

Would be nice to pull it out into a shared `dataBrowseSelection` constant (it does not really belong in `memodel.tsx` either) and reference it from all three.

src/features/data-grid/host/circuit-grid-body.tsx
```diff
@@ -78,6 +80,9 @@ export function CircuitGridBody(props: IBrowseEntityGridProps) {
   // The nested recursive grid reuses the parent's schema columns (ISimpleColumn
   // extends IColumnModel), keeping expanded rows column-identical to the parent.
   const schemaColumns = definition.schema.columns as unknown as ISimpleColumn<ICircuit>[];
+  // Those columns name their renderers by key, so the nested grid needs the same
+  // registry the top-level grid builds or its cells fall back to plain text.
+  const cellRenderers = useMemo(() => buildCellRenderers(definition), [definition]);
```
pgetta: `EntityDataGrid` already builds a registry from the same `definition` (`browse-entity-grid.tsx:200`), so every circuit listing now constructs two `CellRendererRegistry` instances and the top-level grid and the nested grid resolve keys against different objects.

Probably worth exposing the host's registry through `IEntityDataGridOverrides`, or memoising `buildCellRenderers` per definition, so both sites share one.

src/features/data-grid/react/renderer.ts
```diff
@@ -18,6 +18,22 @@ export type TDetailRenderFn<Row> = (args: {
   state?: IGridState;
 }) => ReactNode;
 
+export const ServerGridStateStatus = {
```
pgetta, line 21: A small thing, but `react/index.ts` picks up the `IServerGridState` type and not this const, so `workflow-activity.tsx` imports the type from `@/features/data-grid/react` and the enum from `@/features/data-grid/react/renderer` - two paths for one concept, and the second reaches past the barrel every other consumer goes through. Adding it to the barrel's value exports would keep that consistent.

`TServerGridStateStatus` just below is also not exported, so a host that wants to type a status parameter has to write `IServerGridState['status']`. Is that intentional?

src/features/data-grid/presets/in-memory-grid.tsx
```diff
@@ -331,7 +362,17 @@ export function InMemoryGrid<Row>({
...
+  // Keyed on the slices the query reads, NOT on `state`: selection/expansion changes
+  // must not give `query` a new identity and re-run the in-memory filter/sort/slice.
+  const { page: statePage, pageSize: statePageSize, sort, filters, freeTextSearch } = state;
+  const query = useMemo<IGridQuery>(
+    () =>
+      buildGridQuery(
+        { page: statePage, pageSize: statePageSize, sort, filters, freeTextSearch },
+        serverParams
+      ),
+    [statePage, statePageSize, sort, filters, freeTextSearch, serverParams]
```
pgetta, line 405: `serverParams` is `serverSide?.params`, and `serverSide` is documented and used as an inline object literal, so `params` gets a fresh identity on every parent render. That gives `query` a new identity every render too, and `clientPage` below is memoised on `query`, so `runInMemoryQuery` still re-runs the full filter/sort/slice on every render, which is what the comment above says this avoids.

Maybe memoise the params at the `SimpleGrid` boundary, or key on a serialised form of them instead of the object.

src/features/data-grid/host/browse-entity-grid.tsx
```diff
@@ -443,7 +480,13 @@ export function EntityDataGrid({
             classNames?.miniView
           )}
         >
-          <MiniDetailView section={section} {...miniViewProps} dataType={dataType} />
+          <MiniDetailView
+            section={section}
+            {...miniViewProps}
+            dataType={dataType}
+            isPrivate={scope === WorkspaceScope.Project}
```
pgetta: Two small things here.

It sits after `{...miniViewProps}`, so a caller-supplied `isPrivate` is silently dropped. The legacy listing does the same, so I assume this is deliberate parity, but `miniViewProps` is typed as a full `Partial<ComponentProps<typeof MiniDetailView>>`, so it is a little surprising.

The other one is that privacy is derived from the listing scope rather than from the selected row, so on a `WorkspaceScope.Combined` listing every row is treated as public and `MiniDetailView` loses its private-only actions on rows I own. Both Combined-scope mounts we have (`related-artifacts/ion-channel-model/e-model.tsx` and `ion-channel-recording.tsx`) pass `requireMiniDetailView: false`, so it is latent, but it will bite the first Combined listing that shows the panel.

src/__tests__/data-grid/presets/in-memory-grid-cell-renderers.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    </span>
+  ));
+}
+
+/**
+ * Nested grids (e.g. the circuit hierarchy's subcircuit levels) reuse the parent's
+ * schema columns, which carry renderer KEYS. Without a registry those cells silently
+ * degrade to the plain `getValue` text, so the pills disappear below the top level.
+ */
+describe('InMemoryGrid — keyed cell renderers', () => {
```
pgetta, line 43: Nice, these three cases (keyed renderer resolved, plain fallback with no registry, inline `renderCell` winning) are exactly the ones that were silently degrading, and the comment explaining why nested grids need it is very helpful.

pgetta (commented):
Nice work, Bilal,

I've left a handful of comments, mostly questions with a few suggestions. A couple of the longer ones we could also address in a follow-up if you prefer.

pgetta (commented):
Nice work, Bilal,

I've left a handful of comments, mostly questions with a few suggestions. A couple of the longer ones we could also address in a follow-up if you prefer.

src/features/data-grid/host/grid-search.tsx
pgetta, line 41: Bilal, I'm a bit concerned about this effect racing the 300ms debounce.

Let's imagine the following: I type "m" and pause for a moment, so `commit` flushes and dispatches `SetFreeTextSearch` with "m". That wakes `EntityDataGrid`, `DataGrid` and AG Grid, which is not a cheap render. While it is in flight I type "o", so the local `text` is already "mo". React then commits the store-driven render and runs this effect, which unconditionally does `setText(value)`, so my "o" is gone from the input and from the eventual commit. The same thing happens in the other direction on `controller.resetState()`: the effect clears the input, then the still-pending debounce (we never `commit.cancel()` here) re-dispatches my stale term and the effect puts it back.

Possible solutions:
* Keep a `lastCommittedRef` and adopt `value` only when it differs from what this input last sent - I think that is the cleanest, it turns "the echo of this input's own commit" into an explicit check instead of relying on React bailing out.
* Call `commit.cancel()` inside the effect before `setText`, which at least fixes the reset direction.
* Adopt only on a controller swap (key the effect on the controller identity) rather than on every `value` change.

Or maybe I'm missing something and the window is narrower than I think.

src/ui/segments/workflows/elements/workflow-activity.tsx
pgetta, line 260: Just want to note a deadlock that this makes easier to hit: `gridMeta` now comes only from the grid, and the grid is exactly what we unmount when `shouldShowEmptyState` is true.

`gridMeta` is reset to `Idle` only inside `updateActivity` and `updateEntityType`. But `entityType` also comes from the URL through `useQueryStates`, so pressing browser Back lands me on a different type without going through either handler: the `key` changes, yet the grid is never rendered, so it never fetches, so `onStateChange` never fires and "You don't have any activities yet" stays on screen until a full reload.

Possible solutions:
* Keep the grid mounted and render the empty state as an overlay on top of it, so the reporter never disappears - looks the most robust to me.
* Reset `gridMeta` from an effect keyed on `[resolvedActivityType, resolvedEntityType]` instead of from the two handlers, so any source of the change is covered. Smaller change if we want to keep the current layout.
* Read total/status from the same React Query key the grid uses instead of routing it through the grid.

Please correct me if the URL cannot change without going through those handlers.

src/features/data-grid/presets/in-memory-grid.tsx
pgetta: This one throws away rows we still have in hand.

`placeholderData: keepPreviousData` is set as the base option above, so on a query key change the previous page stays rendered. Concretely in `ExemplarTraces`: page 1 loads fine, I click page 2, that request fails (`retry: false` is our app default), so `isError` is true while `data` is still the page 1 placeholder. We then replace the whole grid, page 1 rows and pager included, with the error block, and there is no way back to page 1 short of a reload.

Possible solutions:
* Take this path only when there is nothing to show (`serverResult.data === undefined`) and render a smaller inline banner above the grid otherwise - I think that is the best of the three.
* Keep the pager outside the early return so navigation still works.
* Leave it and document `renderError` as first-load-failure only.

src/features/data-grid/presets/simple-grid.tsx
pgetta, line 430: This is about the `enhanced` check above rather than this line itself: `cellRenderers` is not part of it, so a caller passing only `columns` with renderer keys plus `cellRenderers` lands in `SimpleGridBasic`, which ignores the prop completely. And because `buildSimpleColDefs` short-circuits on `keepsBlankWhenEmpty(c)` (true for any column with a `cellRenderer`), those cells get neither the registry component nor the empty placeholder, so a missing value renders as a fully blank cell.

Probably worth adding `props.cellRenderers` to the `enhanced` predicate, otherwise the prop silently no-ops for exactly the case its doc describes.

src/features/data-grid/presets/in-memory-grid.tsx
pgetta, line 631: This sits in the `else` branch only, so the expander host column (the `isHost` case above) still falls back to `getValue`/`field` and drops its renderer key. The shared mapper handles that case by composing `rendererKey` into `AgExpandHostCell` (`renderers/aggrid/col-def-mapper.ts:107-113`), so the two are not quite the same wiring.

`EntityCoreFields.CircuitSubCircuit` declares no `cellRenderer` today, so nothing is broken right now, but as soon as a host column gains one we are back to the plain text degradation this PR is fixing. Would be nice to cover it here too, or at least call out the gap in the comment.

src/features/entities/e-model/detail-view/exemplar-traces.tsx
pgetta: From a user's perspective this reads a bit misleading: every failure ends up as "No exemplar traces found", so a 403 because I am not a member of the project, a 500, or a dropped connection all look like "this e-model has no exemplar traces". The `error` argument is ignored, so there is nothing left to tell them apart, and the red error icon on a not-found message is a slightly odd mix too.

`renderListingError` in `browse-entity-grid.tsx` already branches on `ApiError.cause?.code === 'NOT_AUTHORIZED'` for the listing, maybe we can do something similar here and keep this copy for the genuinely empty case.

src/features/data-grid/react/persistence/storage-persistence.ts
pgetta, line 16: The comment above says the keys are "versioned so a breaking state shape can bump it", and renaming a persisted field is exactly that, but `data-grid:v1:s:` stays as it is.

`load()` returns the raw parsed JSON without filtering to `SESSION_KEYS`, and the controller merges it as `hydrated = { ...hydrated, ...slice }`, so a tab holding a pre-upgrade entry loses its search term and gets a stray `quickFilter` key merged into the live `IGridState`. It goes away on the next `save()` (which does `pick`), so the blast radius is small, but bumping to `v2` here would make it a non-event.

src/features/data-grid/bindings/entitycore/cell-renderers.ts
pgetta, line 25: I'm curious about the reasoning here. The comment says the shared catalog emits the preview and contributors keys, but from what I can see only `CONTRIBUTORS_RENDERER` comes from the catalog (`columns/catalog.ts:350`) - every `ENTITY_PREVIEW_RENDERER` emitter is a per-schema column.

The key that does match the argument and is still missing is `EM_DATASET_RENDERER`: `emDatasetColumn` is a shared catalog helper (`columns/catalog.ts:806`) and the renderer is registered only inside `em-cell-mesh.ts`. Only that one schema composes it today, so nothing degrades yet, but by the same reasoning it belongs here.

Btw, now that this runs unconditionally the ~15 in-schema `registerSharedRenderers(registry)` calls are dead. In `synthesized-cell-morphology.tsx` and `universal-cell-morphology.ts` that line is the whole body of `registerCellRenderers`, so those two hooks and their imports could go away. Not a call for action for this PR, but leaving them makes it likely the next author copies the pattern.

src/features/data-grid/core/domain/query.ts
pgetta: Probably better to keep this one binding-agnostic. `searchMode` defaults to `'plain'` and emits `params.search` rather than `ilike_search` (the serializer test asserts both modes), and the in-memory binding matches the needle across every column via `getValue`, not name plus description. The previous wording covered both, this one describes one binding in one mode inside the transport-agnostic core.

src/features/data-grid/bindings/entitycore/schemas/circuit-simulations.ts
pgetta: We already have this exact spec, comment included, as `memodelDataBrowseSelection` in `schemas/memodel.tsx` - both `memodel.tsx` and `me-model-circuit.tsx` consume it. With the eight campaign definitions here it becomes a third independent copy of the same rule, so the next change to the Data-browse selection policy has to be found in three places.

Would be nice to pull it out into a shared `dataBrowseSelection` constant (it does not really belong in `memodel.tsx` either) and reference it from all three.

src/features/data-grid/host/circuit-grid-body.tsx
pgetta: `EntityDataGrid` already builds a registry from the same `definition` (`browse-entity-grid.tsx:200`), so every circuit listing now constructs two `CellRendererRegistry` instances and the top-level grid and the nested grid resolve keys against different objects.

Probably worth exposing the host's registry through `IEntityDataGridOverrides`, or memoising `buildCellRenderers` per definition, so both sites share one.

src/features/data-grid/react/renderer.ts
pgetta, line 21: A small thing, but `react/index.ts` picks up the `IServerGridState` type and not this const, so `workflow-activity.tsx` imports the type from `@/features/data-grid/react` and the enum from `@/features/data-grid/react/renderer` - two paths for one concept, and the second reaches past the barrel every other consumer goes through. Adding it to the barrel's value exports would keep that consistent.

`TServerGridStateStatus` just below is also not exported, so a host that wants to type a status parameter has to write `IServerGridState['status']`. Is that intentional?

src/features/data-grid/presets/in-memory-grid.tsx
pgetta, line 405: `serverParams` is `serverSide?.params`, and `serverSide` is documented and used as an inline object literal, so `params` gets a fresh identity on every parent render. That gives `query` a new identity every render too, and `clientPage` below is memoised on `query`, so `runInMemoryQuery` still re-runs the full filter/sort/slice on every render, which is what the comment above says this avoids.

Maybe memoise the params at the `SimpleGrid` boundary, or key on a serialised form of them instead of the object.

src/features/data-grid/host/browse-entity-grid.tsx
pgetta: Two small things here.

It sits after `{...miniViewProps}`, so a caller-supplied `isPrivate` is silently dropped. The legacy listing does the same, so I assume this is deliberate parity, but `miniViewProps` is typed as a full `Partial<ComponentProps<typeof MiniDetailView>>`, so it is a little surprising.

The other one is that privacy is derived from the listing scope rather than from the selected row, so on a `WorkspaceScope.Combined` listing every row is treated as public and `MiniDetailView` loses its private-only actions on rows I own. Both Combined-scope mounts we have (`related-artifacts/ion-channel-model/e-model.tsx` and `ion-channel-recording.tsx`) pass `requireMiniDetailView: false`, so it is latent, but it will bite the first Combined listing that shows the panel.

src/__tests__/data-grid/presets/in-memory-grid-cell-renderers.test.tsx
pgetta, line 43: Nice, these three cases (keyed renderer resolved, plain fallback with no registry, inline `renderCell` winning) are exactly the ones that were silently degrading, and the comment explaining why nested grids need it is very helpful.

src/features/data-grid/host/grid-search.tsx
bilalesi: fixed

src/features/data-grid/presets/in-memory-grid.tsx
bilalesi: fixed

src/features/data-grid/presets/in-memory-grid.tsx
bilalesi: covered. Rather than copy the wiring, I pulled the resolve-and-render block into a shared renderKeyedCell used by AgCellHost, AgExpandHostCell and the in-memory host — it was three copies that had already drifted apart on the miss path. The in-memory host now resolves off context.cellRenderers like the others.

src/features/data-grid/presets/simple-grid.tsx
bilalesi: Fixed with your first option

src/features/entities/e-model/detail-view/exemplar-traces.tsx
bilalesi: Fixed. It now branches like renderListingError does

src/features/data-grid/react/persistence/storage-persistence.ts
bilalesi: bumped the session slice to data-grid:v2:s: and updated the two tests that pinned the old key.

src/features/data-grid/bindings/entitycore/cell-renderers.ts
bilalesi: good catch, registered emDataset

src/features/data-grid/core/domain/query.ts
bilalesi: reverted to neutral wording

src/features/data-grid/bindings/entitycore/schemas/circuit-simulations.ts
bilalesi: pulled it into schemas/selection-policies.ts

src/features/data-grid/host/circuit-grid-body.tsx
bilalesi: added getCellRenderers(definition)

src/features/data-grid/react/renderer.ts
bilalesi: both exported from the barrel now

src/features/data-grid/presets/in-memory-grid.tsx
bilalesi: I skip the in-memory pass entirely in server mode

src/features/data-grid/host/browse-entity-grid.tsx
bilalesi: both fixed

src/__tests__/data-grid/presets/in-memory-grid-cell-renderers.test.tsx
bilalesi: not an issue

pgetta (approved):

PR #1872, opened 2026-08-18
feat(44): add `build > circuit synaptic physiology` workflow

bilalesi:
configure and launch synaptic parameterization
for circuits of scale ≤ small microcircuit via the obi-one generate-grid and task launch endpoints.

related: https://github.com/openbraininstitute/prod-build-circuit/issues/44

### DIFF
--- a/src/__tests__/entity-routing/url-builder.test.ts
+++ b/src/__tests__/entity-routing/url-builder.test.ts
@@ -56,6 +56,12 @@ describe('resolveConcreteEntityPathParam', () => {
         resolveConcreteEntityPathParam(ExtendedEntitiesTypeDict.EmSynapseMappingCampaign)
       ).toBe('circuit');
     });
+
+    it('routes circuit synaptic physiology campaigns to the circuit listing (no browse page of their own)', () => {
+      expect(
+        resolveConcreteEntityPathParam(ExtendedEntitiesTypeDict.CircuitSynapticPhysiologyCampaign)
+      ).toBe('circuit');
+    });
   });
 
   describe('other entities map to themselves (kebab-cased)', () => {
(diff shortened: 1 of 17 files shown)


src/features/scan-config/components/ui-elements/discrete-probabilities.tsx
```diff
@@ -0,0 +1,99 @@
...
+    probability: probabilities[i],
+  }));
+
+  const total = probabilities.reduce((sum, p) => sum + (Number.isFinite(p) ? p : 0), 0);
+
+  const update = (i: number, next: Partial<{ value: number; probability: number }>) => {
+    const nextValues = [...values];
+    const nextProbabilities = [...probabilities];
+    if (next.value !== undefined) nextValues[i] = next.value;
+    if (next.probability !== undefined) nextProbabilities[i] = next.probability;
```
pgetta: It seems editing a row beyond the end of the shorter array shifts the value to another row. Let's imagine `values=[1, 2, 3, 4]` and `probabilities=[0.5, 0.5]` (the case covered by the mismatched pair test): entering `0.3` into "Probability 4" produces a sparse `[0.5, 0.5, <empty>, 0.3]`. The parent then reads it back through `asNumbers` in `ui-elements/index.tsx`, which drops the hole, so on the next render the `0.3` shows up in row 3, row 4 is empty again and the wrong pairing gets submitted. The same applies to a value typed past the end of `values`.

Probably it's best to pad both arrays to `rowCount` (e.g. with `0`) before assigning, so the positions stay aligned. Or maybe I'm missing something here.

src/features/scan-config/components/ui-elements/discrete-probabilities.tsx
```diff
@@ -0,0 +1,99 @@
...
+        <div className="text-red-500 text-xs">At least one probability must be above zero.</div>
+      )}
+
+      {!disabled && (
+        <button
+          type="button"
+          className="mt-2 flex min-h-[40px] min-w-[150px] items-center justify-between float-right rounded-full border border-gray-200 px-3 py-2 font-bold text-primary-8"
+          onClick={() => {
+            const last = values.at(-1);
+            onChange([...values, (last ?? 0) + 1], [...probabilities, 0]);
```
pgetta: Same issue here: with arrays of different length each one is appended at its own end, so for `[1, 2, 3, 4]` / `[0.5, 0.5]` the new `0` probability lands in row 3 rather than row 5. Padding both arrays to `rowCount` first would fix both places, and it would be nice to cover the mismatched case with tests for editing and adding as well.

pgetta (commented):
Nice work, Bilal, I've left a couple of small comments.

pgetta (approved):

src/features/scan-config/components/ui-elements/discrete-probabilities.tsx
bilalesi: fixed

pgetta (dismissed):

pgetta (approved):

bilalesi:
thank you @pgetta

PR #1873, opened 2026-08-19
fix(offline-auth): only reuse a consent grant within its own session

james-isbister:
An issue I encounted when launching a simulation through the launch service. Had to run "localStorage.removeItem('auth.offline-token-consent.v1.state')" to fix it

This is Claude's proposed solution and description, I don't understand it, but it would be nice to solve the issue so that launching doesn't fail - maybe this is a useful starting point. I think I've had the same issue several times before

## The bug

Launching a microcircuit simulation failed with the generic toast **"We ran into a problem submitting your simulation(s). Please try again later."** Retrying never helped.

Traced on staging (2026-08-19 06:53:24 UTC):

```
frontend  POST /api/obi-one/declared/task/launch                    -> 500
obi-one   POST /api/launch-system/job                               -> 500
launch    POST /api/auth-manager/v1/offline-token-id                -> 404
          {"error":"No token found for the session/type requested",
           "code":"token_not_found"}
```

By the time it failed, obi-one had already resolved the circuit, estimated the cost, **reserved the credits** and created the `SimulationExecution`. It then unwound all of it (reservation cancelled, activity patched) and returned a 500.

## Why

`results.tsx` gates the launch behind `useEnsureOfflineTokenConsent({ useCache: true })`. That gate short-circuited on a cached `granted` state whenever `isOfflineTokenConsentStateFresh()` passed — and that only checks `updatedAt` is under 24h.

auth-manager stores the offline token against the Keycloak session's `session_state_id`. `store.ts` was already recording `sessionStateId` alongside the grant, but nothing ever read it back — `grep` for any comparison of it returns nothing. So a grant made in yesterday's session kept satisfying the client gate for 24 hours while the token the launch system asks for did not exist for today's session.

Confirmed session-specific rather than an outage: over the same 24h window `offline-token-id` returned `{200: 1, 404: 3}` — a different user succeeded while this one failed on all three attempts.

## The change

- `isOfflineTokenConsentGrantedForSession()` in `store.ts` — fresh **and** granted **and** belonging to the session we are in. An unknown session id on either side counts as a miss, so the consent flow runs again rather than being skipped on a guess.
- `ensure()` reads the cached grant *after* auth-manager has said which session we are in, instead of short-circuiting ahead of that request. `useCache` now saves the trip through the consent page rather than the request itself; its doc comment says so.
- A prefetched consent response is only used inside `PREFETCH_TTL_MS`, since it now carries the session id the grant is checked against — an old one would answer for a session we may no longer be in.
- Unit tests covering same-session, cross-session, unknown-session, expired, denied, and the storage round trip.

## Not covered here

The failure is unrecoverable but reported as retryable — obi-one turns `token_not_found` into a generic `INTERNAL_ERROR` and the UI says "try again later", which is the opposite of what helps. Surfacing it as "your session needs to re-authorise" is a separate change, across obi-one and the frontend.

## Testing

- `npx vitest run src/features/offline-auth-management/store.test.ts` — 9 passed
- Full `vitest run` — 176 files, 2027 passed, 3 skipped
- `biome check` on the changed files — clean (one pre-existing `catch (err: any)` info)
- `tsc --noEmit` does not run on this repo at all right now: TypeScript is pinned to 7.0.2 and `tsconfig.json` still sets `baseUrl`, which 7.x removed. Fails identically on unmodified `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pgetta (dismissed):

src/__tests__/auth-manager/offline-token-consent-store.test.ts
bilalesi, line 1: can we move the test under `src/__tests__/auth-manager`

james-isbister: Moved to `src/__tests__/auth-manager/offline-token-consent-store.test.ts` in f74b1edad, with the relative imports switched to the `@/` alias the other suites there use. Renamed off bare `store.test.ts` so it names the module it covers once other auth-manager tests land in that directory — happy to go back to `store.test.ts` if you prefer.

james-isbister: Lol, I didn't write the comment as you may have guessed...

I'll let you guys merge it when it's ready

pgetta (approved):
Looks good!

PR #1874, opened 2026-08-19
fix(212): restore first-load checkbox selection on a new campaign

bilalesi:
related to https://github.com/openbraininstitute/prod-build-emodel/issues/212


pgetta:
It seems there is something broken with the auto-selection: when I create a (sim) campaign and switch to simulations - all of them are selected, after a click on **one** of them - selections dissapear for **all**. Is that expected?

P.S. Tried that in the preview deployment.

bilalesi:
https://github.com/user-attachments/assets/fdc63da8-0604-4008-bdb5-595d2a53d4a8

if this is the flow you described, then it works for me, can you check again please @pgetta ?

pgetta:
I meant selection via the checkboxes:

https://github.com/user-attachments/assets/6c024d34-1b08-40e6-a765-c291172d93af

bilalesi:
Good catch @pgetta, i fixed it now

pgetta (approved):

PR #1875, opened 2026-08-19
feat(circuit-viz): large circuits onto the shared nodes worker

pgetta:
Region-scale circuits were downloading their `nodes.h5` **twice**: once streamed into the worker's filesystem for the nodes table and colour-by, and once into main-thread memory via `ServiceSonataLoader`, which then materialised every column as a JS array through `dataset.json_value`. `CircuitPreview` already opens a worker session for colour-by, so both copies were always in flight together.

This removes the second copy, and `src/services/sonata/` (471 lines) with it.

Placement now comes from `useNodeGeometry`, which owns the config, the worker session and the one `getGeometry` call. The session is keyed by circuit, asset and population, so a viewer and the nodes table looking at the same population download the file once between them.

Reading the `morphology` column is opt-in. Somas-only rendering does not need names: morphoviewer's `morphologyId` reaches no renderer, only a change-detection diff that compares positions in the same loop. Leaving it out skips the `@library` decode inside the worker as well as one JS string per node across the wire.

### Also here

- A `circuit_config.json` that parses but declares no node populations now says so. The worker session is gated on having one, so without this the viewer sat on its spinner with no data and no error.
- `fetchToFS` and `writeToFS` cap assets at 2 GiB. h5wasm is a 32-bit build, so the file and every column read out of it share one 4 GiB address space; past a point the tab dies rather than failing. Checked up front from `Content-Length`, and again as bytes arrive for responses that declare none.
- `resolveCircuitAssetPath` composes manifest expansion with path normalisation. Expansion routinely yields `$BASE/../networks/x.h5`, and these strings leave as `asset_path` query parameters, where a `..` is a segment the server matches literally. Kept separate from `resolveManifestPath` because the circuit download panel looks paths up as keys in a directory tree, where collapsing a segment changes the key.

### Behaviour change

Large circuits draw the population chosen in the selector rather than the first one `node_sets.json` happened to name. Counts are unchanged — node_sets only ever picked the population, it never filtered within it.

### Relationship to #1863

Built on top of it. This PR adopts the `MorphoViewerCircuitMultipleNeuronsSomaOnly` rename and leaves the `circuitNodes` / `circuitSynapses` query keys alone; the follow-up removes them. `src/features/circuit-nodes/**` is untouched by #1863, so the overlap here is small.

A second PR — https://github.com/openbraininstitute/core-web-app/pull/1876 — stacks on this one and does the same for single-scale circuits, deleting `CircuitLoader`.

### Verification

- 2082 unit tests pass; `pnpm build` clean; Biome clean on changed files.
- `tsc` against a patched config (the repo's `baseUrl` breaks `pnpm typecheck` outright) at main's 267-error baseline, with **zero** errors in touched files.

**Not yet done, and needed before merge:** open a microcircuit and a region-scale circuit — confirm in DevTools Network that `nodes.h5` is fetched once rather than twice, watch peak JS heap, confirm the population selector moves the 3D view, and confirm colour-by still repaints through the `colorOnly` path.


pgetta:
This is a preparation for implementing https://github.com/openbraininstitute/prod-circuit-simulation/issues/159.

src/features/circuit-nodes/worker/nodes-h5.ts
```diff
@@ -364,6 +375,62 @@ export class NodesSession {
...
+
+    const morphologies =
+      withMorphologies && this.columnIndex.has(MORPHOLOGY_COLUMN)
+        ? asStrings(this.getColumnValues(MORPHOLOGY_COLUMN).values)
+        : null;
+
+    return {
+      count: this.rowCount,
+      positions,
+      orientations: this.packColumns(ORIENTATION_COLUMNS),
```
bilalesi: Orientations are always read and always sent, but nothing in this PR uses them. The somas-only viewer draws points.

So every load reads 4 extra columns and packs a `count * 4` `Float64Array` (128 MB on a 4M-node circuit) that is transferred and then dropped.

This is the exact cost you avoid for `morphology`, so please treat it the same way:

```ts
getGeometry({ withMorphologies = false, withOrientations = false }: NodeGeometryOptions = {}) {
  ...
  orientations: withOrientations ? this.packColumns(ORIENTATION_COLUMNS) : null,
}
```

src/features/circuit-nodes/worker/nodes-h5.ts
```diff
@@ -364,6 +375,62 @@ export class NodesSession {
...
+    // only to discard them evicts three entries from a cache the nodes table
+    // is sharing.
+    if (!names.every((name) => this.columnIndex.has(name))) return null;
+
+    const columns: LoadedColumn[] = names.map((name) => this.loadColumn(name));
+
+    const stride = names.length;
+    const out = new Float64Array(this.rowCount * stride);
+    for (let c = 0; c < stride; c++) {
+      const column = columns[c] as Float32Array | Float64Array;
```
bilalesi: This cast is not safe. `LoadedColumn` can also be `string[]` or `Uint32Array`.

- If `x` is a string column, every position becomes `NaN`.
- If `x` is categorical, positions become category indices.

Both are silent. The viewer just draws in the wrong place.

Check the kind instead of casting, on line 421:

```ts
if (!names.every((n) => this.columnIndex.get(n)?.kind === 'numeric')) return null;
```

src/features/circuit-nodes/geometry-utils.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/**
+ * Read one node's placement, or null where the index is not in the population.
+ *
+ * Null rather than a default placement: a caller asking about a node that isn't
+ * drawn — an edge file's `target_node_id` pointing outside the drawn population,
+ * say — wants to know that, not to be handed the origin.
+ */
+export function placementAt(geometry: NodeGeometry, index: number): NodePlacement | null {
```
bilalesi: `placementAt`, `NodePlacement` and `IDENTITY_QUATERNION` have no caller. Only `positionAt` is imported.

Please move them to #1876, where something actually uses them. Code with no caller cannot be reviewed for correctness.

src/features/circuit-nodes/types.ts
```diff
@@ -116,6 +153,15 @@ export type ParsedCircuitConfig = {
...
   circuitAssetId: string;
+  /**
+   * The `circuit_config.json` as served, manifest variables unresolved.
+   *
+   * Kept alongside the parsed populations because morphology lookup needs
+   * fields this shape does not carry (`components`, `alternate_morphologies`),
+   * and re-downloading the config to read them would be a second request for
+   * bytes already in hand.
+   */
+  raw: ICircuitSonataConfiguration;
```
bilalesi: `raw` is added but never read in this PR. The comment explains it is for morphology lookup, which is #1876.

Same point as `placementAt`: please add it in the PR that uses it.

src/features/circuit-nodes/types.ts
```diff
@@ -98,6 +100,41 @@ export type OpenResponse = {
...
+  /**
+   * Read the `morphology` column too. Off by default, because it is the one
+   * part of {@link NodeGeometry} that cannot cross the worker boundary as a
+   * transferable: it decodes to one JS string per node, which at region scale
+   * is the single largest allocation in the payload.
+   *
+   * Only callers that resolve a morphology *file* from the name need it; the
+   * somas-only viewer does not.
+   */
+  withMorphologies?: boolean;
```
bilalesi, line 133: No caller passes `withMorphologies: true` outside the test. The option and its test are fine to keep if #1876 lands right after, but flagging it as the third piece of API here with no consumer.

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
```diff
@@ -130,46 +156,42 @@ export function LargeCircuitPreview({
...
-          />
-        ))}
+      {!geometry && !error && <VisualizationLoadingIndicator />}
+      {error && (
+        <div className={styles.error}>
+          <h2>
+            Unable to load circuit <strong>{circuit.name}</strong>!
+          </h2>
+          <p>{error.message}</p>
+          <ErrorDetail cause={error.cause} />
```
bilalesi: `ErrorDetail` is dead now.

It renders the `loader.steps` array that `ServiceSonataLoader` attached as `error.cause`. None of the new errors set a `cause`, so `parseCause` always returns null and this always renders nothing.

Please delete `ErrorDetail` and `parseCause` along with this line.

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
```diff
@@ -28,6 +31,8 @@ import styles from './large-circuit-preview.module.css';
 export interface LargeCircuitPreviewProps {
   className?: string;
   circuit: ICircuit;
+  /** Host-owned, so `colorsByNode` is indexed against the cells it colours. */
+  population?: NodePopulation;
```
bilalesi: Make this required.

If a caller forgets it, the user sees "This circuit's circuit_config.json declares no node populations" — which blames the data for a caller bug. The host always has a population, so there is no reason for it to be optional.

src/features/circuit-nodes/hooks/use-node-geometry.ts
```diff
@@ -0,0 +1,90 @@
...
+    circuitAssetId: config?.circuitAssetId ?? '',
+    population,
+  });
+
+  const [geometry, setGeometry] = useState<NodeGeometry | null>(null);
+  const [geometryError, setGeometryError] = useState<Error | null>(null);
+
+  useEffect(() => {
+    if (status !== 'ready') {
+      setGeometry(null);
```
bilalesi, line 49: `setGeometryError(null)` only runs on the `status === 'ready'` path, so the old error survives here.

Pick a population that fails, then pick a good one: the old error stays on screen for the whole download instead of a spinner.

```ts
if (status !== 'ready') {
  setGeometry(null);
  setGeometryError(null);
  return;
}
```

src/utils/circuit-manifest.ts
```diff
@@ -51,6 +51,21 @@ export function resolveManifestPath(path: string, manifest?: Record<string, stri
...
+ *
+ * Expansion routinely produces `$BASE/../networks/x.h5`, and these strings leave
+ * as `asset_path` query parameters, where a `..` is a segment the server matches
+ * literally rather than a step up.
+ *
+ * Composed rather than folded into `resolveManifestPath` because the circuit
+ * download panel wants the expansion unnormalised — it looks paths up as keys in
+ * a directory tree, where collapsing a segment changes the key.
+ */
+export function resolveCircuitAssetPath(path: string, manifest?: Record<string, string>): string {
```
bilalesi, line 65: `circuit-loader/circuit-config.ts:41` already does `normalizeRelativePath(resolveManifestPath(...))` inline. Good that it now has a name — please use it there too, so there is one helper and not two spellings of the same thing.

Also: `circuit-manifest.nodetest.ts` covers the other two functions. Please add a case for this one.

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
```diff
@@ -79,41 +92,54 @@ export function LargeCircuitPreview({
...
-        color: colorsByNode?.[index],
-      };
-      return cell;
-    });
-  }, [nodes, colorsByNode]);
+  const cellInfos = React.useMemo(
+    () =>
+      positions.map(
+        (position, i): CellInfo => ({
+          morphologyId: String(i),
```
bilalesi: Optional: this allocates one string per node.

I checked morphoviewer — `morphologyId` is only read in `sameGeometry`, compared next to the position in the same loop. So one shared constant behaves identically and allocates nothing:

```ts
const CELL_ID = ''; // then morphologyId: CELL_ID
```

Right now the PR avoids one JS string per node in the worker, then creates one per node here.

src/features/circuit-nodes/hooks/use-nodes-worker.ts
```diff
@@ -102,6 +106,13 @@ export function useNodesWorker({ enabled, circuitId, circuitAssetId, population
     };
   }, [key]);
 
+  const getGeometry = useMemo(() => {
```
bilalesi: Nit: `useMemo` returning a function should be `useCallback`, matching `retry` right below.

src/utils/h5/fs.ts
```diff
@@ -115,6 +147,9 @@ function streamToFS({
         if (done) break;
         FS.write(stream, value, 0, value.byteLength, offset);
         offset += value.byteLength;
+        // A response that declared no length gets checked as it arrives; one
+        // that did was already rejected before the first read.
+        if (offset > MAX_ASSET_BYTES) throw tooLargeError(MAX_ASSET_BYTES, false);
```
bilalesi: Nit: the check runs after `FS.write`, so one chunk always goes over the limit. Check before writing.

Also: no test covers the new cap. `src/__tests__/utils/h5/fs.test.ts` already exists — two cases would do it: a `Content-Length` over the limit, and `writeToFS` with non-HDF5 bytes.

src/utils/h5/fs.ts
```diff
@@ -25,6 +25,38 @@ const PROGRESS_BYTE_STEP = 2 * 1024 * 1024;
...
+ * and MEMFS holds the file as one contiguous allocation. Past a certain size
+ * the download does not fail so much as the tab dies, several layers from
+ * anything that names a cause.
+ *
+ * A ceiling here turns that into a sentence. It is deliberately not the
+ * theoretical maximum: leaving room for h5wasm's own working set is the point.
+ */
+const MAX_ASSET_BYTES = 2 * 1024 * 1024 * 1024;
+
+function formatSize(bytes: number): string {
```
bilalesi, line 48: Nit: this divides by 1024³ but prints `GB`. That is GiB.

src/utils/h5/fs.ts
```diff
@@ -25,6 +25,38 @@ const PROGRESS_BYTE_STEP = 2 * 1024 * 1024;
...
+  return `${(bytes / 1024 / 1024 / 1024).toFixed(1)} GB`;
+}
+
+function tooLargeError(bytes: number, exact: boolean): AssetFetchError {
```
bilalesi, line 57: Nit: `tooLargeError(MAX_ASSET_BYTES, false)` reads "over 2.0 GB; limit 2.0 GB", which tells the user nothing. Pass the bytes received so far for the first number.

bilalesi (changes requested):
Good direction. One worker session instead of two downloads, and 471 lines of duplicate loading gone. I checked the `morphologyId` claim against morphoviewer and it is correct: `sameGeometry` is the only reader.

src/features/circuit-nodes/worker/nodes-h5.ts
pgetta: Done - `withOrientations`, off by default, same shape as morphologies. The two orientation tests ask for it explicitly now, and there is a new one asserting the default leaves the four columns unread.

src/features/circuit-nodes/worker/nodes-h5.ts
pgetta: Good catch, that one was real. It guards on the handle kind before the load now, so the reason for checking up front (not evicting three cache entries to then discard them) still holds.

Added a case with a string `x` column: it used to hand back a population drawn at NaN, now it says the population has no x/y/z columns.

src/features/circuit-nodes/geometry-utils.ts
pgetta: Moved to #1876, along with `NodePlacement` and the identity quaternion. `positionAt` stays, since the somas path is its caller.

src/features/circuit-nodes/types.ts
pgetta: Moved. Same for the `config` field `useNodeGeometry` hands back - you did not flag that one, but it was there for `raw` and nothing in this PR reads it either.

src/features/circuit-nodes/types.ts
pgetta: This one I would rather keep here, if you are fine with it.

The line I drew after your other two comments is that untested consumer plumbing moves down and tested worker surface stays. `placementAt` and `raw` had no caller and no test; both branches of this flag are covered in `nodes-geometry.test.ts` - the column left unread by default, and the `@library` decode when asked. Moving the flag means moving those two cases into a PR that does not own the worker.

`withOrientations` sits in the same position after your other comment, so at least they are symmetric.

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
pgetta: Right, gone - both functions plus the `isType` import that only `parseCause` used.

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
pgetta: Taking the point, but not quite the type.

The host's value really is undefined for a while: `resolvePopulation` has nothing to resolve until `circuit_config.json` lands, so `NodePopulation` would be a lie the caller ends up casting around.

What catches the bug you are describing is a required key with a nullable value - `population: NodePopulation | undefined`. A caller can no longer leave it out, and the loading state stays honest. Let me know if you would rather have it the other way.

src/features/circuit-nodes/hooks/use-node-geometry.ts
pgetta: Fixed. Reproduced it exactly as you wrote it - failing population, then a good one, and the old message sits there for the whole download.

src/utils/circuit-manifest.ts
pgetta: First half done, `circuit-config.ts:41` calls it now.

On the test, a case in `circuit-manifest.nodetest.ts` would not have run: vitest excludes `**/*.nodetest.*` ([vitest.config.ts#L24](https://github.com/openbraininstitute/core-web-app/blob/cdcdd515326d0f9c26c247696303a66cc02ddc90/vitest.config.ts#L24)), and `pnpm test:node` fails outright on this repo (`mock.module is not a function`, and Node's runner cannot load `.tsx` at all). So I ported the file to `src/__tests__/utils/circuit-manifest.test.ts` and put three cases for the new helper there. Side effect is that the 13 cases it already had are back in CI.

Btw, the rest of the `*.nodetest.*` files look unreachable for the same reason. Probably worth retiring them separately.

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
pgetta: Done, one shared constant. The position comparison in `sameGeometry` is doing the discriminating either way.

src/features/circuit-nodes/hooks/use-nodes-worker.ts
pgetta: Done, and `getColumn` right above it too, so the file settles on one idiom.

src/utils/h5/fs.ts
pgetta: Moved above `FS.write`, so the chunk that would cross the limit is never stored.

Two cases added: a `Content-Length` over the cap rejected before a byte is written, and `writeToFS` with non-HDF5 bytes (plus a buffer too short to hold the signature).

The streaming overshoot I left uncovered - it needs a body over 2 GiB, and making the limit injectable only to reach it seemed like more surface than the case is worth. Happy to add it if you see it differently.

src/utils/h5/fs.ts
pgetta: Fixed, GiB.

src/utils/h5/fs.ts
pgetta: Fixed - it passes the bytes received so far, so it reads "over 2.1 GiB; limit 2.0 GiB".

pgetta:
Thanks Bilal, all of it addressed and pushed as an amend on the same commit.

Two I did differently, both explained in the threads: the `population` type, and the manifest test - the `*.nodetest.*` file it would have gone into is not run by anything in CI.

bilalesi (approved):

pgetta:
Thank you.

PR #1876, opened 2026-08-19
refactor(circuit-viz): fold the single-scale reader into the shared worker

pgetta:
> Stacked on #1875 — review that one first. This PR's diff is against it, not `main`.

#1875 put large circuits on the shared SONATA nodes worker. Small and single-scale circuits were still read by `CircuitLoader`, which opened the same `nodes.h5` a second time on the main thread and materialised every column as a JS array. This puts them on the worker too and deletes the class.

### What replaces what

| Gone | Replacement |
|---|---|
| `CircuitLoader.getNodes` | `useNodeGeometry` — same columns, read once for the viewer, the nodes table and colour-by together, as typed arrays rather than one JS object per node |
| `CircuitLoader.getAfferentSynapses` + the projection in `use-circuit-synapses` | `synapses/synapse-loader.ts` — same edge datasets, same signed-distance projection onto the drawn soma, same `rescueOffSurface` tolerance |
| `resolveMorphologyRule` / `morphologyFileOf` / `listBiophysicalPopulations` | `sources/resolve-morphology-path.ts` — same resolution, as a module that does not hang off a class being deleted |

The synapse loader takes an injected file handle instead of reaching into a loader, and being a function over an open file rather than a method on a downloader is what makes it testable: the new `afferent-synapses.test.ts` builds a real HDF5 edge file and covers the reading contract — `target_node_id` sitting outside `0/`, populations with no `afferent_surface` datasets, and the handle closing when one fails.

### Kept from #1863, deliberately

`sequential-loader.ts` and its two queues, the OBI-One API client, `build-morpho-tree.ts`, and the `sonataSectionIds` index. The morphology-location picking depends on all of it, and the source hook returns the index exactly as before. The single-cell section-type colouring and soma radius are kept as they are too, so this PR changes no appearance.

### Two fixes carried over rather than lost

`circuit-loader.test.ts` covered two things the replacement modules did not, so its cases live on in `resolve-morphology-path.test.ts` and the new `circuit-nodes/circuit-config.test.ts`:

- Manifest expansion has to reach `nodes_file` and `edges_file` alike; getting it wrong on one fails as a 404 far from the config that caused it.
- A population with no declared `type` now counts as biophysical when picking the default, which is what SONATA says it means. This was a real gap — `pickDefaultPopulation` compared `type === 'biophysical'` exactly, so a config listing a virtual population before an untyped one opened the virtual population, which has nothing to draw.

### Behaviour change

A circuit with more than one node population now draws the one chosen in the selector rather than every biophysical population concatenated. This is what the nodes table and colour-by have always done, and it is what keeps `colorsByNode` indexed against the cells it colours.

### Also gone

`viewer-layout/`, dead since its only entry point stopped being rendered, and the `useDownloadHandler` debug hook it was the last importer of — #1863 had already dropped the "Download SONATA" control that used it. The `MorphoViewerTreeItemType` enum and `MorphoViewerTreeItem` interface hand-copied into `scan-config/types.ts`, now re-exported from the package, since a structurally identical enum is still a different enum to TypeScript. And `SynapseGroupSchema`, left behind when the OBI-One synapse endpoint was reverted.

This does **not** remove main-thread h5wasm outright — edge files are still read there. What it removes is the second `nodes.h5` download and the second implementation of reading it.

### Verification

- 2082 unit tests pass, including #1863's `sequential-loader.test.ts` and `morphology-location-selection.test.tsx` untouched; `pnpm build` clean; Biome clean on changed files.
- `tsc` at main's 267-error baseline with **zero** errors in touched files.

**Not yet done, and needed before merge:** a single-scale circuit with an `@library`-encoded `morphology` column, one ASC-only circuit, and the morphology-location picking from #1863 end to end — click a branch, check the form row, edit the offset, watch the marker move. That last one matters most: this PR swaps the node source underneath a feature that is already live.


pgetta:
This is a preparation for implementing https://github.com/openbraininstitute/prod-circuit-simulation/issues/159.

src/features/scan-config/components/circuit-viz/synapses/synapse-loader.ts
```diff
@@ -0,0 +1,416 @@
...
+    coordinates[i * 3 + 2] = point[2];
+  }
+  report.logTask(
+    `Loaded ${arrXs.length} afferent synapses for "${populationName}": ` +
+      `${somaTotal} on a soma, ${projected} projected` +
+      (projected > 0 ? `, worst residual ${worstResidual.toFixed(3)}µm` : '') +
+      (rescued > 0 ? `, ${rescued} rescued off the surface near a soma` : '') +
+      '.'
+  );
+  probeNeuriteSynapses(report, populationName, coordinates, arrSectionId, arrTarget, surfaces);
```
bilalesi: This runs on every load in production, and the doc says "Nothing acts on the result".

It is up to 256 SDF queries, each walking every segment of a cell, only to print a median. Please put it behind the same debug flag as the report.

src/features/scan-config/components/circuit-viz/synapses/report.ts
```diff
@@ -0,0 +1,39 @@
...
+    return this;
+  }
+
+  debug() {
+    const { tasks } = this;
+    const text = tasks.map((task) => `%c${task.message}`).join('\n');
+    const styles = tasks.map((task) =>
+      task.failure ? 'background:#000;color:#f66' : 'background:#000;color:#9cd'
+    );
+    console.debug(text, ...styles);
```
bilalesi, line 35: `console.debug` fires on every synapse load, for every user. Gate it on the debug flag so normal users get a clean console.

src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts
```diff
@@ -0,0 +1,210 @@
...
+        return loaded;
+      } catch (e) {
+        // Cancellations are the axon toggle clearing the queue, and those cells
+        // are about to be re-requested anyway.
+        if (e instanceof SequentialLoaderClearedError) return null;
+        // A failed cell is drawn as its placeholder soma rather than taking the
+        // viewer down, but the failure is surfaced so `retry` has something to
+        // clear.
+        logError(`Unable to load cell "${cellId}":`, e);
+        setLoadError((previous) => previous ?? (e instanceof Error ? e : new Error(String(e))));
```
bilalesi: The comment above says the cell is drawn as a placeholder "rather than taking the viewer down". It does take it down.

`loadError` goes into `error`, and `circuit-viz.tsx:281` renders it as `absolute inset-0` — a full-canvas "Couldn't load the visualization" over a circuit that drew fine apart from one cell.

Pick one: either drop `setLoadError` here and keep the log, or fix the comment. I would drop it — the placeholder soma already tells the user.

src/features/scan-config/components/circuit-viz/synapses/synapse-loader.ts
```diff
@@ -0,0 +1,416 @@
...
+import type { Report } from './report';
+
+/** One edge population's afferent synapses, as flat `[x, y, z, …]` world coordinates. */
+export type AfferentSynapseGroup = {
+  coordinates: Float32Array;
+  populationName: string;
+};
+
+/** Where a cell sits and how it is turned, which is what puts its morphology in the world. */
+export type CellPlacement = {
```
bilalesi: This is `NodePlacement` from `circuit-nodes/geometry-utils` again — same fields, same doc comment word for word. And `placementAt` is what fills it.

Import that type instead of declaring a second one.

src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts
```diff
@@ -0,0 +1,210 @@
...
+}: TOptions): TSmallCircuitSource {
+  const { virtualLabId, projectId } = useWorkspace();
+  const circuitId = circuit.id;
+
+  // `withMorphologies` because this source resolves a morphology *file* per
+  // cell; the somas-only viewer deliberately does not.
+  const { geometry, config, isLoading, error } = useNodeGeometry({
+    circuit,
+    population,
+    withMorphologies: true,
```
bilalesi, line 76: Linking this to my #1875 comment: `placementAt` reads `geometry.orientations`, so this is the caller that actually needs them. The somas-only viewer does not.

If you make orientations opt-in over there, this call needs `withOrientations: true`.

src/morpho-viewer/tree-item-type.ts
```diff
@@ -0,0 +1,16 @@
...
+ * reads WebGL constants off `globalThis` and instantiates a WASM module while
+ * its module body runs — so importing the enum through it costs a renderer, and
+ * outside a browser it throws outright. This module is a bare enum declaration
+ * with no imports of its own, so code that only needs to tell a soma from an
+ * axon (the synapse pipeline, the section schema) can say so without any of it.
+ *
+ * The coupling to the package's internal layout is the price, which is why it
+ * is spent here, once, rather than at each call site.
+ */
+export { MorphoViewerTreeItemType } from '@openbraininstitute/morphoviewer/dist/components/morpho-viewer-simul/types/public';
```
bilalesi, line 14: The reason is right and the comment explains it well. Two things:

- `src/morpho-viewer/index.ts:8` still re-exports the same enum from the barrel. So there are now two import paths for one enum, one safe and one not. Make `index.ts` re-export from this file.
- morphoviewer 0.32.4 has no `exports` map, so the deep path works. The day it gets one, this breaks. Worth asking the package for a subpath export.

src/features/scan-config/components/circuit-viz/sources/use-afferent-synapses.ts
```diff
@@ -0,0 +1,161 @@
...
+    const openEdgesFile = async (file: string) => {
+      const response = await downloadAsset({
+        ctx,
+        entityType: EntityTypeDict.Circuit,
+        entityId: circuitId,
+        id: circuitAssetId,
+        assetPath: file,
+        asRawResponse: true,
+      });
+      const buffer = await response.arrayBuffer();
```
bilalesi, line 84: Consider: edge files are pulled fully into main-thread memory here, then read with `json_value`. That is the same pattern the two PRs just removed for nodes.

You call this out in the description, so no objection to the scope. Just noting `writeToFS`'s new 2 GiB cap now applies to edge files too, and this is the next thing to move to the worker.

src/features/scan-config/components/circuit-viz/synapses/synapse-loader.ts
```diff
@@ -0,0 +1,416 @@
...
+}
+
+async function readPopulation(
+  input: SynapseLoaderInput,
+  edgesFile: H5File,
+  populationName: string
+): Promise<AfferentSynapseGroup | null> {
+  const { report } = input;
+  report.logTask(`Reading afferent synapse positions for population "${populationName}"...`);
+  if (!hasDataset(edgesFile, `edges/${populationName}/0/afferent_surface_x`)) {
```
bilalesi: Only `afferent_surface_x` is checked. If `y` or `z` is missing, `getNumberArray` throws, the error escapes `loadAfferentSynapses`, and every population loses its synapses — not just this one.

Check all three here, or catch per population.

src/features/scan-config/components/circuit-viz/synapses/synapse-loader.ts
```diff
@@ -0,0 +1,416 @@
...
+    if (isSomaSection(sectionIds[i])) continue;
+
+    const sdf = surfaces.get(targetNodeIds[i])?.whole;
+    if (sdf) candidates.push({ index: i, sdf });
+  }
+  if (candidates.length === 0) return;
+
+  // Evenly spread over the population rather than taking a prefix, so one
+  // badly placed branch can't stand in for the whole cell.
+  const stride = Math.max(1, Math.floor(candidates.length / NEURITE_PROBE_LIMIT));
```
bilalesi: Nit: this does not cap at 256. With 511 candidates the stride is 1, so it probes all 511. Only matters for the probe, but the constant reads like a limit.

src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts
```diff
@@ -0,0 +1,210 @@
...
+    circuit,
+    config,
+    geometry,
+    loadTree,
+  });
+
+  const retry = useCallback(() => {
+    setLoadError(null);
+    setSonataSectionIds(EMPTY_SECTION_IDS);
+    sequentialCellLoader.clear();
```
bilalesi, line 207: Nit: `retry` clears `sequentialCellLoader` but not `projectionCellLoader`. If a synapse projection failed, retry will not redo it.

src/features/scan-config/components/circuit-viz/sources/use-afferent-synapses.ts
```diff
@@ -0,0 +1,161 @@
...
+  for (const edge of edges) {
+    const populations = byFile.get(edge.file);
+    if (populations) populations.push(edge.name);
+    else byFile.set(edge.file, [edge.name]);
+  }
+  return [...byFile].map(([file, populations]) => ({ file, populations }));
+}
+
+/** MEMFS is flat, so the asset path has to survive as a single filename. */
+function fsKeyOf(circuitAssetId: string, file: string): string {
```
bilalesi, line 165: Nit: `/` becomes `_`, so `a/b.h5` and `a_b.h5` map to the same FS name. Very unlikely, but a hash of the path would remove the question.

bilalesi (changes requested):
Good cleanup. The tests are the best part, building a real HDF5 edge file instead of stubbing the reader is the right call.

Two things to fix:
1. The neurite probe and the report both run in production. They only log.
2. One failed cell hides the whole viewer, which your own comment says it should not.

Rest is small, all inline.

src/features/scan-config/components/circuit-viz/synapses/synapse-loader.ts
pgetta: Gated now, together with the report below. Both go through `isMorphoViewerDebugMode`, so whoever is asking the question pays for it and nobody else does.

src/features/scan-config/components/circuit-viz/synapses/report.ts
pgetta: Done - the check sits inside `debug()` so every caller inherits it.

The flag reader lives in its own leaf module (`morpho-viewer/debug-mode.ts`) rather than coming from the hook, for the same reason `tree-item-type` exists: importing the barrel to read a localStorage key would drag in a WebGL renderer.

src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts
pgetta: You are right, it does take it down. I dropped `loadError` altogether - the state, the setter and the `??` - and kept the log. The placeholder soma is the signal, as you say.

src/features/scan-config/components/circuit-viz/synapses/synapse-loader.ts
pgetta: Fixed, it imports `NodePlacement` now. Both came out of the same move and I simply did not notice I had written it twice.

src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts
pgetta: Yes - `withOrientations: true` here, and the flag itself is in #1875.

src/morpho-viewer/tree-item-type.ts
pgetta: First one fixed: the barrel re-exports from `./tree-item-type`, so there is one path and it is the cheap one.

On the subpath export, agreed - and it would let both this file and the new `debug-mode.ts` disappear. I have not raised it yet. Do you want to, or shall I?

src/features/scan-config/components/circuit-viz/sources/use-afferent-synapses.ts
pgetta: Agreed, and it is the next thing I would pick up. Worth saying that `writeToFS`'s cap bites harder here than for nodes: an edge file is routinely the larger of the two.

src/features/scan-config/components/circuit-viz/synapses/synapse-loader.ts
pgetta: Good catch. It checks all five datasets it goes on to read now, and the log names the missing ones.

I also wrapped the per-population call in a try/catch, because the guard only covers the missing-dataset case and anything else thrown in there had the same reach. Two cases for it: a population missing one axis, and one failing mid-read while its neighbour still loads.

src/features/scan-config/components/circuit-viz/synapses/synapse-loader.ts
pgetta: Fixed, `Math.ceil`.

src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts
pgetta: Fixed, both queues cleared. The comment explaining why the projection has its own queue was the thing that made it easy to miss.

src/features/scan-config/components/circuit-viz/sources/use-afferent-synapses.ts
pgetta: Fixed with a small FNV-1a hash of the path, so the separators stop mattering.

pgetta:
Same here, all addressed.

The `afferent_surface_x` one was the most valuable of the set: a population missing an axis would have cost every other population its synapses too, and no test would have caught it.

bilalesi:
@pgetta there is few regressions, 
- morphology locations now is not displayed,
- the locations label is not displayed too (from the viewer settings menu)

pgetta:
Good catch, thanks Bilal - both are the same regression.

`useMorphologyLocationSelection` looks the SONATA section-id index up by `cell.id`, which is the viewer id: node key plus the axon flag. The new source keyed that index by the argument `loadCell` receives instead, and the two are not the same string - morphoviewer splits a cell id on `?` and calls `loadCell` with the path part alone (`painter-cell.js`, `const [path] = cell.id.split("?")`). So every lookup missed, no stored location resolved to a section name, and nothing was placed. Labels are derived from the markers, which is the second half of what you saw.

The old source re-keyed through `makeVizCellId` for exactly this reason and the rewrite dropped the step. The flag has to stay in the key either way, since the index is built from the filtered sections and names a different set with axons hidden.

Fixed in 5d00ce6a1, with two cases that both fail on the previous code: one calls `loadCell` the way the viewer does and reads the index back by the cell's own id, the other pins the two axon states to separate keys. Could you re-check once the preview redeploys?

bilalesi:
still another regression, 
when the user change the block, should be able to select locations in the viewer with some delay, it's the last merge to main,

pgetta:
I've clicked around but not sure I understand what does it mean to "select locations in the viewer with some delay". I've rebased the branch onto main so now it includes your latest changes - maybe that was it.

Could you please check again, or explain what you meant?

bilalesi (approved):

bilalesi:
> I've clicked around but not sure I understand what does it mean to "select locations in the viewer with some delay". I've rebased the branch onto main so now it includes your latest changes - maybe that was it.
> 
> Could you please check again, or explain what you meant?

what i mean is the following:
- if the user in the morph locations block, the popover to select the location should appear instantly
- if the user in another block other than morph location, the popover should appear with some delay (from the requirement)

PR #1879, opened 2026-08-20
feat: migrate memodel to the new 3D viewer, with a dendrogram

bilalesi:


bilalesi:
this requires morphoviewer PR: https://github.com/openbraininstitute/morphoviewer/pull/47
to be merged first

bilalesi:
> this requires morphoviewer PR: [openbraininstitute/morphoviewer#47](https://github.com/openbraininstitute/morphoviewer/pull/47) to be merged first

ready for review

src/features/scan-config/components/model-preview/circuit-preview.tsx
```diff
@@ -57,9 +60,35 @@ function circuitHasDesignerImage(circuit: ICircuit): boolean {
...
+  }
+  if (!hasDesignerImage) return ViewerModeDict.Visualization;
+  return mode;
+}
+
 interface ICircuitPreviewProps {
   className?: string;
-  circuit: ICircuit;
+  /** The circuit to preview. Omit for an MEModel and pass {@link memodelId} instead. */
+  circuit?: ICircuit;
```
pgetta: I wonder, would it make sense to have these as a union - `{ circuit: ICircuit } | { memodelId: string }`? Right now `<CircuitPreview enableVisualization />` type-checks with neither of them passed, and it renders the whole viewer chrome with its settings menu over an empty canvas: `mountViz` is true and the inner `memodelId ? ... : circuit && ...` collapses to `false`. Before this PR `circuit` was required, so TS would have caught that.

It would also let us drop the five `circuit &&` guards that are now repeated below - `mountImage && circuit`, the `CircuitVisualization` branch, `mountViz && largeCircuit && circuit`, `showTable && circuit`, `Boolean(circuit) && (features?.nodesTable ?? true)` - since only one of the two shapes could be there.

Or maybe I'm missing a caller that legitimately passes neither.

src/features/scan-config/components/model-preview/circuit-preview.tsx
```diff
@@ -57,9 +60,35 @@ function circuitHasDesignerImage(circuit: ICircuit): boolean {
...
+  /** Preview a single MEModel instead of a circuit. */
+  memodelId?: string;
```
pgetta: Just want to note a small user-facing consequence of taking only the id here - the snapshot filename. `useCircuitColorBy(enableVisualization ? circuit : undefined, ...)` gets `undefined` for a MEModel, so [`downloadCircuitImage(image, circuit?.name ?? '', ...)`](https://github.com/openbraininstitute/core-web-app/blob/591ff9e846ade020ccfa02afa2a31e3da2273104/src/features/scan-config/components/color-by/use-circuit-color-by.ts#L117) gets an empty name, and [`name || 'circuit'`](https://github.com/openbraininstitute/core-web-app/blob/591ff9e846ade020ccfa02afa2a31e3da2273104/src/features/scan-config/components/shared/3d-viewer.ts#L80) turns it into `circuit.png`. So every snapshot taken on a MEModel is downloaded as `circuit.png`, no matter which model it is.

Probably a `name` prop, or passing the MEModel entity instead of only its id, would be enough to fix it.

src/features/scan-config/components/circuit-viz/sources/use-memodel-viz-source.ts
```diff
@@ -0,0 +1,106 @@
...
+    () =>
+      sections
+        ? [
+            {
+              id: cellId,
+              center: ORIGIN,
+              orientation: IDENTITY_QUATERNION,
+              somaRadius: PLACEHOLDER_SOMA_RADIUS,
+              // Colour-by wins; otherwise colour by section type.
+              color: colorsByNode?.[0] ?? SECTION_TYPE_COLORS,
```
pgetta: From what I can see the `SECTION_TYPE_COLORS` fallback is the only branch that ever runs here: `ModelPreview` hard-sets `colorBy: false` for MEModels, so `CircuitPreview` always passes `colorsByNode: undefined` down this path - and that array is indexed by SONATA node index anyway, which a MEModel has no notion of.

Probably the `colorsByNode` option and its plumbing through `MemodelVisualization` can go. Or is it there for something coming later?

src/features/scan-config/components/model-preview/circuit-preview.tsx
bilalesi: Good catch, and you're right that it was reachable — done in 986ecf7.

`circuit`/`memodel` are a union now:

```ts
type TPreviewSubject =
  | { circuit: ICircuit; memodel?: never }
  | { memodel: TPreviewedMemodel; circuit?: never };
```

I checked it actually catches the case you described rather than assuming: `<CircuitPreview enableVisualization />` is now `TS2322`, and so is passing both.

On dropping the five `circuit &&` guards — I tried and I don't think it's worth it, so I've left them. The component destructures its props in the signature, so TS never sees an object it can discriminate and the union doesn't narrow inside the body. Three of the five are also gating on more than "is it a circuit" (`hasDesignerImage`, `largeCircuit`, `showTable`), so removing `circuit &&` there would need a non-null assertion, which is worse than the guard. The only structure that removes them is splitting into circuit/memodel inner components behind a dispatcher, and those two would share ~200 lines of overlay/table/chrome state. I've added a one-line comment at the destructure so this isn't re-litigated. Happy to do the split if you'd rather.

src/features/scan-config/components/model-preview/circuit-preview.tsx
bilalesi: You traced that exactly right — every MEModel snapshot came down as `circuit.png`. Fixed in 986ecf7 by passing the entity rather than the id, as you suggested.

`useCircuitColorBy` takes a `subject` option and falls back to the circuit:

```ts
const shown = subject ?? circuit;
useViewerConfig(shown?.id ?? '', …)
downloadCircuitImage(image, shown?.name ?? '', …)
```

It also fixes a second thing on the same line that I hadn't noticed: the saved viewer-config key was `circuit?.id ?? ''`, so **every** MEModel shared the `''` bucket. Both are inert for circuits (`subject` is undefined there, so the key is unchanged) and `PERSIST_VIEWER_CONFIG` is still `false`, so nothing is written today either way — but it stops being a trap when that flag flips.

The prop is typed as just the `id`/`name` it reads rather than `IMEModel`, which turned out to matter: `.with({ type: EntityTypeDict.Memodel })` does **not** narrow to a single entity — `type` is a wide `TEntityTypeDict` rather than a literal per entity, so the match still admits five members including `ICircuit`. A `model as IMEModel` cast would have compiled and hidden that. Asking only for what's used means the host needs no cast at all.

src/features/scan-config/components/circuit-viz/sources/use-memodel-viz-source.ts
bilalesi: You're right on both counts, and it's gone in 986ecf7 — the option, the parameter, the `useMemo` dep, and the pass-through in `MemodelVisualization`. Nothing coming later needed it.

I also narrowed the type so it can't quietly come back: `TMemodelVizProps` was `Omit<TCircuitVizViewProps, 'source'>`, which still advertised `colorsByNode`, `defaultColor` and `population` even though `CircuitVizView` drops them on this path. Those three are omitted now, so the type says what the MEModel path actually honours.

src/features/scan-config/components/model-preview/circuit-preview.tsx
bilalesi: fxed by passing the entity rather than the id, but actually it's not an issue as colorby is not enabled for memodel

src/features/scan-config/components/circuit-viz/sources/use-memodel-viz-source.ts
bilalesi: Removed

pgetta (dismissed):

bilalesi:
thanks @pgetta

pgetta (approved):

PR #1880, opened 2026-08-20
feat: zoom slider over the canvas, and a crisp gizmo

bilalesi:


bilalesi:
Second pass on the review feedback:

1. **Size** — the ruler went from 108x264 to **100x182**, so it no longer dwarfs the scalebar. Knob 14→10, labels 11→9, readout 15→11.
2. **Ticks now ride the bend.** They are drawn against the line, so the bend carries them left as it passes — that is the detail in the clip, and it is what makes the ruler read as one elastic piece instead of a line laid over a fixed grid. Track and ticks now read the same `deflection()` (a bell rather than two beziers, so it can be evaluated at any height); `zoom-slider-geometry.test.ts` pins that they agree.
3. **Readout fits.** It was clipped at the right edge, and `0.25x` was clipped at the left; the gutter and width now hold both.
4. **Reset view** — my bug, fixed in openbraininstitute/morphoviewer#48. The zoom setter was writing `cameraManager.target.zoom`, which is exactly where a camera reset returns to, so after using the slider Reset view had nothing to undo. It no longer touches the target.
5. **Gizmo** — default box 128 → **88** CSS pixels in morphoviewer#48. The axis balls and their X/Y/Z letters are drawn as a fraction of that box, so they shrink with it. The letters are baked into a texture inside `@tolokoban/tgd`, so their size *relative* to the balls is not something this repo can change — if 88 is still too loud, the next dial is the box size, or a change in tgd.

418 tests green in the touched areas (6 new for the geometry), `tsc` at baseline with nothing in touched files, Biome clean, `next build` compiles.

bilalesi:
Third pass:

1. **Left margin** — the slider now sits further in from the canvas edge (`left-3` → `left-6`). Say the word if you meant a different amount; it is one class.
2. **100x is on the ruler.** The top of the range gets the number and its own tick. 64x keeps its tick but loses its label — on a log scale 64x and 100x are only about 11px apart, so two numbers there would collide, and the one worth showing is the end of the range. Labels are now 0.25x, 1x, 4x, 16x, 100x.
3. **Halo no longer cut.** It is drawn with `overflow-visible`, so the full glow shows at the top and bottom of the track instead of being clipped by the widget box. It is `pointer-events: none`, so spilling past the box changes nothing about what is clickable.
4. **Gizmo down to 68 CSS pixels** (was 128 originally, 88 last round). The X/Y/Z letters shrink with it since they are drawn as a fraction of that box.

One limit worth stating plainly on the letters: their size *relative* to the axis balls is baked into a texture inside `@tolokoban/tgd` (`uniRadius` is hardcoded and `TgdPainterGizmo` exposes only `alignX/alignY/size/margin`). So from these two repos the box size is the only dial — it shrinks the letters in absolute terms but not their proportion. If 68 still reads as too heavy, the next step is a change in tgd itself.

419 tests green, `tsc` at the 250-error baseline with nothing in touched files, Biome clean, `next build` compiles.

bilalesi:
Two more:

**Scale bar toggle** — new "Scale bar" switch in the settings menu, on by default, wired through to both the small-circuit and large-circuit viewers (passing `false` is how the package itself leaves it out). Icon is `tabler:ruler-measure` from icones.js.org — a graduated bar with a measurement span, which is what a scale bar actually is; the other candidates (`lucide:ruler`, `tabler:ruler-2`) are diagonal drawing rulers, and `material-symbols:straighten` is a solid fill that would not sit well beside the line icons already in that menu.

I also kept your `left-1` inset on the zoom slider rather than reverting it to my `left-6`.

**On the gizmo still not being pixel-free** — I found the cause, and it is not something I can fix from either repo. Here is the chain:

`TgdPainterGizmo` does not draw the gizmo into the scene. It renders it into an offscreen framebuffer texture, then blits that texture back as a screen-aligned quad. That texture is created with `magFilter: "NEAREST"` / `minFilter: "NEAREST"`, and the quad's position comes from a float NDC centre computed from the canvas size. When that centre does not land on an exact device-pixel boundary — which it usually does not, since it is `(W - size - 2*margin)/W` for whatever `W` the canvas happens to be — nearest-neighbour sampling duplicates one row/column of texels and drops another. That is the loose pixels you are seeing, and it is why raising the resolution helps but never fully cleans it up.

The scalebar has none of this because it is a 2D canvas drawn directly at device resolution, with no intermediate texture and no resampling.

`textureFramebuffer` is private, the tips texture is built through a private path, and `TgdPainterGizmo` exposes only `alignX/alignY/size/margin` — so neither the filter nor the placement is reachable.

**The fix that would actually work**: stop blitting. `TgdCanvasGizmo` (in tgd) and morphoviewer's own `GizmoCanvas` (in `lib/src/gizmo/`, currently unused) both render the gizmo directly into their own small canvas as point sprites with a `LINEAR`-filtered atlas — no framebuffer round trip, so nothing to resample. Rendering it as a corner canvas element would fix the pixels, and as a bonus:

- the letters become ours to size (morphoviewer's `TipsPainter.paintDisk` bakes the discs and glyphs locally, so the padding you asked for is a one-line change there),
- the hand cursor becomes plain CSS on a DOM element instead of the hit-test I had to write.

I have not done it, because it swaps a working component for code that is currently dead in the repo, and I cannot verify WebGL output from here — that needs a look on your screen. Want me to go ahead? It is contained (a corner canvas plus click-to-face wiring), and it is the only route that gets the gizmo to the same crispness as the scalebar.

bilalesi:
Your screenshot pinned it — and it turned out to be fixable without rewriting the gizmo after all. Both fixes are in openbraininstitute/morphoviewer#48.

**Gizmo stair-stepping.** The cause is what I described: the gizmo is drawn into a texture and blitted back as a screen-aligned quad, placed at a float centre that rarely lands on a whole device pixel. What I had missed is that `TgdTexture2D.setParams` is **public** even though the texture itself is private — so the sampling can be changed from outside after all. It is `LINEAR` now instead of `NEAREST`, which resolves that half pixel rather than snapping to the wrong texel. That is the jagged ring around each disc and the blocky glyph edges in your image.

This is a two-line change rather than the canvas rewrite I was proposing, so I went ahead with it. It also re-applies after a resize, since resizing reloads the texture.

**Background not applying until the panel closes.** A real bug, and a plain one: the background setter updated the clear painter's colour but never asked for a frame. Nothing else did either, so the canvas kept the old colour until some unrelated repaint came along — closing the settings popover being the obvious one. It paints immediately now.

Both need looking at running, since I cannot see WebGL output from here. If the gizmo edges are clean now, the canvas rewrite is off the table and the only thing still out of reach is the glyph-to-disc padding (baked into a texture inside tgd).

402 tests green, `next build` compiles.

src/ui/molecules/slider/ruler-slider.tsx
```diff
@@ -0,0 +1,294 @@
...
+  const snapToTick = () => {
+    const target = snapSliderValue(valueAt(x.get()), min, max, step);
+    const snapped = travelOf(target);
+    const id = ++gesture.current;
+    if (reduce) {
+      x.set(snapped);
+      interacting.current = false;
+      return;
+    }
+    animate(x, snapped, SPRING_SNAP).then(() => {
```
pgetta: I think `interacting.current` can get stuck at `true` here. In the non-`reduce` path this `.then()` is the only thing that clears it, and framer doesn't resolve `finished` when an animation is cancelled - `JSAnimation.finish()` calls `notifyFinished()`, `cancel()` doesn't. Drag's `onSessionStart` calls `x.stop()` on pointerdown, before the drag threshold is crossed.

So let's imagine the following situation: momentum ends, the snap spring starts, and the user taps the strip without dragging it. The spring is cancelled, this callback never runs, and there is no `onDragEnd` after it to recover the flag. From then on the sync effect below bails out and the needle stops following canvas scroll-zoom, while the readout keeps updating (it renders `current`, not `x`) - so the needle and the number visibly disagree. It heals on the next completed drag or arrow key, but not before that.

Clearing the flag from an `onStop`/`onComplete` callback instead of the promise would cover both paths. Please correct me if I'm wrong about how framer resolves `finished` here.

src/ui/molecules/slider/use-slider.ts
```diff
@@ -0,0 +1,170 @@
...
+  const hi = max > min ? max : min;
+  const stride = step > 0 ? step : 1;
+  const current = clamp(controlled ? value : internal, lo, hi);
+  const percent = hi > lo ? ((current - lo) / (hi - lo)) * 100 : 0;
+
+  const commit = useCallback(
+    (next: number) => {
+      const clean = snapSliderValue(next, lo, hi, stride);
+      if (!controlled) setInternal(clean);
+      onValueChange?.(clean);
```
pgetta: `commit` fires `onValueChange` unconditionally, and `RulerSlider` calls it from `useMotionValueEvent(x, 'change', ...)` - so during a drag this runs on every frame, with the same snapped stop for most of them. Each call goes `setZoom` -> `signals.setZoom.dispatch` -> `manager.zoom`'s equal-value branch -> `eventZoom` -> `setValue` -> a full `CircuitPreview` render, on top of the WebGL canvas.

Would be nice to bail out when the snapped value hasn't changed, something like `if (clean === current) return;` before the dispatch. That is exactly the churn the `sharedVizProps` memo comment is trying to avoid.

pgetta (commented):
Nice work, Bilal - putting the labelled ticks on powers of two makes the whole 0.25x-100x range readable in 200px.

I've left two comments, both around the drag path.

src/ui/molecules/slider/use-slider.ts
bilalesi: fixed

pgetta (approved):

PR #1883, opened 2026-08-24
feat(526): block draft and disqualified entities from workflows

bilalesi:


src/entity-configuration/domain/workflow-lifecycle-eligibility.ts
```diff
@@ -0,0 +1,69 @@
...
+export function isEntitySelectableForWorkflow(entity: TLifecycleStatusCarrier): boolean {
+  return getWorkflowLifecycleBlockReason(entity) == null;
+}
+
+export function getWorkflowLifecycleBlockReason(
+  entity: TLifecycleStatusCarrier
+): string | undefined {
+  const status = entity.lifecycle_status;
+  if (status == null) return undefined;
+  return BLOCKED_LIFECYCLE_REASONS[status as TEntityLifecycleStatus];
```
pgetta: The JSDoc above says unknown wire values stay allowed, but this is a plain index into an object literal, so it walks the prototype chain too. If a status like `constructor` or `toString` ever comes over the wire, this returns a function instead of `undefined` - `isEntitySelectableForWorkflow` then flips to `false` and the reason lands as a React child in `WorkflowBlockedActionTooltip`, which throws.

Granted, that will not happen with a `StrEnum` on the entitycore side, but [`Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) keeps the documented behaviour for free:

```suggestion
  return Object.hasOwn(BLOCKED_LIFECYCLE_REASONS, status)
    ? BLOCKED_LIFECYCLE_REASONS[status as TEntityLifecycleStatus]
    : undefined;
```

src/api/entitycore/types/shared/global.ts
```diff
@@ -20,12 +20,30 @@ export type EntityCoreType = {
...
+  [EntityLifecycleStatus.Draft]: 'Draft',
+  [EntityLifecycleStatus.Active]: 'Active',
+  [EntityLifecycleStatus.Disqualified]: 'Disqualified',
+} as const satisfies Record<TEntityLifecycleStatus, string>;
+
+export function getEntityLifecycleStatusLabel(
+  status: TEntityLifecycleStatus | string | null | undefined
+): string {
+  if (status == null) return '';
+  return status in EntityLifecycleStatusLabel
```
pgetta: Same here - `in` walks the prototype chain, so `getEntityLifecycleStatusLabel('toString')` gives back a function rather than the raw status. This one is a bit more visible than the other, since the value also feeds the grid column's `getValue`, which the comment there says backs free-text search and export as well.

```suggestion
  return Object.hasOwn(EntityLifecycleStatusLabel, status)
```

src/ui/segments/workflows/config/entity-entry.ts
```diff
@@ -460,7 +461,10 @@ export async function resolveWorkflowConfigureHrefForEntity({
   if (output) return buildStoredConfigurationHref({ match: output, workspace, entityId, mode });
 
   const source = findWorkflowFor('sourceType', candidates, flags);
-  if (source) return buildNewConfigurationHref({ match: source, workspace, entityId });
+  if (source) {
+    if (!isEntitySelectableForWorkflow(entity)) return null;
```
pgetta: I wonder, should this return `null`? Judging by the JSDoc a few lines above, `null` already means "no available workflow accepts the entity", and the only caller turns it into `No workflow configuration could be opened for "<label>"`.

So if a Get started example points at a source model that is still Draft, the user gets a message naming the wrong reason, with no hint that the lifecycle status is behind it - everywhere else in this PR the block comes with an explanatory tooltip.

Probably a discriminated result here, or doing the check in the caller so the notification can say why, would read better. Although I am not sure any of the current examples point at a draft source model - if none do, then this is more a note for the future than something to fix now.

pgetta (commented):
Nice work, Bilal!

I've left three small comments, nothing blocking.

src/ui/segments/workflows/config/entity-entry.ts
bilalesi: the function now returns a result object saying why it failed

src/api/entitycore/types/entities/entity.ts
```diff
@@ -3,8 +3,12 @@ import type {
   EntityAuthorization,
   EntityCoreIdentifiable,
   EntityCoreType,
+  TEntityLifecycleStatus,
 } from '@/api/entitycore/types/shared/global';
 
-export interface IEntity extends EntityCoreIdentifiable, EntityCoreType, EntityAuthorization {}
+export interface IEntity extends EntityCoreIdentifiable, EntityCoreType, EntityAuthorization {
+  /** optional here: entitycore sets it on every entity, but older reads may predate the field. */
+  lifecycle_status?: TEntityLifecycleStatus | null;
```
g-bar: What does it mean older reads may predate the field? 
I think it should closely reflect the db schema

g-bar (approved):
left a minor comment but lgtm

src/api/entitycore/types/entities/entity.ts
bilalesi: that's wrong comment, will remove it in next PR

bilalesi:
thanks both

PR #1885, opened 2026-08-24
fix(159): redraw the raster in the same task that resizes it

pgetta:
## What

Dragging the split divider in the spike replay view made the raster blink, and the 3D pane next to it (and the scan-config circuit preview) flashed black and white on a window resize. This PR fixes the raster and picks up morphoviewer 0.35.2, which fixes the 3D side (openbraininstitute/morphoviewer#51).

## Why

`RasterRenderer`'s ResizeObserver hopped to an animation frame, where `handleResize` set both canvases' sizes, which resets them, and queued the redraw for the frame after. The browser composited the empty canvases once per resize step. A smooth drag resizes on each frame, so the plot blinked on each frame.

tgd's resize handler ran the same sequence for the 3D pane.

## How

`handleResize` now runs inside the observer callback and redraws in the same task. ResizeObserver callbacks fire after layout and before paint, so the compositor reads refilled canvases. Without the rAF hop the plot also stops lagging two frames behind the divider. Resizing the canvases does not resize the observed element, since both are absolutely positioned inside an `overflow: hidden` container, so the observer does not loop.

The dependency goes from 0.35.1 to 0.35.2.

## Verification

- Biome clean, no new tsc errors in the renderer, the four spike-viewer Vitest suites pass (24 tests).
- I dragged the divider in the split view: the plot and axes stay put. jsdom has no WebGL, so there is no unit test.
- The 3D measurements are in openbraininstitute/morphoviewer#51.


pgetta:
Relates to https://github.com/openbraininstitute/prod-circuit-simulation/issues/159.

bilalesi (approved):

pgetta:
Thanks!

PR #1886, opened 2026-08-25
feat(spike-replay): default view to split

pgetta:
## Summary
- Default the spike replay view to the raster+3D split instead of raster-only, so both panes are visible without an extra click.

## Test plan
- [x] Open a spike replay for a circuit and confirm it lands on the split view.
- [x] Confirm behavior without a circuit (raster only) is unchanged.


pgetta:
Relates to https://github.com/openbraininstitute/prod-circuit-simulation/issues/159.

bilalesi (approved):

pgetta:
Thank you Bilal.

PR #1887, opened 2026-08-27
Draw every node population in the 3D circuit viewer, at region scale

pgetta:
The standalone circuit views (scan-config preview, entity details) now draw every
node population a circuit has, not just the selected one. At region scale that is
4.7M somas instead of a few hundred thousand, so most of the branch is the
performance work behind it.

## Draft: needs morphoviewer 0.36.0 published

The viewer side is openbraininstitute/morphoviewer#52, which adds the flat
`positions` input and the `cellColors` palette this branch uses. Neither is in the
published `0.35.2`, so **CI's build fails until 0.36.0 is out and the dependency
here is bumped**. `package.json` points at a tarball packed from that branch in
the meantime.

Order: #52 merges, 0.36.0 publishes, one commit bumps the dependency, then out of
draft.

## Multi-population

Every population carrying positions is drawn, in declared order, with the ones not
on show receded around the selected one. Clicking a receded soma or cell selects
its population, in the nodes table as well. `usePopulationsPlacement` reads them
all through the shared worker registry and keeps the placements, so a selection
change recolours the somas already on screen instead of rebuilding the scene and
resetting the camera. Populations without positions are not drawn.

Node keys now name the population (`circuit/population #index`), since every
population counts its nodes from 0 and the index alone was ambiguous. Spike replay
still draws a single population, because its cell indices are relative to that
population. A failed population stays subscribed and retryable; previously a retry
healed the nodes table and the colour-by panel but not the viewer.

## Choosing what is in the scene

A "Populations 3 of 4" pill sits with the nodes table and the view settings. Each
row carries two targets, because they answer different questions: the checkbox
draws a population or leaves it out, the name puts it on show. `Only` is the one
gesture that answers both.

The two viewers hide differently. The soma-only path gives the population a
palette column that draws nothing, so the `positions` array keeps its identity and
the camera stays where the user left it. The morphology path stops listing those
cells, so their morphologies are never requested from OBI-One at all.

Two states need a way out. The population on show can be hidden like any other,
which leaves the colour key and the nodes table describing cells that are not on
screen: a line under the pill names it, with a Show that puts back that one
population. Hiding everything leaves an empty scene, which gets a centred notice
with Show all rather than the loading indicator. Both live in the chrome, so the
two viewers get them from one place.

Emptying the scene unmounts the morphology viewer, so bringing everything back
refits the camera. Hiding some of the populations still does not.

## Performance

- Colouring by a continuous property took close to ten seconds, nearly all of it
  spent deciding whether the column held more than twelve distinct values by
  stringifying every node into a set: 3.7M mostly-distinct strings for a question
  the first thirteen rows had already settled. Counting the raw values and
  stopping at the limit brings it to about a third of a second.
- `ColorMapping` carries a bounded palette plus a `Uint16Array` column per node
  instead of a string per node, and a column crosses the worker boundary in the
  form the worker already holds it, transferred rather than structured-cloned. A
  region-scale recolour is two typed-array passes, and the retained cache holds
  typed arrays rather than millions of strings.
- Positions and orientations are `Float32Array` end to end: 113 MB down to 56 MB
  per population, and half the transfer. Over a ~10⁴ µm span that keeps ~10⁻³ µm
  of resolution, enough for placement, picking and electrode maths.
- The preview concatenates the placed geometries into one `Float32Array` for the
  viewer's `positions` prop, dropping the memo that built an object per soma
  (~400 MB at region scale). The debug download serialises from the typed array in
  chunked Blob parts instead of `JSON.stringify` over 4.7M tuples.

With the viewer keeping its cloud across a recolour, a selection or colour-by
change no longer reparses geometry, recomputes ambient occlusion or re-uploads
every buffer. Palette order, fallback colours, legend, soma radius and camera fit
are unchanged.

## Testing

Vitest green (212 files / 2302 tests) against the vendored 0.36.0, biome clean,
`tsc` delta clean. Coverage for population placement, the colour-by mode
threshold, the per-population colour memory, the mapping cache, the checklist and
its two notices, and what the preview hands the viewer, including that a selection
change returns the same `positions` array, since a new one would reset the camera.

Not measured yet: redraw times and steady-state heap in the browser at 4.7M. The
populations UI has only been exercised in jsdom.


pgetta:
Relates to https://github.com/openbraininstitute/prod-circuit-simulation/issues/240.

src/features/circuit-nodes/hooks/use-populations-placement.ts
```diff
@@ -0,0 +1,265 @@
...
+    };
+    // A file that has finished downloading keeps its final reading in the sum:
+    // the registry clears `progress` on ready, and subtracting a finished file
+    // would take the total backwards while its neighbours are still coming.
+    const report = (file: string, progress: DownloadProgress) => {
+      if (cancelled) return;
+      setDownloads((previous) => new Map(previous).set(file, progress));
+    };
+    // Nothing further is needed from this session, so release it.
+    const close = (population: NodePopulation) => {
```
bilalesi: path: src/features/circuit-nodes/hooks/use-populations-placement.ts

**Required.** The second population in the same file opens a new worker right after the first is released here. `fetchToFS` fires `cache.put` without awaiting it (`src/utils/h5/fs.ts:321`), so the second worker can miss the cache and download a region-scale file again. The "second read comes from the first's cache" claim in the doc comment is a race.

potential fix: await `cachePut` in the worker's `open` before resolving when the response had a length, or expose that promise on the session so this hook waits for it before opening a same-file sibling. Longer term: one session per file serving several populations.

src/features/circuit-viewer/circuit-scene.tsx
```diff
@@ -192,6 +204,16 @@ export function CircuitScene({
     onPopulationChange?.(population);
   }, [population, onPopulationChange]);
 
+  // What the viewers draw, in declared order: every population, or only the
+  // one on show.
+  const populations = useMemo((): readonly NodePopulation[] => {
+    if (showUnselectedPopulations) return circuitConfig?.nodes ?? [];
```
bilalesi: path: src/features/circuit-viewer/circuit-scene.tsx

**Required.** `populations` handed to `usePopulationsPlacement` ignores `hiddenPopulations`, so a virtual population hidden by default is still downloaded in full, read with morphologies and orientations, and blocks first paint. On main only the selected population was read.

potential fix: filter by `hiddenPopulations` before placing, and read a hidden population when it is un-hidden. A camera refit on that explicit gesture is acceptable, as already accepted for the pick buffer.

src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts
```diff
@@ -54,94 +68,204 @@ type TOptions = {
...
+      if (hidden.has(candidate.name)) return [];
+      const onShow = candidate.name === population.name;
+      const result = new Array<MorphoViewerSmallCircuitCell>(geometry.count);
+      for (let i = 0; i < geometry.count; i++) {
+        result[i] = {
+          id: makeVizCellId(makeNodeKey(circuitId, candidate.name, i), { showAxons }),
+          center: positionAt(geometry, i),
+          orientation: placementAt(geometry, i)?.orientation ?? IDENTITY_QUATERNION,
+          somaRadius: PLACEHOLDER_SOMA_RADIUS,
+          color: onShow ? (palette[columnByNode[i]] ?? paint) : recededColor,
```
bilalesi: path: src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts

**Required.** If the population on show failed to place (a virtual input picked from the checklist), `subject` is null, nothing is `onShow`, and every drawn cell gets `recededColor` with no notice. The large preview only recedes when the subject is drawn (`hasContext && subject`).

potential fix: recede only when `subject` is placed and visible, otherwise use the normal paint. Consider a chrome notice for "selected population has no positions".

src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts
```diff
@@ -54,94 +68,204 @@ type TOptions = {
...
+    !settled || placed.length > 0 ? null : (failures.values().next().value ?? null);
+  const error = configError ?? noPopulation ?? noPlacement;
+
+  // Keep what is on screen until the next scene can be drawn. A population
+  // joining or leaving the list takes a moment to read, and emptying the scene
+  // meanwhile would unmount the viewer, giving a black frame and then a camera
+  // reset. This does not apply after a failure: the error panel would sit on
+  // cells that 'Try again' is about to replace.
+  const shownRef = useRef<MorphoViewerSmallCircuitCell[]>(NO_CELLS);
+  const cells = built ?? (error ? NO_CELLS : shownRef.current);
```
bilalesi: path: src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts

**Required.** Reading `shownRef.current` in render makes React Compiler skip this hook ("Cannot access refs during render"). It compiled on main.

potential fix: keep the last built scene in state:
```ts
const [shown, setShown] = useState(NO_CELLS);
if (built && built !== shown) setShown(built);
const cells = built ?? (error ? NO_CELLS : shown);
```

src/features/scan-config/components/color-by/palette.ts
```diff
@@ -198,59 +207,156 @@ function buildCategorical(
...
+  library: string[],
+  indices: Uint32Array,
+  overrides?: Record<string, string>,
+  background?: string
+): ColorMapping {
+  // occurrences per library slot. Indices past the library (malformed files)
+  // count under their own `String(slot)` name, as before.
+  let slotCount = library.length;
+  for (let i = 0; i < indices.length; i++) {
+    if (indices[i] >= slotCount) slotCount = indices[i] + 1;
```
bilalesi: path: src/features/scan-config/components/color-by/palette.ts

**Optional.** One out-of-range `@library` index in a file makes `slotCount` that large, and `new Uint32Array(slotCount)` below allocates gigabytes or throws. The old code only did a lookup with a string fallback.

potential fix: cap `slotCount` at `library.length` and route bad indices to one "unknown" slot.

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
```diff
@@ -120,54 +128,160 @@ export function LargeCircuitPreview({
...
+    // takes a palette column on first use.
+    let ownRampColumn: number | undefined;
+    let recededColumn: number | undefined;
+    let hiddenColumn: number | undefined;
+
+    const columnByCell = new Uint16Array(positions.length / 3);
+    let index = 0;
+    for (const { population: candidate, geometry } of placed) {
+      if (hidden.has(candidate.name)) {
+        hiddenColumn ??= palette.push(false) - 1;
```
bilalesi: path: src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx

**Optional.** The React Compiler version installed here (1.0.0) does not lower `??=`, so this whole component is no longer compiled (it was on main). This memo is also the hot recolour path.

potential fix: replace the three `??=` (lines 249, 257, 260) with plain `if (x === undefined) x = ...`.

src/features/scan-config/components/color-by/use-circuit-color-by.ts
```diff
@@ -145,10 +163,18 @@ export function useCircuitColorBy(
...
+    [update]
+  );
+
   const colorBy: ColorByControls = useMemo(
     () => ({
       selectedProperty: property,
-      onSelectProperty: (p) => update({ colorByProperty: p }),
+      onSelectProperty: (p) => {
+        if (populationName === undefined) return;
+        update({ colorByProperty: { ...config.colorByProperty, [populationName]: p } });
```
bilalesi: path: src/features/scan-config/components/color-by/use-circuit-color-by.ts

**Optional.** This merges from the render-time `config`, so two updates in one tick clobber each other.

potential fix: a functional `update`, or a `setColorByProperty(population, p)` that merges inside `setConfig`.

src/__tests__/circuit-nodes/populations-menu.test.tsx
```diff
@@ -0,0 +1,210 @@
...
+    expect(screen.queryByRole('button', { name: /^cortex/ })).toBeNull();
+    // The checkbox is still there: pinning which population is on show says
+    // nothing about which ones are drawn around it.
+    expect(checkbox('cortex')).toBeInTheDocument();
+  });
+});
+
+describe('PopulationsMenu introduction', () => {
+  // No record of a previous visit, the only case it opens itself in.
+  beforeEach(() => localStorage.clear());
```
bilalesi: path: src/__tests__/circuit-nodes/populations-menu.test.tsx

**Optional.** On Node 24+ the global `localStorage` is Node's own stub and is `undefined` without a flag, so this whole describe block fails locally (16 failures across this file and `circuit-viewer-chrome-populations.test.tsx` on Node 26; passes on Node 22).

potential fix: call `installLocalStorage()` from `src/__tests__/auth-manager/install-local-storage.ts` in `beforeEach`.

src/__tests__/scan-config/circuit-viewer-chrome-populations.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+  onSelect: vi.fn(),
+};
+
+function renderChrome(viz: ICircuitViewerChromeProps['viz'], vizActive = true) {
+  render(<CircuitViewerChrome vizActive={vizActive} viz={viz} />);
+}
+
+// Seeded as already introduced, so the checklist stays shut and its "Show all"
+// is not on screen beside the notices below.
+beforeEach(() => localStorage.setItem(POPULATIONS_MENU_INTRODUCED_KEY, '1'));
```
bilalesi: path: src/__tests__/scan-config/circuit-viewer-chrome-populations.test.tsx

**Optional.** Same as in `populations-menu.test.tsx`: `localStorage` is `undefined` on Node 24+, so every test in this file fails there.

potential fix: `installLocalStorage()` in `beforeEach`.

src/__tests__/scan-config/circuit-viz-empty-scene.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+  // The indicator waits on the viewer's own paint progress, and with no cells
+  // the viewer is never mounted to report any. Waiting on it would leave the
+  // indicator up for good over a scene the user emptied on purpose.
+  it('reads a finished, empty scene as finished', async () => {
+    draw({ cells: [], isLoading: false });
+
+    // Waited out: the indicator holds itself back for a moment before it
+    // appears, so an immediate check would read as absence either way. By role,
+    // since which phase it would have named is not the point.
+    await expect(screen.findByRole('status', {}, { timeout: 400 })).rejects.toThrow();
```
bilalesi: path: src/__tests__/scan-config/circuit-viz-empty-scene.test.tsx

**Optional.** A real 300 ms `QUIET_MS` timer races a 400 ms `findByRole` timeout. On a slow runner a wrong implementation can still pass, and a right one idles 400 ms.

potential fix: `vi.useFakeTimers()`, `act(() => vi.advanceTimersByTime(QUIET_MS + 1))`, then a synchronous `expect(screen.queryByRole('status')).toBeNull()`.

src/__tests__/scan-config/visualization-loading-indicator.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+  it('shows nothing at all while the wait is still short', () => {
+    const { container } = render(<VisualizationLoadingIndicator download={null} />);
+
+    expect(container).toBeEmptyDOMElement();
+  });
+
+  it('names the node download, with the bytes on both sides', async () => {
+    render(<VisualizationLoadingIndicator download={{ received: RECEIVED, total: TOTAL }} />);
+
+    expect(await screen.findByText('Downloading nodes… 42 MB of 180 MB')).toBeInTheDocument();
```
bilalesi: path: src/__tests__/scan-config/visualization-loading-indicator.test.tsx

**Nit.** Each `findByText` here waits out the real 300 ms quiet period.

potential fix: fake timers and advance past `QUIET_MS`, then assert synchronously.

src/features/circuit-nodes/components/populations-menu.tsx
```diff
@@ -0,0 +1,285 @@
...
+}: PopulationsMenuProps) {
+  const [open, setOpen] = useState(false);
+  const contentRef = useRef<HTMLDivElement>(null);
+  const triggerRef = useRef<HTMLButtonElement>(null);
+
+  // The pill counts what is on screen without saying that the populations
+  // missing from it were left out by default, or that a click brings them back.
+  // The panel says so itself, once.
+  useEffect(() => {
+    if (!autoOpen) return;
```
bilalesi: path: src/features/circuit-nodes/components/populations-menu.tsx

**Nit.** The one-time introduction opens even when `hidden` is empty, so the one shot is spent on a circuit where the pill hides nothing.

potential fix: add `hidden.length > 0` to this condition.

bilalesi (commented):
Review summary: the design (placements kept, selection = recolour, typed arrays end to end) is sound. Inline comments below, most severe first. Verified locally: changed tests pass on Node 22 (CI), TypeScript 5.9 clean, React Compiler output compared against main.

src/__tests__/circuit-nodes/populations-menu.test.tsx
```diff
@@ -0,0 +1,210 @@
...
+    expect(onChange).not.toHaveBeenCalled();
+  });
+
+  // The panel is portalled, so closing on an outside click is our own
+  // handler's job rather than the DOM's.
+  it('closes on a pointer landing outside it', () => {
+    open();
+    expect(screen.getByTestId('populations-menu-content')).toBeInTheDocument();
+
+    fireEvent.pointerDown(document.body);
```
bilalesi: path: src/__tests__/circuit-nodes/populations-menu.test.tsx

**Nit.** A pointerdown on `document.body` is already handled by Radix, so this test passes with the custom capture-phase listener deleted. The case the listener exists for (a canvas that stops propagation) is not exercised.

potential fix: dispatch on a sibling `div` whose bubble-phase `pointerdown` handler calls `stopPropagation()`, and assert the panel closed.

pgetta: Bilal, I checked this one by deleting the listener and the test does fail - Radix doesn't close the panel on a `pointerdown` landing on `document.body` here, so it is our own handler that closes it.

Your underlying point stands though: nothing was exercising the capture phase. I've kept this test and added one that dispatches on a sibling whose bubble-phase handler calls `stopPropagation()` - that one fails if the listener is registered with `false` instead of `true`.

src/features/circuit-viewer/circuit-scene.tsx
pgetta: It seems that adding lazy population loading creates multiple issues (state management, camera re-centering, degraded interactivity, etc.), requires changes to the morphoviewer lib and actually doesn't bring that much of a value, since virtual populations in most of the circuits on the platform have smaller size than the biophysical population itself.

For now I propose to leave this as is with potential to revise later once we will have circuits of such a scale where this will matter.

src/features/circuit-nodes/hooks/use-populations-placement.ts
pgetta: Fixed, and the race was reproducible. `fetchToFS` awaits the `cache.put` before returning now, so the sibling worker's `cache.match` sees a committed entry. The cost is a finished progress bar held while the disk drains the tee's backlog, which seemed the better of the two.

Two tests in `fs.test.ts` pin it. The fake cache had to grow a real turn of the event loop in its `put` first, otherwise an unawaited put still commits in time and the race cannot be reproduced at all.

src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts
pgetta: Good catch, the two viewers really did disagree here. The colour is decided per population now, `onShow || !subject ? paint : recededColor`, so a selection that could not be placed leaves the circuit painted as usual rather than grey.

I left the chrome notice out for the moment. The source would need a new signal up to `CircuitViewerChrome` for it, and that felt like a change of its own.

src/features/scan-config/components/circuit-viz/sources/use-small-circuit-source.ts
pgetta: Confirmed, I ran the plugin over the file and it bails with `Cannot access refs during render`, which it does not do on main.

The scene is kept in state now, with one difference from your snippet: `const cells = built ?? (error ? NO_CELLS : shown); if (cells !== shown) setShown(cells);`. Storing `built` alone would hold the last scene across an error, and dropping it is what the `error ? NO_CELLS` branch is for.

src/features/scan-config/components/color-by/palette.ts
pgetta: Done, close to what you suggested. The counts array is `library.length + 1` and every index the library does not name shares the last slot, which the key calls `unknown`.

Added a test with indices at 4e9 and 2^32-2, which fails against the old code.

src/features/scan-config/components/model-preview/large-circuit-preview/large-circuit-preview.tsx
pgetta: Confirmed with the plugin, `(BuildHIR::lowerExpression) Handle ??= operators in AssignmentExpression`. The three are plain `if (x === undefined)` now.

Btw, while checking this I ran the compiler over every file the branch changes and diffed the result against main. There was a third one: `populations-menu.tsx` bails on the `||` and the optional chain inside its try. Fixed as well. The three that still bail (`circuit-scene`, `use-viewer-config`, `morpho-viewer/index`) bail on main too, so the branch adds none.

src/features/scan-config/components/color-by/use-circuit-color-by.ts
pgetta: `useViewerConfig.update` takes `Partial<ViewerConfig> | ((previous) => Partial<ViewerConfig>)` now, and both per-key merges go through it, `onSelectProperty` and `onChangeCategoryColor` alike. It also drops `config.colorByProperty` and `config.colorOverrides` from the deps.

src/__tests__/circuit-nodes/populations-menu.test.tsx
pgetta: Done. They do still pass on Node 24.19 here since jsdom supplies the global, but `installLocalStorage()` in `beforeEach` costs nothing and the auth-manager tests already do it.

src/__tests__/scan-config/circuit-viewer-chrome-populations.test.tsx
pgetta: Same here, `installLocalStorage()` in the file's `beforeEach` before it seeds the introduced key.

src/__tests__/scan-config/circuit-viz-empty-scene.test.tsx
pgetta: Done. `QUIET_MS` is exported, the file runs on `vi.useFakeTimers()`, and the three tests advance past it and assert synchronously.

You are right that the 400 ms was the weak part of it: a wrong implementation could win that race on a slow runner and the test would still be green.

src/__tests__/scan-config/visualization-loading-indicator.test.tsx
pgetta: Done, same treatment. The five assertions in that file no longer wait out a real 300 ms each.

src/features/circuit-nodes/components/populations-menu.tsx
pgetta: Added, with a test for it. It also caught two tests in `circuit-viewer-chrome-populations.test.tsx` that were introducing a checklist with nothing hidden, which is not the case the introduction is there for.

bilalesi (approved):
nice job Pavlo!

pgetta:
Thank you!

PR #1888, opened 2026-08-27
chore: upgrade nextjs to 16.3.3

bilalesi:
https://nextjs.org/blog/august-2026-security-release

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -11,15 +11,15 @@ minimumReleaseAgeStrict: true
 minimumReleaseAgeExclude:
   - '@openbraininstitute/*'
   - '@tolokoban/tgd'
-  - '@next/env@16.3.0'
-  - '@next/swc-darwin-arm64@16.3.0'
-  - '@next/swc-darwin-x64@16.3.0'
-  - '@next/swc-linux-arm64-gnu@16.3.0'
-  - '@next/swc-linux-arm64-musl@16.3.0'
-  - '@next/swc-linux-x64-gnu@16.3.0'
-  - '@next/swc-linux-x64-musl@16.3.0'
-  - '@next/swc-win32-arm64-msvc@16.3.0'
-  - '@next/swc-win32-x64-msvc@16.3.0'
-  - next@16.3.0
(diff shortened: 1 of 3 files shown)


pnpm-workspace.yaml
```diff
@@ -11,15 +11,15 @@ minimumReleaseAgeStrict: true
...
+  - '@next/swc-darwin-arm64@16.3.0 || 16.3.3'
+  - '@next/swc-darwin-x64@16.3.0 || 16.3.3'
+  - '@next/swc-linux-arm64-gnu@16.3.0 || 16.3.3'
+  - '@next/swc-linux-arm64-musl@16.3.0 || 16.3.3'
+  - '@next/swc-linux-x64-gnu@16.3.0 || 16.3.3'
+  - '@next/swc-linux-x64-musl@16.3.0 || 16.3.3'
+  - '@next/swc-win32-arm64-msvc@16.3.0 || 16.3.3'
+  - '@next/swc-win32-x64-msvc@16.3.0 || 16.3.3'
+  - next@16.3.0 || 16.3.3
 blockExoticSubdeps: true
```
pgetta: I was surprised to learn we have so many overrides for the `minimumReleaseAge` which contradicts why that param was put there in the first place - to prevent the supply chain attacks.

It does make sense to keep dependencies we control, e.g. `@openbraininstitute/*` but as for the rest - I would remove them and follow the policy. Unless there is a critical bug fixed - do we need to rush to install something right after the release?

bilalesi: i would exclude patches specifically for the main framework (nextjs)
all the above are related to nextjs, but i agree for other libraries

pgetta: But what's the reason to treat Next.js or its dependencies differently?

It seems to me any external package carries more or less same level of risk in comparison to our org packages (given the same protection applies for their dependencies).

bilalesi: this is different, these are security fixes, and should be treated differently

pgetta: I've missed the link in the description, thought it's a normal dependency update.

pgetta (approved):

PR #1889, opened 2026-08-28
Add Courses.md 

g-bar:
To share with Eilif as requested by Natali.


bilalesi (approved):

courses/COURSES.md
```diff
@@ -0,0 +1,72 @@
...
+![Courses Interface](courses-interface.png)
+
+To enable the courses feature, you need to get an Education license for your course. Please contact your OBI representative for information.
+
+Once we've received your payment we'll provision a special virtual lab for your course and a template project. It will be available at:
+
+`https://openbraininstitute.org/app/virtual-lab/{virtual-lab-id}/{project-id}/course`
+
+
+Before we provision your course you have to provide the course start date, period 1 end date and course end date.
```
pgetta: Maybe it's worth to mention what the "_period 1" is?

I've found what it does further down in the document but a reader might be initially confused.

The term seems to me very generic, maybe there is a better name for that?

courses/COURSES.md
```diff
@@ -0,0 +1,72 @@
...
+    - Period 1 has not ended
+
+### Course end date
+
+When the course ends, the students will lose access to their projects, but faculty (the virtual lab admin) will continue to have access to all student projects. Any unused credits in either the template or the student projects will be deprovisioned.
+
+
+### Assigning Students
+
+Through the UI you can add students to the course, you will need to provide a CSV file with 2 columns:
```
pgetta: Are headers needed (or supported) in the CSV? If both options work - it's good as is, otherwise it's worth to mention explicitly.

pgetta (commented):
Nice work Gil,

I've left couple of comments.

courses/COURSES.md
g-bar: No headers needed.

courses/COURSES.md
g-bar: this is from Georges plan: https://openbraininstitute-my.sharepoint.com/:w:/g/personal/georges_khazen_openbraininstitute_org/IQDwMTGDrvpfR5i-PpzZtbjvAdb9_1VNXLgJv3C0scnR9Rk?e=p6AbWs

I believe it's standard in academia, it's the period during which students can still enrol after the semester has already started, (usually 2 weeks after start date)

g-bar:
I'll merge for now we can improve later.

PR #1890, opened 2026-08-31
chore(listing): remove the legacy antd data-table system

bilalesi:
the aggrid migration is complete: every routed dataType resolves to a grid definition (asserted by registry-coverage.test.ts), so both antd fallbacks were unreachable.

Removed:
- browse-entity-legacy.tsx and the router branch that selected it
- the parallel circuit router (explore/circuit/index.tsx); the circuit page now renders the shared BrowseEntityScope like every other listing
- ui/segments/data-table: MainTable, BaseTable, search, skeleton, pagination, table-tools, controls, expandable-row/, listing-filter-panel/ and orphaned CSS
- explore/circuit legacy table + expandable-row helpers
- specs covering only the deleted components

Untangled first, so nothing live was deleted:
- BrowseEntityScopeProps moved to listing/types.ts; mainTableProps is now an explicit ListingTableProps instead of Partial<ComponentProps<typeof MainTable>>, which was the type-level tie between the new host and the antd table
- circuitListingRowClass moved into explore/circuit/helpers.ts, freeing circuit-listing-grid.tsx (its only surviving export)
- dropped the lazy `import('@/ui/segments/data-table')` prefetch that kept the whole antd chunk reachable from the workflow browse listing
- removed searchOpenOnMount / keepSelectionOnScopeChange from call sites; both were read only by the legacy table

An unregistered dataType now logs and renders nothing rather than silently falling back to a second implementation.

40 files, ~6.4k lines.


pgetta (commented):
Bilal, that is a big cleanup, 6.4k lines out and the AG path standing on its own.

I've left two blocking comments (the `EntityCoreResponse` import, and `EFeatureExtractionResult` reaching a blank listing) and the rest are optional, mostly leftovers the deletion orphaned.

Happy to take the optional ones in a follow-up PR.

src/features/views/listing/types.ts
```diff
@@ -0,0 +1,104 @@
+import type { ComponentProps } from 'react';
+import type { TExtendedEntitiesTypeDict } from '@/api/entitycore/types/extended-entity-type';
+import type {
+  EntityCoreIdentifiableNamed,
+  EntityCoreResponse,
+} from '@/api/entitycore/types/shared/global';
```
pgetta: **Blocking.** This import breaks the typecheck: `EntityCoreResponse` is not exported from `shared/global`, it lives in `shared/response.ts`, which is where the deleted `browse-entity-legacy.tsx` imported it from.

`tsc --noEmit` reports TS2305 here, so `listQueryFn`'s return type stays unresolved for both call sites (`workflows/browse/listing.tsx`, `browse-widget.tsx`). It gets through the build only because `next.config.ts:91` sets `typescript.ignoreBuildErrors: true`.

```suggestion
import type { EntityCoreIdentifiableNamed } from '@/api/entitycore/types/shared/global';
import type { EntityCoreResponse } from '@/api/entitycore/types/shared/response';
```

src/features/views/listing/browse-entity.tsx
```diff
@@ -19,13 +18,16 @@ const BrowseEntityGrid = dynamic(
 export type { BrowseEntityScopeProps };
 
 /**
- * Per-entity router: a dataType registered in the data-grid entitycore registry renders
- * the AG Grid listing, everything else the legacy antd table.
+ * Entity listing. Every routed dataType must be registered in the data-grid entitycore
+ * registry — see `registry-coverage.test.ts`, which asserts exactly that for each real
+ * call site. An unregistered type renders nothing rather than silently falling back to
```
pgetta: **Blocking.** The guardrail this docstring leans on has a hole, and the hole is already open.

`EFeatureExtractionResult` is in the route's `AllowedEntities` (`entity/[type]/page.tsx:21`) and `getEntityByExtendedType` resolves it, so the page reaches `BrowseEntityScope`. But `DATA_BROWSE_ENTITIES` in `registry-coverage.test.ts:18` does not list it, and no schema in `bindings/entitycore/schemas/` declares that dataType, so `getEntityGridDefinition` returns undefined and the user gets a blank page. It is nav-linked from `BrowseExperimentalDataExtendedTypes` (`explore/helpers.ts:62`) whenever `eFeatureExtractionFlag` is on, so local and preview hit it today. Before this PR the same input rendered a working antd table.

Btw, for the picker the "each real call site" claim does not hold in principle either: `browse-widget.tsx:117` resolves accepted types at runtime from the server-supplied scan-config schema (`acceptedEntityTypesFromField`, `helpers.ts:203`), and schema-derived types take precedence over the static `configurationInputs`. A hand-maintained list cannot mirror that, so the docstring promises more than the test can deliver.

src/features/views/listing/browse-entity.tsx
```diff
@@ -19,13 +18,16 @@ const BrowseEntityGrid = dynamic(
...
+ * a second implementation.
  */
 export function BrowseEntityScope(props: BrowseEntityScopeProps) {
   const definition = getEntityGridDefinition(props.dataType);
-  if (definition) {
-    return <BrowseEntityGrid {...props} definition={definition} />;
+  if (!definition) {
+    log('error', `[BrowseEntityScope] no grid definition registered for "${props.dataType}"`);
```
pgetta: Not blocking, but I would still do it here: `log` is a no-op in production (`utils/logger.ts:8` gates on `DEPLOYMENT_ENV !== 'production'`). A dataType that slips through gives the user an empty panel and an empty console, with no way to tell it apart from an empty result set and nothing to report to support.

The grid host already owns the right UX for this, `renderListingError` + `GenericError` in `browse-entity-grid.tsx:537`. Reusing it here is what makes dropping the fallback safe.

src/features/views/listing/browse-entity-legacy.tsx
```diff
@@ -1,580 +0,0 @@
-'use client';
```
pgetta: Not blocking. Three places still tell the next developer that the fallback exists:

* `src/features/data-grid/GUIDE.md:102-131` shows `return <BrowseEntityScopeLegacy {...props} />`, states "Absent => the untouched legacy antd table. **Rollback is deleting the line**", and points at `explore/circuit/index.tsx`, which this PR deletes. It also describes the parity tests as pinned against "the **legacy** implementation, which is still in the tree and is the oracle".
* `bindings/entitycore/registry.ts:53-57`: "removing the entry falls back to the legacy antd table".
* `registry-coverage.test.ts:9`: "an unregistered dataType silently falls back to the legacy antd table".

The GUIDE one is the one I would fix first. Someone hits a bad schema, follows the documented rollback and ships an empty page.

src/__tests__/browse-entity/table-download.spec.tsx
```diff
@@ -1,191 +0,0 @@
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
```
pgetta: Not blocking, but this spec was testing `EntityDownloadButton`, not the wrapper that got deleted. `TableControls` was a ~15 line pass-through (`controls.tsx:36-42`), and all seven assertions here (singular/plural label, `downloadArchive(type, ids, workspace, archiveName)` with the archive name only for a single row, disabled while loading, selection cleared after 2s) describe `download-button.tsx`, which `browse-entity-grid.tsx:53` still imports and renders.

`grep -rn EntityDownloadButton src/__tests__` now returns nothing, so bulk download for every data listing is uncovered. Swapping the props would have kept it.

src/ui/segments/workflows/browse/listing.tsx
```diff
@@ -432,14 +432,6 @@ function WorkflowNewBrowsePage({ activity, section, targetType }: WorkflowNewBro
     handleBack,
   ]);
 
-  // warm the lazy data-table chunk while the user is on the prerequisite step, so confirming
-  // resolves near-instantly instead of waiting on a first-time chunk download
-  useEffect(() => {
-    if (showPrerequisitePhase) {
-      void import('@/ui/segments/data-table');
-    }
-  }, [showPrerequisitePhase]);
```
pgetta: I wonder, was this meant to be repointed rather than removed? The listing that mounts after the prerequisite step is `BrowseEntityGrid`, loaded through `next/dynamic` in `browse-entity.tsx:13`, which is a different chunk and nothing warms it now. On a cold cache the user waits for a first-time AG Grid download at the exact moment this effect was written to avoid.

`void import('@/features/data-grid/host/browse-entity-grid')` in the same effect would keep the behaviour. Not blocking.

src/app/app/virtual-lab/[virtualLabId]/[projectId]/data/(browse)/browse/entity/circuit/page.tsx
```diff
@@ -1,16 +1,12 @@
...
   return (
-    <BrowseCircuit
+    <BrowseEntityScope
       section={WorkspaceSection.Data}
       dataType={ExtendedEntitiesTypeDict.Circuit}
-      mainTableProps={{
-        selectionType: undefined,
-        searchOpenOnMount: true,
-      }}
     />
```
Comment (pgetta) [inline src/app/app/virtual-lab/[virtualLabId]/[projectId]/data/(browse)/browse/entity/circuit/page.tsx]: Not blocking, a note for whoever touches this next. The deleted `BrowseCircuit` stripped `mainTableProps` and carried a comment explaining why; now the behaviour depends on the prop being absent and nothing says so.

`browse-entity-grid.tsx:531-533` mounts `CircuitGridBody` only when `section === Data && !props.mainTableProps?.selectionType`. Adding `selectionType: 'checkbox'` here, which is the natural edit since `[type]/page.tsx:71` gives every other data listing that plus `allowDownload`/`allowDelete`, drops the flat/hierarchy toggle, subcircuit expansion and the view-toggle slot with no warning.

A one-line comment would save someone that debugging session.

src/ui/segments/data-table/table.tsx
```diff
@@ -1,349 +0,0 @@
-'use client';
```
pgetta: Not blocking, cleanup for a follow-up if you prefer. This file was the only consumer of `use-row-selection.ts`, so `grep -rn useRowSelection src/` now matches its own definition and nothing else. The ~95 lines survive because `download-button.tsx` and `delete-button.tsx` import `RenderButtonProps` from it, which also keeps `TableProps` from antd and `RowSelectionType`/`TableRowSelection` from `antd/es/table/interface` in the surviving tree. Line 39 still declares `keepSelectionOnScopeChange`, the prop this PR removes from both call sites.

Moving `RenderButtonProps` next to the buttons would let the rest of the file go with it.

`listExpandedViewRegistry` (`entity-configuration/definitions/list-expanded-view-defs/index.ts:21`) is in the same position: `browse-entity-legacy.tsx` was its only consumer. The per-entity configs it aggregates are still imported directly by `circuit-simulations.ts` and `campaign-common.tsx`, so only the registry itself is dead.

src/ui/segments/explore/circuit/index.tsx
```diff
@@ -1,476 +0,0 @@
-'use client';
```
pgetta: Not blocking. Deleting this file leaves three things with no importers:

* `use-filter-state-watcher.ts` - the whole file, and it keeps `data-table/elements/context` alive.
* `elements/expand-icon.tsx` - the whole file. The PR body lists the expandable-row helpers as removed, this is the one that got missed.
* the `useHierarchy` export in `use-hierarchy.tsx:184` - the file stays for `useHierarchyAllLevels`/`useHierarchyDerivationTree`, but the grid fetches hierarchy through `createCircuitDataSource` now.

I would drop the `useHierarchy` export at least, leaving it there invites someone to wire the legacy fetch path back into the grid.

src/ui/segments/explore/circuit/elements/circuit-listing-grid.test.tsx
```diff
@@ -1,7 +1,6 @@
 import { describe, expect, it } from 'vitest';
 
-import { CircuitRepresentationView } from '../helpers';
-import { circuitListingRowClass } from './circuit-listing-grid';
+import { CircuitRepresentationView, circuitListingRowClass } from '../helpers';
```
pgetta: Now that the subject moved to `../helpers.ts`, would it make sense to move the file as well? It sits in `elements/` under the name of a module that no longer exists, so coverage for `helpers.ts` is hard to find and `elements/` looks like it holds a test for a deleted file. `explore/circuit/helpers.test.ts` would say it. Not blocking.

src/features/views/listing/types.ts
```diff
@@ -0,0 +1,104 @@
...
+import type { TViewVariant, TWorkspaceScope, TWorkspaceSection } from '@/constants';
+import type { TFacets } from '@/features/data-grid/core';
+import type { WorkspaceContext } from '@/types/common';
+import type { MiniDetailView } from '@/ui/segments/mini-detail-view';
+
+/**
+ * Row-interaction options for a listing. Previously derived from the antd table's own
+ * props; declared explicitly here so the listing contract does not depend on a
+ * rendering component.
+ */
```
pgetta: Nice, declaring the listing contract here instead of deriving it from the antd table props is what makes the rest of this deletion possible.

src/features/views/listing/types.ts
bilalesi: fixed

src/features/views/listing/browse-entity.tsx
bilalesi: nice catch , fixed

src/ui/segments/explore/circuit/index.tsx
bilalesi: done

src/ui/segments/data-table/table.tsx
bilalesi: done

src/ui/segments/workflows/browse/listing.tsx
bilalesi: fixed

pgetta (approved):
Nice!

bilalesi:
thank you  @pgetta

